PR #28, opened 2019-01-21
Do not require ECV key if there are no encrypted values

### LINKED ISSUE #25: Do not require ECV key if there are no encrypted values

bmoylan:
Fixes #25

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/28)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,11 @@ configuration (for example, in-memory providers).
 
 `witchcraft-server` also supports using `encrypted-config-value` to automatically decrypt encrypted configuration
 values. The default configuration expects a key file to be at `var/conf/encrypted-config-value.key`. It is possible to 
-use code to specify a different source for the key (or to specify that no key should be used). 
+use code to specify a different source for the key (or to specify that no key should be used). If the configuration
+does not contain encrypted values, any specified ECV key will not be read. If the install configuration contains
+encrypted values but the encryption key is missing or malformed, the server will fail to start. If the runtime config
+contains encrypted values but fails to decrypt them, a warning will be logged and the encrypted values passed to the
+server.
 
 `witchcraft-server` defines base configuration for its install and runtime configuration. Servers that want to provide
 their own install and/or runtime configuration should embed the base configuration structs within the definition of 
--- a/integration/config_test.go
+++ b/integration/config_test.go
@@ -1,0 +1,145 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


nmiyake (changes requested):
Thanks! As part of this, can you add a bit of clarification to the README as well? Right now, the README states:

```
witchcraft-server also supports using encrypted-config-value to automatically decrypt encrypted configuration values. The default configuration expects a key file to be at var/conf/encrypted-config-value.key. It is possible to use code to specify a different source for the key (or to specify that no key should be used).
```

I think we should add a section afterwards that describes the other aspects of the behavior:
* If an ECV key file does not exist at the config path on startup, no ECV key is used
* It is acceptable to run with no ECV key defined as long as the neither the install nor runtime config contain encrypted values
* If the server does not have an ECV key defined and encounters any config with encrypted values, it will be treated as an error condition

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/28#-:-LWlgUQYBvvfruypuXur:b21kl9z)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[integration/config_test.go, line 32 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/28#-LWleYzo21m94XHNVOTm:-LWleYzp0xhcoH6TR38r:bpvpro9) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9c06ba55f421e58cbee12ca9f956b41e9dadc794/integration/config_test.go#L32)):*
> ```Go
> 
> func TestEncryptedConfig(t *testing.T) {
> 	decryptedValue := "hello world"
> ```

Since these are all string constants, prefer declaring as constants in block notation here:

```go
const (
    decryptedValue = "hello world"
    ...
)
```

nmiyake:
Thanks! Implementation looks good -- only request is to add some documentation on the behavior.

bmoylan (commented):
done

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/28#-:-LWlsNQ40r2sUJoDOItN:bb7f2mf)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[integration/config_test.go, line 32 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/28#-LWleYzo21m94XHNVOTm:-LWlsOSD8XEQRLyGhhhA:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9c06ba55f421e58cbee12ca9f956b41e9dadc794/integration/config_test.go#L32)):*

Since these are all string constants, prefer declaring as constants in block notation here:

```go
const (
    decryptedValue = "hello world"
    ...
)
```
</blockquote></details>

Done.

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/28)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

nmiyake (approved):

PR #37, opened 2019-02-09
Implement and add prometheus metrics collection endpoints

gdearment:
- Adds a witchcraft/prometheus package which includes a prometheus.Gatherer
  implementation for gathering metrics stored in a metrics.Registry.
- Registers a `/prometheus/metrics` endpoint on the management router
  to enable metrics to be collected via prometheus collectors

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/37)


gdearment:
Taking a dependency on the `prometheus/promhttp` package ends up adding quite a bit to the vendor tree. We could look to do a simplified version of the `http.Handler` and only support text format if we wanted to avoid adding protobuf to the vendor tree.

nmiyake:
Question: is there a way to make this pluggable?

So far, most of `witchcraft-go-server` has not made explicit coupling decisions with external APIs. This change would explicitly bake in a Prometheus endpoint (as opposed to some other metric/monitoring framework).

I feel like a more appropriate composition would be to isolate the logic that performs the built-in metrics -> prometheus translation, but not hook it up by default. This may require `witchcraft-go-server` to expose access to the metrics to APIs or something like that, but I think that's more doable.

Ideally, I would like the state to be such that a project can import/use `witchcraft-go-server` without pulling in a Prometheus dependency unless the project actually wants to use it.

If we want to dictate that `withcraft-go-server` is opinionated in its coupling with Prometheus, then I think that's something we should call out/explain more explicitly in top-level documentation.

PR #41, opened 2019-02-15
fix Server.Start() error handling

jdhenke:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)

### DIFF
--- a/witchcraft/server_run.go
+++ b/witchcraft/server_run.go
@@ -85,8 +85,7 @@ func newServerStartShutdownFns(
 			svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
 			return nil
 		}
-		svcLogger.Error(fmt.Sprintf("%s failed", serverName), svc1log.Stacktrace(err))
-		return err
+		return werror.Wrap(err, "server was closed", werror.SafeParam("serverName", serverName))
 	}, httpServer.Shutdown, nil
 }
 
--- a/witchcraft/server_run.go
+++ b/witchcraft/server_run.go
@@ -32,7 +32,7 @@ import (
 	"github.com/palantir/witchcraft-go-server/config"
 )
 
-func (s *Server) newServer(productName string, serverConfig config.Server, handler http.Handler) (rHTTPServer *http.Server, rStart func(), rShutdown func(context.Context) error, rErr error) {
+func (s *Server) newServer(productName string, serverConfig config.Server, handler http.Handler) (rHTTPServer *http.Server, rStart func() error, rShutdown func(context.Context) error, rErr error) {
 	return newServerStartShutdownFns(
 		serverConfig,
 		s.useSelfSignedServerCertificate,
@@ -43,7 +43,7 @@ func (s *Server) newServer(productName string, serverConfig config.Server, handl
 	)
 }
 
-func (s *Server) newMgmtServer(productName string, serverConfig config.Server, handler http.Handler) (rStart func(), rShutdown func(context.Context) error, rErr error) {
+func (s *Server) newMgmtServer(productName string, serverConfig config.Server, handler http.Handler) (rStart func() error, rShutdown func(context.Context) error, rErr error) {
 	serverConfig.Port = serverConfig.ManagementPort
 	_, start, shutdown, err := newServerStartShutdownFns(
 		serverConfig,
@@ -63,7 +63,7 @@ func newServerStartShutdownFns(
 	serverName string,
 	svcLogger svc1log.Logger,
 	handler http.Handler,
-) (rHTTPServer *http.Server, start func(), shutdown func(context.Context) error, rErr error) {
+) (rHTTPServer *http.Server, start func() error, shutdown func(context.Context) error, rErr error) {
 	tlsConfig, err := newTLSConfig(serverConfig, useSelfSignedServerCertificate, clientAuthType)
 	if err != nil {
 		return nil, nil, nil, err
@@ -75,17 +75,18 @@ func newServerStartShutdownFns(
 		TLSConfig: tlsConfig,
 		Handler:   handler,
 	}
-	return httpServer, func() {
+	return httpServer, func() error {
 		svcLogger.Info("Listening to https", svc1log.SafeParam("address", addr), svc1log.SafeParam("server", serverName))
 
 		// cert and key specified in TLS config so no need to pass in here
-		if err := httpServer.ListenAndServeTLS("", ""); err != nil {
-			if err == http.ErrServerClosed {
-				svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
-				return
-			}
-			svcLogger.Error(fmt.Sprintf("%s failed", serverName), svc1log.Stacktrace(err))
+		err := httpServer.ListenAndServeTLS("", "")
+		// ListenAndServeTLS always returns a non-nil error, so we only need to check for ErrServerClosed
+		if err == http.ErrServerClosed {
+			svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
+			return nil
 		}
+		svcLogger.Error(fmt.Sprintf("%s failed", serverName), svc1log.Stacktrace(err))
+		return err
 	}, httpServer.Shutdown, nil
 }
 
--- a/witchcraft/server_run.go
+++ b/witchcraft/server_run.go
@@ -79,13 +79,14 @@ func newServerStartShutdownFns(
 		svcLogger.Info("Listening to https", svc1log.SafeParam("address", addr), svc1log.SafeParam("server", serverName))
 
 		// cert and key specified in TLS config so no need to pass in here
-		err := httpServer.ListenAndServeTLS("", "")
-		// ListenAndServeTLS always returns a non-nil error, so we only need to check for ErrServerClosed
-		if err == http.ErrServerClosed {
-			svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
-			return nil
+		if err := httpServer.ListenAndServeTLS("", ""); err != nil {
+			if err == http.ErrServerClosed {
+				svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
+				return nil
+			}
+			return werror.Wrap(err, "server failed", werror.SafeParam("serverName", serverName))
 		}
-		return werror.Wrap(err, "server failed", werror.SafeParam("serverName", serverName))
+		return nil
 	}, httpServer.Shutdown, nil
 }
 
--- a/witchcraft/server_run.go
+++ b/witchcraft/server_run.go
@@ -85,7 +85,7 @@ func newServerStartShutdownFns(
 			svcLogger.Info(fmt.Sprintf("%s was closed", serverName))
 			return nil
 		}
-		return werror.Wrap(err, "server was closed", werror.SafeParam("serverName", serverName))
+		return werror.Wrap(err, "server failed", werror.SafeParam("serverName", serverName))
 	}, httpServer.Shutdown, nil
 }
 
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -534,7 +534,9 @@ func (s *Server) Start() (rErr error) {
 		}
 
 		// start management server in its own goroutine
-		go mgmtStart()
+		go func() {
+			_ = mgmtStart()
+		}()
 		defer func() {
 			// nothing to be done if shutdown fails
 			_ = mgmtShutdown(ctx)
@@ -548,8 +550,7 @@ func (s *Server) Start() (rErr error) {
 
 	s.httpServer = httpServer
 	s.stateManager.setState(ServerRunning)
-	svrStart()
-	return nil
+	return svrStart()
 }
 
 type configurableRouterImpl struct {
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -536,7 +536,7 @@ func (s *Server) Start() (rErr error) {
 		// start management server in its own goroutine
 		go func() {
 			if err := mgmtStart(); err != nil {
-				svc1log.FromContext(ctx).Error("management server was closed", svc1log.Stacktrace(err))
+				svc1log.FromContext(ctx).Error("management server failed", svc1log.Stacktrace(err))
 			}
 		}()
 		defer func() {
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -535,7 +535,9 @@ func (s *Server) Start() (rErr error) {
 
 		// start management server in its own goroutine
 		go func() {
-			_ = mgmtStart()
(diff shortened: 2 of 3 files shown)


bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)* status: all files reviewed, 3 unresolved discussions (waiting on @jdhenke)

___
*[witchcraft/server_run.go, line 88 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidMPQABXa4RjtVj4F:-LYidMPQABXa4RjtVj4G:b-2oio1f) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/server_run.go#L88)):*
> ```Go
> 			return nil
> 		}
> 		svcLogger.Error(fmt.Sprintf("%s failed", serverName), svc1log.Stacktrace(err))
> ```

We should be able do delete this log line; by returning an error it will get logged by the server's `Start()` method. We probably want to `werror.Wrap` it with a param for the serverName though.
___
*[witchcraft/witchcraft.go, line 538 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidilRATPWNht4XBwI:-LYidilRATPWNht4XBwJ:bl74ezr) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/witchcraft.go#L538)):*
> ```Go
> 		// start management server in its own goroutine
> 		go func() {
> 			_ = mgmtStart()
> ```

hm I wonder if we can be smarter about handling this error; it should probably be logged if we are not going to return it
___
*[witchcraft/witchcraft_test.go, line 105 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidsgxCztWmxacd86o:-LYidsgxCztWmxacd86p:b-8op5ul) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/witchcraft_test.go#L105)):*
> ```Go
> 
> 	// create a listener to ensure a port is taken
> 	ln, err := net.Listen("tcp", "127.0.0.1:0")
> ```

we've got a little utility for this if you'd like: https://github.com/palantir/pkg/blob/master/httpserver/httpserver.go#L20

jdhenke (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @bmoylan)

___
*[witchcraft/server_run.go, line 88 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidMPQABXa4RjtVj4F:-LYnqH6R53Y1US0o4I7U:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/server_run.go#L88)):*

We should be able do delete this log line; by returning an error it will get logged by the server's `Start()` method. We probably want to `werror.Wrap` it with a param for the serverName though.
</blockquote></details>

Done.
___
*[witchcraft/witchcraft.go, line 538 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidilRATPWNht4XBwI:-LYnrXlZ6ifutzms4LlU:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/witchcraft.go#L538)):*

hm I wonder if we can be smarter about handling this error; it should probably be logged if we are not going to return it
</blockquote></details>

Done.
___
*[witchcraft/witchcraft_test.go, line 105 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYidsgxCztWmxacd86o:-LYnrgY86wvKZvAooQ_r:b-tinz77) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0912108424bfa81799dd3e52f32eb9d0baf090b2/witchcraft/witchcraft_test.go#L105)):*

we've got a little utility for this if you'd like: https://github.com/palantir/pkg/blob/master/httpserver/httpserver.go#L20
</blockquote></details>

Nice! I think I'll stick with just using `net.Listen` with a port of 0 because we actually want to keep the listener alive for the duration of the test to ensure the port **is** in use. Will keep in mind for the future though :thumbsup: .

bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)* status: all files reviewed, 1 unresolved discussion (waiting on @jdhenke)

___
*[witchcraft/server_run.go, line 88 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYwaA4c5tO9vTnQP0oc:-LYwaA4c5tO9vTnQP0od:b6gmp4c) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b2e10467f87182b4cd1f49243d17850624edb89b/witchcraft/server_run.go#L88)):*
> ```Go
> 			return nil
> 		}
> 		return werror.Wrap(err, "server was closed", werror.SafeParam("serverName", serverName))
> ```

nit: this is actually the case where the server failed but was _not_ closed. Let's go with `server failed` instead

jdhenke (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/server_run.go, line 88 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LYwaA4c5tO9vTnQP0oc:-LZ0EIGlEyDH0vVCkXV1:bb6un1j) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b2e10467f87182b4cd1f49243d17850624edb89b/witchcraft/server_run.go#L88)):*

nit: this is actually the case where the server failed but was _not_ closed. Let's go with `server failed` instead
</blockquote></details>

Done

nmiyake (changes requested):
Thanks for the PR! Some minor feedback on code structure in main logic and tests, but in sync with over-all PR.

> Reviewed 1 of 3 files at r1, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-:-LZ10xtk8qCuVuP46gfU:b-fklpdp)* status: all files reviewed, 5 unresolved discussions (waiting on @bmoylan and @jdhenke)

___
*[witchcraft/server_run.go, line 82 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ10QtS8CzJL2V1_n0U:-LZ10QtS8CzJL2V1_n0V:byhgg9z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/server_run.go#L82)):*
> ```Go
> 
> 		// cert and key specified in TLS config so no need to pass in here
> 		err := httpServer.ListenAndServeTLS("", "")
> ```

nit: slight preference for keeping the previous construction here:

```go
if err := httpServer.ListenAndServeTLS("", ""); err != nil {
      if err == http.ErrServerClosed {
          ...
      }
      return werror.Wrap(err, "server failed", werror.SafeParam("serverName", serverName))
}
return nil
```

I know that this is functionally equivalent, but in tracing through code I think it's nice to have the `return nil` be explicit rather than depending on behavior that `werror.Wrap` returns nil if provided with a `nil` error.
___
*[witchcraft/witchcraft_test.go, line 134 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ0p6T96qdLjl-OyB-X:-LZ0p6T96qdLjl-OyB-Y:bm1vknf) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L134)):*
> ```Go
> 	case serr := <-errc:
> 		assert.NotNil(t, serr, "start returned a nil error when its port was already in use")
> 	case <-time.Tick(2 * time.Second):
> ```

nit: semantically, since we are treating this as a timeout, I think that `<-time.After(2 * time.Second)` is more appropriate. `time.Tick` should be used for scenarios where you want something to be sent on a channel in a recurring manner (like the example in https://golang.org/pkg/time/#Tick)
___
*[witchcraft/witchcraft_test.go, line 163 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ0vVe-1cUk04ToCjzl:-LZ0vVe-1cUk04ToCjzm:bovi92n) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L163)):*
> ```Go
> 		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
> 		defer cancel()
> 		for !server.Running() && ctx.Err() == nil {
> ```

Even though it might be a bit more code, would recommend restructuring this logic to use channels:

```go
started := make(chan bool)
go func() {
	for _ = range time.Tick(100 * time.Millisecond) {
                if server.Running() {
			started <- true
		}
	}
	started <- true
}()
select {
case <-started:
	break
case <-time.After(2 * time.Second):
	assert.Fail("case %s: timed out waiting for server to start")
}
```

Personally, I think this makes it clearer to reason about the timeout etc. (as currently written, took a second for me to figure out that the `ctx.Err()  == nil` in the loop conditional is effectively functioning as a timeout from the context).
___
*[witchcraft/witchcraft_test.go, line 168 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ1-OsL7UFbjwIjhd_d:-LZ1-OsM5qAddcQPMo4X:b-xz9psg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L168)):*
> ```Go
> 		assert.NoError(t, ctx.Err(), "case %s: timed out waiting for server to start", name)
> 		assert.NoError(t, shutdown(server, ctx), "error running %s", name)
> 		assert.Nil(t, <-errc, "%s: Start() incorrectly returned a non-nil error when the server was gracefully shutdown", name)
> ```

Reading from a channel in a test case should always be done in a select with a timeout to ensure that the test fails gracefully if for some reason the receive on the channel doesn't execute:

```go
select {
   case err <- errc:
      assert.Nil(...)
   case <-time.After(2*time.Second):
      assert.Fail("timed out waiting for errc")
}
```

jdhenke (commented):
Sweet, thanks for the thorough review! I made adjustments per your comments and refactored the test code to improve cleanup a bit more. 

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-:-LZ5lUJ638Q9nCK5FIg1:b-2j3par)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @bmoylan and @nmiyake)

___
*[witchcraft/server_run.go, line 82 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ10QtS8CzJL2V1_n0U:-LZ5_jhLArjJ-z7cVIqA:bq6r8y8) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/server_run.go#L82)):*

nit: slight preference for keeping the previous construction here:

```go
if err := httpServer.ListenAndServeTLS("", ""); err != nil {
      if err == http.ErrServerClosed {
          ...
      }
      return werror.Wrap(err, "server failed", werror.SafeParam("serverName", serverName))
}
return nil
```

I know that this is functionally equivalent, but in tracing through code I think it's nice to have the `return nil` be explicit rather than depending on behavior that `werror.Wrap` returns nil if provided with a `nil` error.
</blockquote></details>

Done. Put back as you suggested. Did want to note that, per the comment I added, `http.Server.ListenAndServeTLS` [always returns a non-nil error](https://golang.org/src/net/http/server.go?s=93179:93228#L2992).
___
*[witchcraft/witchcraft_test.go, line 134 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ0p6T96qdLjl-OyB-X:-LZ5aa_x0krBfogKZCHZ:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L134)):*

nit: semantically, since we are treating this as a timeout, I think that `<-time.After(2 * time.Second)` is more appropriate. `time.Tick` should be used for scenarios where you want something to be sent on a channel in a recurring manner (like the example in https://golang.org/pkg/time/#Tick)
</blockquote></details>

Done.
___
*[witchcraft/witchcraft_test.go, line 163 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ0vVe-1cUk04ToCjzl:-LZ5fYtIEtPjkpe-6pBm:b-eu072s) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L163)):*

Even though it might be a bit more code, would recommend restructuring this logic to use channels:

```go
started := make(chan bool)
go func() {
	for _ = range time.Tick(100 * time.Millisecond) {
                if server.Running() {
			started <- true
		}
	}
	started <- true
}()
select {
case <-started:
	break
case <-time.After(2 * time.Second):
	assert.Fail("case %s: timed out waiting for server to start")
}
```

Personally, I think this makes it clearer to reason about the timeout etc. (as currently written, took a second for me to figure out that the `ctx.Err()  == nil` in the loop conditional is effectively functioning as a timeout from the context).
</blockquote></details>

Done, although refactored your proposal a bit so curious what you think.
___
*[witchcraft/witchcraft_test.go, line 168 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/41#-LZ1-OsL7UFbjwIjhd_d:-LZ5fxcZ2cz_JI3lQHb8:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5bed07f5c707ed64da8c9eb6bebc231313d0b0ea/witchcraft/witchcraft_test.go#L168)):*

Reading from a channel in a test case should always be done in a select with a timeout to ensure that the test fails gracefully if for some reason the receive on the channel doesn't execute:

```go
select {
   case err <- errc:
      assert.Nil(...)
   case <-time.After(2*time.Second):
      assert.Fail("timed out waiting for errc")
}
```
</blockquote></details>

Done.

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/41)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

PR #48, opened 2019-03-12
Add metrics registry to server context

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/48)

### DIFF
--- a/witchcraft/internal/middleware/request.go
+++ b/witchcraft/internal/middleware/request.go
@@ -69,6 +69,17 @@ func NewRequestContextLoggers(
 	}
 }
 
+// NewRequestContextMetricsRegistry is request middleware that sets the metrics registry on the request context.
+func NewRequestContextMetricsRegistry(metricsRegistry metrics.Registry) wrouter.RequestHandlerMiddleware {
+	return func(rw http.ResponseWriter, req *http.Request, next http.Handler) {
+		ctx := req.Context()
+		if metricsRegistry != nil {
+			ctx = metrics.WithRegistry(ctx, metricsRegistry)
+		}
+		next.ServeHTTP(rw, req.WithContext(ctx))
+	}
+}
+
 func NewRequestExtractIDs(
 	svcLogger svc1log.Logger,
 	trcLogger trc1log.Logger,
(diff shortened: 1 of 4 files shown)


nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/48)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 505 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/48#-L_mqNNb2EHs5QsUfC8X:-L_mqNNb2EHs5QsUfC8Y:b-youi62) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c6ff0af14f63c835c99b27a770d62888d5666574/witchcraft/witchcraft.go#L505)):*
> ```Go
> 	}
> 	defer metricsDeferFn()
> 	ctx = metrics.WithRegistry(ctx, metricsRegistry)
> ```

This only sets the registry on the server lifecycle context (the one provided to the `initFunc`).

If we want to include this value on the context provided to every request (which I believe we do), then we'll need to store the registry as part of the server struct and then add it to the request contexts in the middleware (`witchcraft/internal/middleware/request.go`)

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/48)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[witchcraft/witchcraft.go, line 505 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/48#-L_mqNNb2EHs5QsUfC8X:-L_mspL9Bv6XDd_1FMAx:bc9pht3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c6ff0af14f63c835c99b27a770d62888d5666574/witchcraft/witchcraft.go#L505)):*

This only sets the registry on the server lifecycle context (the one provided to the `initFunc`).

If we want to include this value on the context provided to every request (which I believe we do), then we'll need to store the registry as part of the server struct and then add it to the request contexts in the middleware (`witchcraft/internal/middleware/request.go`)
</blockquote></details>

thanks, done. I didn't realize the registry passed into the middleware was only used for the builtin server.* metrics.

nmiyake (approved):
> Reviewed 1 of 1 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/48)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #49, opened 2019-03-20
Increase default log retention settings

tabboud:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/49)


nmiyake (approved):
Thanks! This change looks good to me, but would like to wait for @bmoylan or @gdearment to take a look and provide input on whether they think this is a safe change to make for our expected default usage.

bmoylan (approved):
Yep this should be fine

Issue #59, opened 2019-04-03
Attempt graceful shutdown on SIGTERM


najork:
Witchcraft should attempt to gracefully shutdown servers upon receiving a `SIGTERM`. A graceful shutdown should include responding `false` from its readiness endpoint, attempting to drain active connections, and responding to new requests with a `503 Service Unavailable`.

PR #61, opened 2019-04-04
Shutdown server on SIGTERM

najork:
Fixes https://github.com/palantir/witchcraft-go-server/issues/59

This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61)

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,9 @@ the `WithDisableGoRuntimeMetrics` server method.
 `server.WithDisableSigQuitHandler`.  If `server.WithSigQuitHandlerWriter` is used, the stacks will also be written in
 their unparsed form to the provided writer.
 
+### Shutdown signal handling
+`witchcraft-server` attempts to drain active connections and gracefully shut down by calling `server.Shutdown` upon receiving a SIGTERM or SIGINT signal. This behavior can be disabled using `server.WithDisableShutdownSignalHandler`.
+
 Example server initialization
 -----------------------------
 
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -81,6 +81,9 @@ type Server struct {
 	// if true, disables the default behavior of emitting a goroutine dump on SIGQUIT signals.
 	disableSigQuitHandler bool
 
+	// if true, disables the default behavior of shutting down the server on SIGTERM and SIGINT signals.
+	disableShutdownSignalHandler bool
+
 	// provides the bytes for the install configuration for the server. If nil, a default configuration provider that
 	// reads the file at "var/conf/install.yml" is used.
 	installConfigProvider ConfigBytesProvider
@@ -380,6 +383,12 @@ func (s *Server) WithSigQuitHandlerWriter(w io.Writer) *Server {
 // WithDisableSigQuitHandler disables the server's enabled-by-default goroutine dump on SIGQUIT.
 func (s *Server) WithDisableSigQuitHandler() *Server {
 	s.disableSigQuitHandler = true
+	return s
+}
+
+// WithDisableShutdownSignalHandler disables the server's enabled-by-default shutdown on SIGTERM and SIGINT.
+func (s *Server) WithDisableShutdownSignalHandler() *Server {
+	s.disableShutdownSignalHandler = true
 	return s
 }
 
@@ -702,6 +711,10 @@ func (s *Server) initStackTraceHandler(ctx context.Context) {
 }
 
 func (s *Server) initShutdownSignalHandler(ctx context.Context) {
+	if s.disableShutdownSignalHandler {
+		return
+	}
+
 	shutdownSignal := make(chan os.Signal, 1)
 	signal.Notify(shutdownSignal, syscall.SIGTERM, syscall.SIGINT)
 
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -25,6 +25,7 @@ import (
 	"os/signal"
 	"reflect"
 	"runtime/debug"
+	"sync"
 	"syscall"
 	"time"
 
@@ -172,6 +173,9 @@ type Server struct {
 
 	// the http.Server for the main server
 	httpServer *http.Server
+
+	// allows the server to wait until Close() or Shutdown() return prior to returning from Start()
+	shutdownFinished sync.WaitGroup
 }
 
 // InitFunc is a function type used to initialize a server. ctx is a context configured with loggers and is valid for
@@ -423,7 +427,7 @@ const (
 	runtimeConfigPath = "var/conf/runtime.yml"
 )
 
-// Start begins serving HTTPS traffic and blocks until s.Close() or s.Shutdown() are called.
+// Start begins serving HTTPS traffic and blocks until s.Close() or s.Shutdown() return.
 // Errors are logged via s.svcLogger before being returned.
 // Panics are recovered; in the case of a recovered panic, Start will log and return
 // a non-nil error containing the recovered object (overwriting any existing error).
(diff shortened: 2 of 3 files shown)


bmoylan (changes requested):
Thanks for submitting! 

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-:-LbdMPcWE_M8jeI8cEZP:bw0tsge)* status: all files reviewed, 2 unresolved discussions (waiting on @najork)

___
*[witchcraft/witchcraft.go, line 706 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdMSJD7x-E2GVjWZw1:-LbdMSJD7x-E2GVjWZw2:bwoi35a) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L706)):*
> ```Go
> func (s *Server) initSigtermHandler(ctx context.Context) {
> 	sigterm := make(chan os.Signal, 1)
> 	signal.Notify(sigterm, syscall.SIGTERM)
> ```

Let's also catch `SIGINT` here
___
*[witchcraft/witchcraft.go, line 710 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdM6hhDU_O95uG79Ii:-LbdM6hhDU_O95uG79Ij:bd1ziky) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L710)):*
From https://golang.org/pkg/net/http/#Server.Shutdown:
> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.

We need to track whether or not `Shutdown` has been called and add an extra blocking condition to the end of `Start()`. We don't want to hang forever if the shutdown is triggered a different way (setup returns an error, someone calls `Close()`, etc) so maybe we introduce a [WaitGroup](https://golang.org/pkg/sync/#WaitGroup) and increment it in `Shutdown` then defer a `Wait`

najork (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 706 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdMSJD7x-E2GVjWZw1:-LbdPiEw8oZVTelTHGmb:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L706)):*

Let's also catch `SIGINT` here
</blockquote></details>

Done.
___
*[witchcraft/witchcraft.go, line 710 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdM6hhDU_O95uG79Ii:-LbdQ5wZ-c7ypOxbeIbT:b-rah7p1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L710)):*
I'm not sure whether I fully understand this comment. The godoc on [witchcraft.Start](https://github.com/palantir/witchcraft-go-server/blob/develop/witchcraft/witchcraft.go#L419) states:
> Start begins serving HTTPS traffic and blocks until s.Close() or s.Shutdown() are called.

Are we changing the behavior to block until `s.Shutdown()` returns?

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61)* status: all files reviewed, 1 unresolved discussion (waiting on @najork)

___
*[witchcraft/witchcraft.go, line 710 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdM6hhDU_O95uG79Ii:-LbdSX4k202K9fpQmWaK:b-wj1xe1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L710)):*
<details><summary><i>Previously, najork (Max Najork) wrote…</i></summary><blockquote>

I'm not sure whether I fully understand this comment. The godoc on [witchcraft.Start](https://github.com/palantir/witchcraft-go-server/blob/develop/witchcraft/witchcraft.go#L419) states:
> Start begins serving HTTPS traffic and blocks until s.Close() or s.Shutdown() are called.

Are we changing the behavior to block until `s.Shutdown()` returns?

</blockquote></details>

Yeah, so the last line of `Start()` wraps `ListenAndServeTLS`, which blocks while the server is running. As written, this will return when Shutdown is **called**, which does not allow time for graceful connection draining. Instead we need to wait for Shutdown to **return** before `Start` returns, since most applications will exit immediately after Start. 

I mentioned the thing about the WaitGroup because there are multiple ways the server could be closed. Otherwise I'd just add a `shutdownFinished chan struct{}` to `*Server` and `defer func() { <-s.shutdownFinished }` or something

bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61)* status: all files reviewed, 3 unresolved discussions (waiting on @najork)

*[a discussion](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdU0Ah7E8QWaDr-fmi:-LbdU0Ah7E8QWaDr-fmj:b2r5vr7) (no related file):*
Can you add a section to the README explaining this feature?

---
*[a discussion](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdU2zE0XlFyC1XEsks:-LbdU2zE0XlFyC1XEskt:b-nw8sa1) (no related file):*
We should probably have a `.WithDisableShutdownSignals` option that turns this off just in case

---

najork (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @bmoylan)

*[a discussion](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdU0Ah7E8QWaDr-fmi:-Lbdngx8BOXOBYGc7_C7:b-896fix) (no related file):*

Can you add a section to the README explaining this feature?
</blockquote></details>

Done.

---
*[a discussion](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdU2zE0XlFyC1XEsks:-LbdnhR6CRAkB4nzE8Hb:b-896fix) (no related file):*

We should probably have a `.WithDisableShutdownSignals` option that turns this off just in case
</blockquote></details>

Done.

---
___
*[witchcraft/witchcraft.go, line 710 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-LbdM6hhDU_O95uG79Ii:-LbdngCyD7vj9_eE08OH:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/cfcef11f84996839c3b9c563a8d76899f392ef29/witchcraft/witchcraft.go#L710)):*

Yeah, so the last line of `Start()` wraps `ListenAndServeTLS`, which blocks while the server is running. As written, this will return when Shutdown is **called**, which does not allow time for graceful connection draining. Instead we need to wait for Shutdown to **return** before `Start` returns, since most applications will exit immediately after Start. 

I mentioned the thing about the WaitGroup because there are multiple ways the server could be closed. Otherwise I'd just add a `shutdownFinished chan struct{}` to `*Server` and `defer func() { <-s.shutdownFinished }` or something
</blockquote></details>

Done.

bmoylan:
Looks good! 

Can we add a test that 
- starts a server with a slow endpoint (e.g. something that sleeps for a second)
- client makes a request
- test calls shutdown
- ensure client eventually receives a valid response

najork:
I believe this behavior should be covered by https://github.com/palantir/witchcraft-go-server/blob/develop/integration/server_test.go#L59

bmoylan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/61#-:-LbeTMOL7wol-EX5YgCP:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #65, opened 2019-04-10
Add configuration field for trace-sample-rate

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,9 @@ the request specifies a `X-B3-Sampled` header, the value specified in that heade
 header is not present, whether or not the trace is sampled is determined by the sampling source configured for the 
 `witchcraft-server` (by default, all traces are sampled). If a trace is not sampled, `witchcraft-server` will not 
 generate any trace log output for it. However, the infrastructure will still perform all of the trace-related operations
-(such as creating child spans and setting span information on headers).
+(such as creating child spans and setting span information on headers). The install configuration field
+`trace-sample-rate` represents a float between 0 and 1 (inclusive) to control the proportion of traces sampled by
+default. If the `WithTraceSampler` server option is provided, it overrides this configuration.
 
 `witchcraft-server` also ensures that the context for every request has a trace ID. After the logging middleware 
 executes, the request is guaranteed to have a trace ID (either from the incoming request or from the newly generated 
--- a/config/install.go
+++ b/config/install.go
@@ -24,6 +24,7 @@ type Install struct {
 	ProductName          string        `yaml:"product-name,omitempty"`
 	Server               Server        `yaml:"server,omitempty"`
 	MetricsEmitFrequency time.Duration `yaml:"metrics-emit-frequency,omitempty"`
+	TraceSampleRate      *float64      `yaml:"trace-sample-rate,omitempty"`
 	UseConsoleLog        bool          `yaml:"use-console-log,omitempty"`
 }
 
--- a/witchcraft/trace_test.go
+++ b/witchcraft/trace_test.go
@@ -24,68 +24,59 @@ import (
 func TestSamplerForRate(t *testing.T) {
 	for _, test := range []struct {
 		rate     float64
-		expected map[uint64]bool
+		expected [16]bool
 	}{
 		{
+			rate: -1,
+			expected: [16]bool{false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false},
+		},
+		{
 			rate: 0,
-			expected: map[uint64]bool{
-				0: false,
-				1: false,
-				2: false,
-			},
-		},
-		{
-			rate: 1,
-			expected: map[uint64]bool{
-				0: true,
-				1: true,
-				2: true,
-			},
-		},
-		{
-			rate: 0.25,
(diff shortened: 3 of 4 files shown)


nmiyake (changes requested):
The code structure looks good, but have concerns about the sampling algorithm/logic.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65#-:-Lc7ekPG-3w_uJgNYDcw:b-nrov2u)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 833 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/65#-Lc7ciTcCvSvO35j1-gO:-Lc7ciTcCvSvO35j1-gP:b-fgj5yg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d58e6cc5cbf823bdf3f537d56a17604b7ffce5d6/witchcraft/witchcraft.go#L833)):*
> ```Go
> 	}
> 	return func(id uint64) bool {
> 		mod := uint64(1 / sampleRate)
> ```

I believe this is too coarse.

For example, if `sampleRate` is set to `0.9` ("sample 90%"), `1/sampleRate` = 1.111, which turns into 1, which then effectively becomes a 100% sample rate.

I believe that scaling the maximal value of `uint64` based on the input ratio and comparing the ID to that would be a better approach -- for example, https://github.com/openzipkin/zipkin/blob/a40ebcb47986ed1efb85a0543a17611d78404eb2/zipkin-collector/core/src/main/java/zipkin2/collector/CollectorSampler.java#L68.

As noted in the comment there, the approach may not be statistically perfect, but is likely good enough.

MatthewMaclean:
👍  from me

nmiyake:
Need to run `format`

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[witchcraft/trace_test.go, line 27 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/65#-Lc7oyw99h8PaRm-6fte:-Lc7oyw99h8PaRm-6ftf:b-m2gzph) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/31d1ad8ef9e0620dc85f917e795a2bb73690d887/witchcraft/trace_test.go#L27)):*
> ```Go
> 	for _, test := range []struct {
> 		rate     float64
> 		expected [16]bool
> ```

let me know if you have a cleaner way to write this test...
___
*[witchcraft/witchcraft.go, line 833 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/65#-Lc7ciTcCvSvO35j1-gO:-Lc7ovUcBpj3QIrmjfsn:b-trrfty) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d58e6cc5cbf823bdf3f537d56a17604b7ffce5d6/witchcraft/witchcraft.go#L833)):*

I believe this is too coarse.

For example, if `sampleRate` is set to `0.9` ("sample 90%"), `1/sampleRate` = 1.111, which turns into 1, which then effectively becomes a 100% sample rate.

I believe that scaling the maximal value of `uint64` based on the input ratio and comparing the ID to that would be a better approach -- for example, https://github.com/openzipkin/zipkin/blob/a40ebcb47986ed1efb85a0543a17611d78404eb2/zipkin-collector/core/src/main/java/zipkin2/collector/CollectorSampler.java#L68.

As noted in the comment there, the approach may not be statistically perfect, but is likely good enough.
</blockquote></details>

Done! You are totally correct.

nmiyake (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/trace_test.go, line 27 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/65#-Lc7oyw99h8PaRm-6fte:-Lc7pOKI1s68DVNyqvgS:bm733od) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/31d1ad8ef9e0620dc85f917e795a2bb73690d887/witchcraft/trace_test.go#L27)):*

let me know if you have a cleaner way to write this test...
</blockquote></details>

I think this is good -- provides decent coverage of the cases, and it's targeted enough that I don't think it's worth any extra effort to specifically decomp other things (like specifying range of bools without explicitly doing the arrays).

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/65)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

PR #67, opened 2019-04-14
Add inflight request ratelimiter middleware

bmoylan:
An optional middleware for returning 429 when there are too many requests in flight.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)

### DIFF
--- a/integration/ratelimit_test.go
+++ b/integration/ratelimit_test.go
@@ -38,11 +38,8 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 	requireHealthy := func(msg string) {
 		require.Equal(t, string(health.HealthStateHealthy), string(healthComponent.Status()), msg)
 	}
-	requireRepairing := func(msg string) {
-		require.Equal(t, string(health.HealthStateRepairing), string(healthComponent.Status()), msg)
-	}
 
-	limiter := ratelimit.NewInflightLimitMiddleware(2, ratelimit.MatchMutating, healthComponent)
+	limiter := ratelimit.NewInFlightRequestLimitMiddleware(2, ratelimit.MatchMutating, healthComponent)
 
 	wait, closeWait := context.WithCancel(context.Background())
 	defer closeWait()
@@ -81,11 +78,11 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 		getURL := fmt.Sprintf("https://localhost:%d/%s/get", port, basePath)
 		request, err := http.NewRequest(http.MethodGet, getURL, nil)
 		if err != nil {
-			panic(err)
+			assert.NoError(t, err)
 		}
 		resp, err := client.Do(request.WithContext(ctx))
 		if err != nil {
-			panic(err)
+			assert.NoError(t, err)
 		}
 		return resp
 	}
@@ -93,11 +90,11 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 		postURL := fmt.Sprintf("https://localhost:%d/%s/post", port, basePath)
 		request, err := http.NewRequest(http.MethodPost, postURL, nil)
 		if err != nil {
-			panic(err)
+			assert.NoError(t, err)
 		}
 		resp, err := client.Do(request.WithContext(ctx))
 		if err != nil {
-			panic(err)
+			assert.NoError(t, err)
 		}
 		return resp
 	}
@@ -115,12 +112,9 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 
 	resp3 := doPost()
 	require.Equal(t, http.StatusTooManyRequests, resp3.StatusCode, "expected third POST request to be rate limited")
-	requireRepairing("expected repairing after throttled response")
 
 	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
 	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
-
-	requireRepairing("expected unchanged health after unmatched response")
 
 	// free the waiting requests, which should return 200
 	closeWait()
--- a/integration/ratelimit_test.go
+++ b/integration/ratelimit_test.go
@@ -38,6 +38,9 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 	requireHealthy := func(msg string) {
 		require.Equal(t, string(health.HealthStateHealthy), string(healthComponent.Status()), msg)
 	}
+	requireRepairing := func(msg string) {
+		require.Equal(t, string(health.HealthStateRepairing), string(healthComponent.Status()), msg)
+	}
 
 	limiter := ratelimit.NewInFlightRequestLimitMiddleware(2, ratelimit.MatchMutating, healthComponent)
 
@@ -112,9 +115,12 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 
 	resp3 := doPost()
 	require.Equal(t, http.StatusTooManyRequests, resp3.StatusCode, "expected third POST request to be rate limited")
+	requireRepairing("expected repairing after throttled response")
 
 	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
 	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
+
+	requireRepairing("expected unchanged health after unmatched response")
 
 	// free the waiting requests, which should return 200
 	closeWait()
--- a/integration/ratelimit_test.go
+++ b/integration/ratelimit_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/palantir/witchcraft-go-server/status/reporter"
 	"github.com/palantir/witchcraft-go-server/witchcraft"
 	"github.com/palantir/witchcraft-go-server/witchcraft/ratelimit"
+	"github.com/palantir/witchcraft-go-server/witchcraft/refreshable"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -42,7 +43,7 @@ func TestNewInflightLimitMiddleware(t *testing.T) {
 		require.Equal(t, string(health.HealthStateRepairing), string(healthComponent.Status()), msg)
 	}
 
-	limiter := ratelimit.NewInFlightRequestLimitMiddleware(2, ratelimit.MatchMutating, healthComponent)
+	limiter := ratelimit.NewInFlightRequestLimitMiddleware(refreshable.NewInt(refreshable.NewDefaultRefreshable(2)), ratelimit.MatchMutating, healthComponent)
 
 	wait, closeWait := context.WithCancel(context.Background())
 	defer closeWait()
--- a/integration/ratelimit_test.go
+++ b/integration/ratelimit_test.go
@@ -1,0 +1,141 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/palantir/pkg/httpserver"
+	"github.com/palantir/witchcraft-go-server/conjure/witchcraft/api/health"
+	"github.com/palantir/witchcraft-go-server/status/reporter"
+	"github.com/palantir/witchcraft-go-server/witchcraft"
+	"github.com/palantir/witchcraft-go-server/witchcraft/ratelimit"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewInflightLimitMiddleware(t *testing.T) {
+	healthReporter := reporter.NewHealthReporter()
+	healthComponent, err := healthReporter.InitializeHealthComponent("INFLIGHT_MUTATING_REQUESTS")
+	require.NoError(t, err)
+	requireHealthy := func(msg string) {
+		require.Equal(t, string(health.HealthStateHealthy), string(healthComponent.Status()), msg)
+	}
+	requireRepairing := func(msg string) {
+		require.Equal(t, string(health.HealthStateRepairing), string(healthComponent.Status()), msg)
+	}
+
+	limiter := ratelimit.NewInflightLimitMiddleware(2, ratelimit.MatchMutating, healthComponent)
+
+	wait, closeWait := context.WithCancel(context.Background())
+	defer closeWait()
+
+	initFn := func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
+		info.Router.RootRouter().AddRouteHandlerMiddleware(limiter)
+		if err := info.Router.Get("/get", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			rw.WriteHeader(http.StatusOK)
+		})); err != nil {
+			return nil, err
+		}
+		if err := info.Router.Post("/post", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			<-wait.Done()
+			rw.WriteHeader(http.StatusOK)
+		})); err != nil {
+			return nil, err
+		}
+
+		return nil, nil
+	}
+
+	port, err := httpserver.AvailablePort()
+	require.NoError(t, err)
+	server, serverErr, cleanup := createAndRunCustomTestServer(t, port, port, initFn, ioutil.Discard, createTestServer)
+	defer func() {
+		require.NoError(t, server.Close())
+	}()
+	defer cleanup()
+
+	client := testServerClient()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	doGet := func() *http.Response {
+		getURL := fmt.Sprintf("https://localhost:%d/%s/get", port, basePath)
+		request, err := http.NewRequest(http.MethodGet, getURL, nil)
+		if err != nil {
+			panic(err)
+		}
+		resp, err := client.Do(request.WithContext(ctx))
+		if err != nil {
+			panic(err)
+		}
+		return resp
+	}
+	doPost := func() *http.Response {
+		postURL := fmt.Sprintf("https://localhost:%d/%s/post", port, basePath)
+		request, err := http.NewRequest(http.MethodPost, postURL, nil)
+		if err != nil {
+			panic(err)
+		}
+		resp, err := client.Do(request.WithContext(ctx))
+		if err != nil {
+			panic(err)
+		}
+		return resp
+	}
+
+	// Fill up rate limit
+	resp1c := make(chan *http.Response)
+	requireHealthy("expected healthy before the first request")
+	go func() { resp1c <- doPost() }()
+	resp2c := make(chan *http.Response)
+	requireHealthy("expected healthy before the second request")
+	go func() { resp2c <- doPost() }()
+
+	time.Sleep(time.Millisecond) // let things settle
+	requireHealthy("expected healthy before the third request")
+
+	resp3 := doPost()
+	require.Equal(t, http.StatusTooManyRequests, resp3.StatusCode, "expected third POST request to be rate limited")
+	requireRepairing("expected repairing after throttled response")
+
+	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
+	require.Equal(t, http.StatusOK, doGet().StatusCode, "expected get request to be successful")
+
+	requireRepairing("expected unchanged health after unmatched response")
+
+	// free the waiting requests, which should return 200
+	closeWait()
+	resp1 := <-resp1c
+	assert.Equal(t, http.StatusOK, resp1.StatusCode, "expected blocked request 1 to return 200 when unblocked")
+	resp2 := <-resp2c
+	assert.Equal(t, http.StatusOK, resp2.StatusCode, "expected blocked request 2 to return 200 when unblocked")
+
+	// we should now be unblocked and healthy
+	require.Equal(t, http.StatusOK, doPost().StatusCode, "expected fourth POST request to be unblocked")
+	requireHealthy("expected healthy after accepted request")
+
+	select {
+	case err := <-serverErr:
+		require.NoError(t, err)
+	default:
+	}
+}
--- a/witchcraft/ratelimit/middleware.go
+++ b/witchcraft/ratelimit/middleware.go
@@ -15,6 +15,7 @@
 package ratelimit
 
 import (
+	"context"
 	"net/http"
 	"sync/atomic"
 
@@ -26,7 +27,7 @@ import (
 
 type MatchFunc func(req *http.Request, vals wrouter.RequestVals) bool
 
-var MatchReadonly MatchFunc = func(req *http.Request, vals wrouter.RequestVals) bool {
+var MatchReadOnly MatchFunc = func(req *http.Request, vals wrouter.RequestVals) bool {
 	return req.Method == http.MethodGet || req.Method == http.MethodHead || req.Method == http.MethodOptions
 }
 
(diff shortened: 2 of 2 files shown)


nmiyake (changes requested):
On-board with the goal, but some questions/feedback on the initial implementation

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-:-LcWybzA4SVyb-XYtB6m:b-pr2uzf)* status: all files reviewed, 8 unresolved discussions (waiting on @bmoylan and @chewr)

___
*[witchcraft/ratelimit/middleware.go, line 29 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWuNViA6r_GXF5x_2q:-LcWuNViA6r_GXF5x_2r:br0z8cw) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L29)):*
> ```Go
> type MatchFunc func(req *http.Request, vals wrouter.RequestVals) bool
> 
> var MatchReadonly MatchFunc = func(req *http.Request, vals wrouter.RequestVals) bool {
> ```

nit: prefer `ReadOnly`, as the word is compound ("read-only")
___
*[witchcraft/ratelimit/middleware.go, line 37 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWuWd70ClylhZWz7BN:-LcWuWd70ClylhZWz7BO:buw5jeg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L37)):*
> ```Go
> }
> 
> // NewInflightLimitMiddleware returns a middleware which counts and limits the number of
> ```

Would like the name to convey that it limits requests. I feel like one of:

* `NewRequestLimitMiddleware`
* `NewActiveRequestLimitMiddleware`
* `NewInFlightRequestLimitMiddleware`

might make the meaning clearer.

If we do use in-flight, as a nit, would prefer `InFlight` as capitalization and `in-flight` as documentation usage. I know that some sources list "inflight" as a valid word, but I see "in-flight" as a pretty common usage as well (including in Go docs: https://golang.org/pkg/net/http/#Transport.CancelRequest), and feel that compound adjectives should generally be hyphenated.
___
*[witchcraft/ratelimit/middleware.go, line 38 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWtBOy6EIfy7JeSXDQ:-LcWtBOy6EIfy7JeSXDR:b-8g99df) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L38)):*
> ```Go
> 
> // NewInflightLimitMiddleware returns a middleware which counts and limits the number of
> // inflight requests that match a the provided MatchFunc filter. If health is non-nil,
> ```

nit: name used here ("health") and parameter name ("healtcheck") should match.
___
*[witchcraft/ratelimit/middleware.go, line 39 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWtLlj6SoF4_P8wTJN:-LcWtLlj6SoF4_P8wTJO:botwkfi) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L39)):*
> ```Go
> // NewInflightLimitMiddleware returns a middleware which counts and limits the number of
> // inflight requests that match a the provided MatchFunc filter. If health is non-nil,
> // it will be set to REPAIRING when the middleware returns StatusTooManyRequests (429),
> ```

I think the documentation should also note the behavior that:

* Health is set to "Healthy" on construction by this function
* Health is set to "Healthy" 
___
*[witchcraft/ratelimit/middleware.go, line 49 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWs_w8Ay9fYUR1Hhe7:-LcWs_w8Ay9fYUR1Hhe8:b1yfb1z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L49)):*
> ```Go
> 		Health:  healthcheck,
> 	}
> 	healthcheck.Healthy()
> ```

Needs a `nil` check to ensure that this won't panic if `healthcheck` is `nil` (function documentation implies that it can be `nil`, and all accesses in `ServeHTTP` are checked for `nil`)
___
*[witchcraft/ratelimit/middleware.go, line 62 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWxgOE4aGRyT3Mmp6X:-LcWxgOE4aGRyT3Mmp6Y:bbkm19) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L62)):*
> ```Go
> 
> func (l *limiter) ServeHTTP(rw http.ResponseWriter, req *http.Request, reqVals wrouter.RequestVals, next wrouter.RouteRequestHandler) {
> 	if l.Matches(req, reqVals) {
> ```

This will panic if the `MatchFunc` provided by the user in the constructor is `nil`.

Should either change the constructor to return an `error` and reject `nil` values, or change behavior to something like saying that a `nil` match function will match all requests (and document)
___
*[witchcraft/ratelimit/middleware.go, line 68 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWvygw6TmR4_QCXnLH:-LcWvygw6TmR4_QCXnLI:bhk9361) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L68)):*
> ```Go
> 		if current > l.Limit {
> 			msg := "Throttling due to too many inflight requests"
> 			if l.Health != nil && l.Health.Status() != health.HealthStateRepairing {
> ```

In this check and the previous check, is there any reason to perform the check that the state isn't already the target state?

Also, I don't believe that the get (`Status()`) and set (`SetHealth()`) functions are thread-safe themselves, so the check-and-set here and below where health is set could race (we may need to be somewhat concerned about this given that handlers are very much expected to run concurrently in multiple routines)
___
*[witchcraft/ratelimit/middleware.go, line 82 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWwRTDCoucDRcSOlG3:-LcWwRTDCoucDRcSOlG4:b-wr1ykj) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L82)):*
> ```Go
> 		// Mark ourselves healthy!
> 		if l.Health != nil && l.Health.Status() != health.HealthStateHealthy {
> 			l.Health.Healthy()
> ```

Big-picture question: what (if any) are the expectations for the health state after a sequence of actions?

Here's a possible edge case:

* Assume rate limit is 10 and server starts with 0 requests.
* Server gets 11 requests that match the filter, where first 10 are long-running
* When 11th request comes in, counter is incremented to 11 and health is set to "REPAIRING"
* All 11 requests complete and exit, and counter goes down to 0

In this scenario, the health will be marked as "REPAIRING" until the next matching request.

There probably isn't too much down-side here, but just wanted to clarify the contract/expectation.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @chewr and @nmiyake)

___
*[integration/ratelimit_test.go, line 118 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbnaBA1ul3bxN20CBW:-LcbnaBA1ul3bxN20CBX:bwzwm07) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/integration/ratelimit_test.go#L118)):*
> ```Go
> 	resp3 := doPost()
> 	require.Equal(t, http.StatusTooManyRequests, resp3.StatusCode, "expected third POST request to be rate limited")
> 	requireRepairing("expected repairing after throttled response")
> ```

I had to remove these checks because the middleware is so fast that it doesn't leave any leftover state to check after the request returns... ideas welcome for how to better test the health interaction
___
*[witchcraft/ratelimit/middleware.go, line 29 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWuNViA6r_GXF5x_2q:-LcZP04gBWm5xcWUtL8N:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L29)):*

nit: prefer `ReadOnly`, as the word is compound ("read-only")
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 37 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWuWd70ClylhZWz7BN:-LcZUQFv3jPGandVW7co:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L37)):*

Would like the name to convey that it limits requests. I feel like one of:

* `NewRequestLimitMiddleware`
* `NewActiveRequestLimitMiddleware`
* `NewInFlightRequestLimitMiddleware`

might make the meaning clearer.

If we do use in-flight, as a nit, would prefer `InFlight` as capitalization and `in-flight` as documentation usage. I know that some sources list "inflight" as a valid word, but I see "in-flight" as a pretty common usage as well (including in Go docs: https://golang.org/pkg/net/http/#Transport.CancelRequest), and feel that compound adjectives should generally be hyphenated.
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 38 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWtBOy6EIfy7JeSXDQ:-LcZPC9V4DNQ3PA_qcNi:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L38)):*

nit: name used here ("health") and parameter name ("healtcheck") should match.
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 39 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWtLlj6SoF4_P8wTJN:-LcZV0X37KcEx8hkDnpW:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L39)):*

I think the documentation should also note the behavior that:

* Health is set to "Healthy" on construction by this function
* Health is set to "Healthy" 
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 49 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWs_w8Ay9fYUR1Hhe7:-LcZV1C29tbBoFqSqiSM:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L49)):*

Needs a `nil` check to ensure that this won't panic if `healthcheck` is `nil` (function documentation implies that it can be `nil`, and all accesses in `ServeHTTP` are checked for `nil`)
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 62 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWxgOE4aGRyT3Mmp6X:-LcZV2JT3fJBZxUs0iZz:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L62)):*

This will panic if the `MatchFunc` provided by the user in the constructor is `nil`.

Should either change the constructor to return an `error` and reject `nil` values, or change behavior to something like saying that a `nil` match function will match all requests (and document)
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 68 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWvygw6TmR4_QCXnLH:-LcZVfR76ev6LO9myxZO:bf7c12u) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L68)):*

In this check and the previous check, is there any reason to perform the check that the state isn't already the target state?

Also, I don't believe that the get (`Status()`) and set (`SetHealth()`) functions are thread-safe themselves, so the check-and-set here and below where health is set could race (we may need to be somewhat concerned about this given that handlers are very much expected to run concurrently in multiple routines)
</blockquote></details>

I did it this way because it's guarded by a RWLock, so the read path should not contend for the lock and there should be relatively few writers. https://github.com/palantir/witchcraft-go-server/blob/develop/status/reporter/reporter.go#L138-L148
___
*[witchcraft/ratelimit/middleware.go, line 82 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWwRTDCoucDRcSOlG3:-LcZVwNg-9TWwmIuAd38:bo7dm68) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L82)):*

Big-picture question: what (if any) are the expectations for the health state after a sequence of actions?

Here's a possible edge case:

* Assume rate limit is 10 and server starts with 0 requests.
* Server gets 11 requests that match the filter, where first 10 are long-running
* When 11th request comes in, counter is incremented to 11 and health is set to "REPAIRING"
* All 11 requests complete and exit, and counter goes down to 0

In this scenario, the health will be marked as "REPAIRING" until the next matching request.

There probably isn't too much down-side here, but just wanted to clarify the contract/expectation.
</blockquote></details>

Great question. I've updated the logic to update the health status in the deferred function (`decrement`) so that each request is more atomically net-zero.

nmiyake:
___
*[witchcraft/ratelimit/middleware.go, line 68 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcWvygw6TmR4_QCXnLH:-LcbqBS64zgoFqtn9n0u:b-xju14) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/witchcraft/ratelimit/middleware.go#L68)):*

I did it this way because it's guarded by a RWLock, so the read path should not contend for the lock and there should be relatively few writers. https://github.com/palantir/witchcraft-go-server/blob/develop/status/reporter/reporter.go#L138-L148
</blockquote></details>

Oh great, my mistake -- didn't know that the reporter itself was implemented in a thread-safe way. Great!

nmiyake (changes requested):
Thanks! Main issue is one flagged by your issue with writing the test -- I think that the condition for triggering repairing state may need to be shifted by one to provide better semantics/signaling.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-:-LcbvFIY8g7Hb0Nml_WN:btvuumn)* status: all files reviewed, 6 unresolved discussions (waiting on @bmoylan and @chewr)

___
*[integration/ratelimit_test.go, line 118 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbnaBA1ul3bxN20CBW:-Lcbsys77W4pjhhik-Xq:bvvh88z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/integration/ratelimit_test.go#L118)):*

I had to remove these checks because the middleware is so fast that it doesn't leave any leftover state to check after the request returns... ideas welcome for how to better test the health interaction
</blockquote></details>

I think this is actually exposing a possible issue with the underlying logic!

As you point out, whenever a request that causes throttling comes in, the function immediately returns an error response and terminates, and sets the service back to "healthy". I think the logic should probably be changed such that the state is only set to healthy when the connection count is 1 less than the limit (rather than the limit itself).

Consider the case where the limit is 10 and we have 10 long-running requests that are effectively hogging all of the connections. In this case, the status would be "healthy", but every subsequent request that comes in will be throttled, set status to "repairing" for a fraction of a second, but then decrement and go back to "healthy" again.

I think the semantic should instead be that the service is "repairing" if it's at capacity (if subsequent matching requests will be throttled) -- so, if the limit is 10 and there are 10 active connections, the status should be "repairing".

If this logic is implemented, then testing should be pretty easy as well -- just hold open a number of connections equal to the limit.
___
*[witchcraft/ratelimit/middleware.go, line 50 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbtClg3TK3YoOu6zis:-LcbtClg3TK3YoOu6zit:bhyxjlg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L50)):*
> ```Go
> func NewInFlightRequestLimitMiddleware(limit int64, matches MatchFunc, healthcheck reporter.HealthComponent) wrouter.RouteHandlerMiddleware {
> 	l := &limiter{
> 		Limit:   limit,
> ```

If provided limit is <= 0, do we care, or should we explicitly set it to 0 (or some other minimum value like "1")?
___
*[witchcraft/ratelimit/middleware.go, line 68 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbsTeD7doLP69n0Rkt:-LcbsTeD7doLP69n0Rku:b-lxvw9s) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L68)):*
> ```Go
> }
> 
> const inFlightThrottledMessage = "Throttling due to too many inflight requests"
> ```

Can we make this "in-flight"?
___
*[witchcraft/ratelimit/middleware.go, line 84 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbrnGY68t_NhQ9O5Al:-LcbrnGY68t_NhQ9O5Am:b5y8v09) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L84)):*
> ```Go
> }
> 
> func (l *limiter) increment(ctx context.Context) bool {
> ```

Can you add a comment (or name the return value) to indicate what the return value means?
___
*[witchcraft/ratelimit/middleware.go, line 86 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbruhQBJHCDXBC34-1:-LcbruhQBJHCDXBC34-2:b-q3muxn) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L86)):*
> ```Go
> func (l *limiter) increment(ctx context.Context) bool {
> 	current := atomic.AddInt64(&l.current, 1)
> 	if current > l.Limit {
> ```

nit: can we invert the conditional to keep the bulk of the logic at a higher level?

```go
if current <= l.Limit {
     return false
}
// other logic
return true
```
___
*[witchcraft/ratelimit/middleware.go, line 102 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-Lcbv6CM04BnekJqhA81:-LcbsK2v3s2Zzi9ZnHDB:btt1am8) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L102)):*
> ```Go
> func (l *limiter) decrement(ctx context.Context) {
> 	current := atomic.AddInt64(&l.current, -1)
> 	if current <= l.Limit {
> ```

nit: since we're not doing anything in either of the "else" cases,  you can combine this check with the rest using an `&&` if you want to do so.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @bmoylan, @chewr, and @nmiyake)

___
*[integration/ratelimit_test.go, line 118 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbnaBA1ul3bxN20CBW:-LccFSDo4MS4t51dpCh-:b70wwph) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fc70fb4be24545f0c7bc946a6a04fda2e8af4435/integration/ratelimit_test.go#L118)):*

I think this is actually exposing a possible issue with the underlying logic!

As you point out, whenever a request that causes throttling comes in, the function immediately returns an error response and terminates, and sets the service back to "healthy". I think the logic should probably be changed such that the state is only set to healthy when the connection count is 1 less than the limit (rather than the limit itself).

Consider the case where the limit is 10 and we have 10 long-running requests that are effectively hogging all of the connections. In this case, the status would be "healthy", but every subsequent request that comes in will be throttled, set status to "repairing" for a fraction of a second, but then decrement and go back to "healthy" again.

I think the semantic should instead be that the service is "repairing" if it's at capacity (if subsequent matching requests will be throttled) -- so, if the limit is 10 and there are 10 active connections, the status should be "repairing".

If this logic is implemented, then testing should be pretty easy as well -- just hold open a number of connections equal to the limit.
</blockquote></details>

you are right yet again! this makes the behavior more predictable (I think) and also fixes my tests 
___
*[witchcraft/ratelimit/middleware.go, line 68 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbsTeD7doLP69n0Rkt:-LccFedL4LqOFWwyV-c6:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L68)):*

Can we make this "in-flight"?
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 84 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbrnGY68t_NhQ9O5Al:-LccFg7a53-kT5XQyviG:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L84)):*

Can you add a comment (or name the return value) to indicate what the return value means?
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 86 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbruhQBJHCDXBC34-1:-LccG2JK0mvgiCpyEY5k:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L86)):*

nit: can we invert the conditional to keep the bulk of the logic at a higher level?

```go
if current <= l.Limit {
     return false
}
// other logic
return true
```
</blockquote></details>

Done.
___
*[witchcraft/ratelimit/middleware.go, line 102 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-Lcbv6CM04BnekJqhA81:-LccG1JfCP9PtVxIYpAj:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L102)):*

nit: since we're not doing anything in either of the "else" cases,  you can combine this check with the rest using an `&&` if you want to do so.
</blockquote></details>

Done.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @chewr and @nmiyake)

___
*[witchcraft/ratelimit/middleware.go, line 50 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbtClg3TK3YoOu6zis:-LccGqX976d6NHVPN8jy:bxec1xm) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L50)):*

If provided limit is <= 0, do we care, or should we explicitly set it to 0 (or some other minimum value like "1")?
</blockquote></details>

Good question, right now it would block all requests (which is maybe desirable in some weird scenario) but doesn't feel like there's a lot of safety

On a different note, do you think the limit should be a `refreshable.Int` so it's easily live-reloaded?

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: all files reviewed, 2 unresolved discussions (waiting on @bmoylan, @chewr, and @nmiyake)

___
*[witchcraft/ratelimit/middleware.go, line 50 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbtClg3TK3YoOu6zis:-LccHrYM6_NgklDHfZYI:bll2ov4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L50)):*

Good question, right now it would block all requests (which is maybe desirable in some weird scenario) but doesn't feel like there's a lot of safety

On a different note, do you think the limit should be a `refreshable.Int` so it's easily live-reloaded?
</blockquote></details>

Yeah, I don't think a negative limit makes logical sense, so let's just do a check for that and set to 0 if that's the case. A limit of 0 doesn't make too much sense either, but at least that's logically consistent.

Yes, I think using a refreshable for this value could make sense here.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @chewr and @nmiyake)

___
*[witchcraft/ratelimit/middleware.go, line 50 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/67#-LcbtClg3TK3YoOu6zis:-LccKTmI0p8pjqq-Ni1O:b-896fja) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9224e5c34bc993af4f28e5cb2a1c8f9e2e024892/witchcraft/ratelimit/middleware.go#L50)):*

Yeah, I don't think a negative limit makes logical sense, so let's just do a check for that and set to 0 if that's the case. A limit of 0 doesn't make too much sense either, but at least that's logically consistent.

Yes, I think using a refreshable for this value could make sense here.
</blockquote></details>

Done!

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/67)* status: all files reviewed, 1 unresolved discussion (waiting on @chewr and @nmiyake)

bmoylan:
woo! thanks for the thorough review!

PR #74, opened 2019-05-05
 Fix wresource metric tags

### LINKED ISSUE #71: wresource metrics middleware incompatible with pkg/metrics 0.10.2
See failing build here: https://github.com/palantir/witchcraft-go-server/compare/bm/bump-pkg?expand=1

It looks like we were depending on the behavior before https://github.com/palantir/pkg/pull/153.

We assume that the request context will have tags here: https://github.com/palantir/witchcraft-go-server/blob/v1.6.1/witchcraft/internal/middleware/request.go#L166

This worked previously by setting them here: https://github.com/palantir/witchcraft-go-server/blob/v1.6.1/witchcraft/wresource/resource.go#L92

bmoylan:
Fixes #71

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/74)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -90,8 +90,8 @@
     "transform",
   ]
   pruneopts = "UT"
-  revision = "332184900c68094416680529ecbe2592135374b6"
-  version = "0.10.2"
+  revision = "50eac0b01074bf9f02663dc24542be657815b120"
+  version = "0.10.3"
 
 [[projects]]
   branch = "develop"
@@ -105,7 +105,7 @@
   revision = "adb1d93c5741a300abbcce7883ad61f825a4f1a7"
 
 [[projects]]
-  digest = "1:8e1a7e54bcc3948d22a23b1353845e25e13271bc033c7c5b90aeaee2e7f5a755"
+  digest = "1:fcea968e9514fce49783f16c0da2f526c04558bb9735c7802dd8019de80a4c65"
   name = "github.com/palantir/witchcraft-go-logging"
   packages = [
     "conjure/witchcraft/api/logging",
@@ -125,8 +125,8 @@
     "wlog/trclog/trc1log",
   ]
   pruneopts = "UT"
-  revision = "d6e84e65e4cfe1080e8dc1010fc4707ddc9d353c"
-  version = "v1.3.0"
+  revision = "cb1c1596f2463de3c140dac89263c73c16109226"
+  version = "v1.3.4"
 
 [[projects]]
   branch = "develop"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -75,7 +75,7 @@
   version = "1.0.0"
 
 [[projects]]
-  digest = "1:2c19058cc056908dd5a08b24b8f0d0582b4f98ff5b76f1c934573e854fc66450"
+  digest = "1:75aa9f999a195c38150f8f96f473e78f962838b5232cf072c8989e7f07b009e4"
   name = "github.com/palantir/pkg"
   packages = [
     "datetime",
@@ -90,8 +90,8 @@
     "transform",
   ]
   pruneopts = "UT"
-  revision = "7fd8d6e85a713f1ec5376c81b97c7ed1033c9816"
-  version = "0.10.0"
+  revision = "332184900c68094416680529ecbe2592135374b6"
+  version = "0.10.2"
 
 [[projects]]
   branch = "develop"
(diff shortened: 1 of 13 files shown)


nmiyake (changes requested):
Thanks for implementing this! Just to make sure I understand this correctly:

* Metrics are now a concept in `wrouter`, and can be provided as a parameter during route registration
* Any metrics that are provided when registering a route will be logged as part of the request log output for that endpoint
* There is no way to add any tags to what is logged by the request logger except when registering the route (handler code can't modify/add metrics for the request logger)

I believe this is the case based on my reading of the PR, but just wanted to confirm.

If so, I think this is a pretty reasonable trade-off -- the only thing that may not be totally ideal is adding the metric import to `wrouter`, but it is a core concept and the API is pretty locked down so I don't think it's a major issue (and think it's probably better than any of the alternatives to implement this).

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-:-LeDlgLFDZBtziSea1Va:bomv39k)* status: all files reviewed, 2 unresolved discussions (waiting on @bmoylan)

___
*[wrouter/route_params.go, line 101 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-LeDkbQVCDsCeaoCrehp:-LeDkbQVCDsCeaoCrehq:b59sorw) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a475c3a6e93483550030c817f28ad99377c1899b/wrouter/route_params.go#L101)):*
> ```Go
> 
> 		perms := p.perms()
> 		if perms != nil {
> ```

This is a fix for a separate issue, right?
___
*[wrouter/route_params.go, line 134 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-LeDhyBy4xHcoKehRkx2:-LeDhyBy4xHcoKehRkx3:bh7n9as) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a475c3a6e93483550030c817f28ad99377c1899b/wrouter/route_params.go#L134)):*
> ```Go
> 	var tags metrics.Tags
> 	for _, param := range params {
> 		if t := param.metricTags(); len(t) > 0 {
> ```

nit: don't think this check is necessary -- `tags = append(tags, param.metricTags()...)` should provide equivalent behavior (if the return value is `nil` or 0-length, the append is a no-op)

bmoylan (commented):
Yes exactly (sorry for not giving a summary up-front). The slightly weird thing is that `wrouter` has a metrics import and a MetricTags param but doesn't actually use the content and just plumbs the param values to the RequestVals provided to the middleware. Both the tag construction and the actual metrics live in `wresource`.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-:-LeDrBnS2cJTLV3tKBOU:buz6t7d)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[wrouter/route_params.go, line 101 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-LeDkbQVCDsCeaoCrehp:-LeDnoIW-9YYFCRhUiwJ:bz23m7k) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a475c3a6e93483550030c817f28ad99377c1899b/wrouter/route_params.go#L101)):*

This is a fix for a separate issue, right?
</blockquote></details>

In the new `requestMetricTagsFunc` implementation of RouteParam `perms()` returns nil, so this panics without the nil check
___
*[wrouter/route_params.go, line 134 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-LeDhyBy4xHcoKehRkx2:-LeDqjvsBx9V9oooIa6E:b-taqnoy) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a475c3a6e93483550030c817f28ad99377c1899b/wrouter/route_params.go#L134)):*

nit: don't think this check is necessary -- `tags = append(tags, param.metricTags()...)` should provide equivalent behavior (if the return value is `nil` or 0-length, the append is a no-op)
</blockquote></details>

Yep! Done.

nmiyake (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/74#-:-LeDtXFm-xWcxNWO_qCI:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #82, opened 2019-05-31
Return HealthCheckResult in periodic checks instead of individual fields

gregakinman:
And prepend periodic messages on the maybe-present result message, as a result.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/82)

### DIFF
--- a/status/health/periodic/source.go
+++ b/status/health/periodic/source.go
@@ -25,7 +25,7 @@ import (
 	"github.com/palantir/witchcraft-go-server/status"
 )
 
-type CheckFunc func(ctx context.Context) (state health.HealthState, params map[string]interface{})
+type CheckFunc func(ctx context.Context) *health.HealthCheckResult
 
 type Source struct {
 	Checks map[health.CheckType]CheckFunc
@@ -92,10 +92,10 @@ func (h *healthCheckSource) HealthStatus(ctx context.Context) health.HealthStatu
 			result = *checkState.lastSuccess
 		case time.Since(checkState.lastResultTime) <= h.gracePeriod:
 			result = *checkState.lastResult
-			result.Message = stringPtr(fmt.Sprintf("No successful checks during %s grace period", h.gracePeriod.String()))
+			result.Message = stringPtr(wrap(fmt.Sprintf("No successful checks during %s grace period", h.gracePeriod.String()), result.Message))
 		default:
 			result = *checkState.lastResult
-			result.Message = stringPtr(fmt.Sprintf("No completed checks during %s grace period", h.gracePeriod.String()))
+			result.Message = stringPtr(wrap(fmt.Sprintf("No completed checks during %s grace period", h.gracePeriod.String()), result.Message))
 			// Mark REPAIRING if we were healthy before expiration.
 			if result.State == health.HealthStateHealthy {
 				result.State = health.HealthStateRepairing
@@ -133,15 +133,10 @@ func (h *healthCheckSource) doPoll(ctx context.Context) {
 
 	// Run checks
 	resultsWithTimes := make([]resultWithTime, 0, len(h.source.Checks))
-	for checkType, check := range h.source.Checks {
-		state, params := check(ctx)
+	for _, check := range h.source.Checks {
 		resultsWithTimes = append(resultsWithTimes, resultWithTime{
-			time: time.Now(),
-			result: &health.HealthCheckResult{
-				Type:   checkType,
-				State:  state,
-				Params: params,
-			},
+			time:   time.Now(),
+			result: check(ctx),
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


gregakinman:
@bmoylan

bmoylan:
___
*[status/health/periodic/source.go, line 31 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-LgFPBhn-Ph43oIYbNL3:-LgFPBhoEunJoIsNeEfZ:b-2di0dc) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/80fb9be4702b87da3b58321cfcc9f92523546254/status/health/periodic/source.go#L31)):*
> ```Go
> 
> type Source struct {
> 	Checks map[health.CheckType]CheckFunc
> ```

Does this still make sense as a map if the `HealthCheckResult` contains the CheckType? What do we do if they happen to disagree?

gregakinman (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/82)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[status/health/periodic/source.go, line 31 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-LgFPBhn-Ph43oIYbNL3:-LgFPmfZBpsq-dPjXZVL:btfszap) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/80fb9be4702b87da3b58321cfcc9f92523546254/status/health/periodic/source.go#L31)):*

Does this still make sense as a map if the `HealthCheckResult` contains the CheckType? What do we do if they happen to disagree?
</blockquote></details>

It does still make sense as a map, since we need a way of a priori knowing which checks are configured to run for the time period before any checks have run yet (so we can say they haven't run yet, rather than just returning an empty `HealthStatus`). If they don't agree (if the provided function returns a result that has the wrong check type), we don't try to overwrite it for them, we just return what the client gave us.

If anything doesn't make sense, it's `HealthCheckResult` containing the `CheckType`, given that the SLS spec requires `HealthCheckResult` to be keyed by `CheckType` in the response to the health endpoint. So I think, in the end, this duplication kind of just falls out of how the spec is.

bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/82)* status: all files reviewed, 1 unresolved discussion (waiting on @gregakinman)

___
*[status/health/periodic/source.go, line 31 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-LgFPBhn-Ph43oIYbNL3:-LgFUZxy2CHzguNYPHim:bfqo2hq) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/80fb9be4702b87da3b58321cfcc9f92523546254/status/health/periodic/source.go#L31)):*
<details><summary><i>Previously, gregakinman (Greg Kinman) wrote…</i></summary><blockquote>

It does still make sense as a map, since we need a way of a priori knowing which checks are configured to run for the time period before any checks have run yet (so we can say they haven't run yet, rather than just returning an empty `HealthStatus`). If they don't agree (if the provided function returns a result that has the wrong check type), we don't try to overwrite it for them, we just return what the client gave us.

If anything doesn't make sense, it's `HealthCheckResult` containing the `CheckType`, given that the SLS spec requires `HealthCheckResult` to be keyed by `CheckType` in the response to the health endpoint. So I think, in the end, this duplication kind of just falls out of how the spec is.
</blockquote></details>

+1 makes sense
___
*[status/health/periodic/source.go, line 188 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-LgFUhE2CECeFhnqbMNE:-LgFUhE2CECeFhnqbMNF:b-1okox4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/80fb9be4702b87da3b58321cfcc9f92523546254/status/health/periodic/source.go#L188)):*
> ```Go
> }
> 
> func wrap(baseStringPtr *string, prependStr string) string {
> ```

Let's add test(s) with returned checks that have messages

gregakinman (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/82)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[status/health/periodic/source.go, line 188 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-LgFUhE2CECeFhnqbMNE:-LgFWQ0R7dBDyy0eu_3Q:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/80fb9be4702b87da3b58321cfcc9f92523546254/status/health/periodic/source.go#L188)):*

Let's add test(s) with returned checks that have messages
</blockquote></details>

Done.

bmoylan (approved):
:lgtm:

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/82#-:-LgFXBnHEOVU86xCPayE:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #94, opened 2019-07-02
Add `periodic.Option` type for running initial poll before waiting.

jamesross3:
Sometimes it's undesirable to wait for a poll period to run a periodic health check - this introduces the `periodic.Option` type and the `WithInitialPoll()` option to enable calling the `poll` method without waiting for the poll period.

This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/94)

### DIFF
--- a/status/health/periodic/option.go
+++ b/status/health/periodic/option.go
@@ -1,3 +1,17 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package periodic
 
 type Option interface {
--- a/status/health/periodic/option.go
+++ b/status/health/periodic/option.go
@@ -6,7 +6,7 @@ type Option interface {
 
 type optionFn func(source *healthCheckSource)
 
-func(fn optionFn) apply(source *healthCheckSource) {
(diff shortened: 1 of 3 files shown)


bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/94)* status: all files reviewed, 1 unresolved discussion (waiting on @jamesross3)

___
*[status/health/periodic/source.go, line 116 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/94#-LjHqakMD2YItDM5I64j:-LjHqakMD2YItDM5I64k:bwekkuq) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/8e26d1e4003e4b6e52fd443dbb35bcf677afab04/status/health/periodic/source.go#L116)):*
> ```Go
> 
> func (h *healthCheckSource) runPoll(ctx context.Context) {
> 	if h.initialPoll {
> ```

Let's do this after we create the ticker, so that it does not change the interval between calls (`retryInterval` minus duration of `doPoll`)

bmoylan:
___
*[status/health/periodic/option.go, line 27 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/94#-LjILzI_1IimZByrKp1k:-LjILzI_1IimZByrKp1l:b-w0lxkg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/43af893bafece0df5b83c4efbf63331e03da55d4/status/health/periodic/option.go#L27)):*
> ```Go
> }
> 
> func WithInitialPoll() Option {
> ```

can you add godoc explaining what this does?

bmoylan (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/94)* status: :shipit: complete! all files reviewed, all discussions resolved

ashrayjain:
@bmoylan is there a good reason for this not being the default behaviour?

bmoylan:
@ashrayjain hm honestly I can't think of a good one off the top of my head. This already runs in a background routine so eagerly running a poll probably doesn't hurt much. 

Note that if we make this the default behavior, it will still have some `REPAIRING` time during the duration of this initial poll.

PR #95, opened 2019-07-22
Refactor health reporter components

andybradshaw:
## Before this PR:
  * Use of a health component after calling `UnregisterHealthComponent` had some undesirable behavior. The health component still had a reference to the reporter that created it, and could modify the reporter `currentStatus`, even after it had been removed during unregistration. Also, it could cause panics from calls to a components `Status` call, with a confusing error message that the component had never been registered.

## After this PR:
  * Health components are more independent from the health reporter, storing all of the data necessary to create their own health check result. Instead of storing state in addition to the health components, the reporter only stores the components and lets them track their state independently. This prevents both of the behaviors above.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/95)

### DIFF
--- a/status/reporter/component.go
+++ b/status/reporter/component.go
@@ -70,11 +70,12 @@ func (r *healthComponent) Status() health.HealthState {
 
 // Returns the entire HealthCheckResult for the component
 func (r *healthComponent) GetHealthCheck() health.HealthCheckResult {
-	var message string
+	var message *string
 	params := make(map[string]interface{}, len(r.params))
 
 	if r.message != nil {
-		message = *r.message
+		messageCopy := *r.message
+		message = &messageCopy
 	}
 	for key, value := range r.params {
 		params[key] = value
@@ -83,7 +84,7 @@ func (r *healthComponent) GetHealthCheck() health.HealthCheckResult {
 	return health.HealthCheckResult{
 		Type:    r.name,
 		State:   r.state,
-		Message: &message,
+		Message: message,
 		Params:  params,
 	}
 }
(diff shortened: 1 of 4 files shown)


andybradshaw:
@TDAbboud for SA

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/95)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andybradshaw)

___
*[status/reporter/component.go, line 59 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/95#-Lk_xZC_9EKpt0CwXOfA:-Lk_xZC_9EKpt0CwXOfB:b-mo3gm2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/fb7fa2354f1c6324f49b0d1ffbe8344fb05cbe58/status/reporter/component.go#L59)):*
> ```Go
> 
> func (r *healthComponent) SetHealth(healthState health.HealthState, message *string, params map[string]interface{}) {
> 	r.reporter.setHealthCheck(r.name, health.HealthCheckResult{
> ```

It looks like the prior implementation had some concurrency guarantees, but don't see the same level of protection in the new implementation (for example, concurrent call of `SetHealth` and `GetHealthCheck`). Is this intentional/is there a reason these protections aren't needed in the new implementation?

nmiyake (approved):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/95)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #97, opened 2019-07-25
Feature: add ability to configure logger file provider

### LINKED ISSUE #96: Provide mechanism to share logger output locations
`witchcraft-go-server` has a notion of logging and is opinionated on output -- it decides on the `io.Writer` to use for writing logs based on factors such as configuration (`useConsoleLog`) and environment (`isDocker` or `isJail`), and when it performs file-based logging it does so using a `lumberjack.Logger`.

In some instances, a program may want to perform logging before the `witchcraft.Server` is created or started, but still wants to log in the same manner that the server would.

In order to do so, I propose the following:

1. Define a new `FileWriterProvider` interface that, given a path, returns an `io.Writer` that writes to that path
2. Define a `LumberjackFileWriterProvider` implementation that returns a `lumberjack.Logger` as it is currently defined in `newDefaultLogOutput`
3. Define a `CreateLogWriter(logOutputPath string, logToStdout bool, stdoutWriter io.Writer, fileWriterProvider FileWriterProvider) io.Writer` function
4. Add a `WithFileWriterProvider(FileWriterProvider)` function to the builder for `witchcraft.Server` (if not specified, it will use the `LumberjackFileWriterProvider` by default)

These primitives provide a general way to do a few things:
* It makes it possible to configure the file output writer in code. Right now, a file writer is hard-coded to be a `lumberjack.Logger` with a specific set of parameters. This change would allow file writers to be customized at a code level.
* With the above, a client can provide an implementation of `FileWriterProvider` that caches results and returns the same instance of a `*lumberjack.Logger` for a given path
  * This means that multiple loggers can be instantiated with the same lumberjack logger, and will thus not need to worry about overwriting each other
* With the above, a client can use `CreateLogWriter` to get the `io.Writer` that is appropriate for their environment/setup (log to `stdout` based on environment, etc.)

The only real downside I see is that this does expose a vector for customizing the file-based log output location that did not previously exist. However, this configuration point is purely in code (not in end-user configuration), so I don't think it's a huge risk.

If we really wanted to be stringent about this we could modify the `FileWriterProvider` API to more targeted (for example, only allow it to return a `lumberjack.Logger` and always override portions of that config after it is returned), but I think that's overkill and is over-fitting the specific problem trying to be solved here.

nmiyake:
Adds new concepts like a FileWriterProvider which provides a
mechanism for a witchcraft.Server and invoking code to share
common io.Writer objects for performing logging operations.

Fixes #96

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/97)

### DIFF
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -47,7 +47,8 @@ const (
 
 // createAndRunTestServer returns a running witchcraft.Server that is initialized with simple default configuration in a
 // temporary directory. Returns the server, the path to the temporary directory, a channel that returns the error
-// returned by the server when it stops and a cleanup function that will remove the temporary directory.
+// returned by the server when it stops and a cleanup function that will remove the temporary directory. All logger
+// output is written to the provided logOutputBuffer, or to os.Stdout if no logOutputBuffer is specified.
 func createAndRunTestServer(t *testing.T, initFn witchcraft.InitFunc, logOutputBuffer io.Writer) (server *witchcraft.Server, port int, managementPort int, serverErr <-chan error, cleanup func()) {
 	var err error
 	port, err = httpserver.AvailablePort()
--- a/integration/logger_test.go
+++ b/integration/logger_test.go
@@ -45,7 +45,7 @@ func TestUseLoggerFileWriterProvider(t *testing.T) {
 	// is, if "/.dockerenv" exists), the server logic forces output to the console instead. So, if this path exists,
 	// temporarily rename it during the duration of the test run so that output goes to a file.
 	const (
-		dockerEnvPath = "/.dockerenv"
+		dockerEnvPath       = "/.dockerenv"
 		dockerEnvBackupPath = dockerEnvPath + "-backup"
 	)
 	if _, err := os.Stat(dockerEnvPath); err == nil {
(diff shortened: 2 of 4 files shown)


nmiyake:
Would like to get some feedback on usability/API in addition to the underlying issue.

The main trade-off I see is between keeping the API generic/raw versus making it more specialized.

The current approach is quite generic in that it provides a mechanism to return an `io.Writer`. However, this means that consumers need to create loggers themselves, and also need to provide the path to the logger files themselves (`var/log/service.log`, etc.).

Another possible approach would be to have a function that returns a set of loggers (`service1log`, `metric1log`, etc.). This would remove the need for callers constructing their own loggers and providing path to output. However, this makes the API a bit more fragile (if any of the logger types change in the future, the function would need to change), and for things like the service logger we would need to take extra parameters to specify things like the logger level. We would also want to make it clear that the returned logger doesn't update by default, and that if that behavior is desired the client would need to implement it.

Let me know if you guys have thoughts on this.

nmiyake:
I've put up an alternate implementation at https://github.com/palantir/witchcraft-go-server/pull/99. I think I may prefer that alternate approach as one that has better balance/trade-offs, but open to feedback.

PR #98, opened 2019-07-25
Update Go versions to 1.12.7 and 1.11.12

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/98)


bmoylan (approved):
I think we need to update the required checks in the protected branch settings

nmiyake:
Thanks for the review!

Yup, will do that after this merges (didn't want to do it before, since that would block other PRs from being verified)

PR #99, opened 2019-07-25
Logger provider alternate impl

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/99)

### DIFF
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -47,7 +47,8 @@ const (
 
 // createAndRunTestServer returns a running witchcraft.Server that is initialized with simple default configuration in a
 // temporary directory. Returns the server, the path to the temporary directory, a channel that returns the error
-// returned by the server when it stops and a cleanup function that will remove the temporary directory.
+// returned by the server when it stops and a cleanup function that will remove the temporary directory. All logger
+// output is written to the provided logOutputBuffer, or to os.Stdout if no logOutputBuffer is specified.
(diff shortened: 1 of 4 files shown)


bmoylan (approved):
I like this! I agree this more structured approach (compared to #97) moves some complexity into the library, putting less work on application authors. 

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/99#-:-Lla5uIa9GKSX-MP_i_H:b-ys875x)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[integration/logger_test.go, line 73 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/99#-Lla6BDhBnCbBcZ4tuDh:-Lla6BDhBnCbBcZ4tuDi:b-kzm4xm) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e6778caa3e1d293fb64f6bced177de9c415ffe0e/integration/logger_test.go#L73)):*
> ```Go
> 	require.NoError(t, err)
> 
> 	cachingWriterProvider := witchcraft.NewCachingFileWriterProvider(witchcraft.DefaultFileWriterProvider())
> ```

looks like this is the magic line that applications needing to do this would use. Maybe worth putting a brief section in the readme about it? Or would that just encourage bad behavior that we only really need in special cases :smile:

PR #111, opened 2019-09-20
Add a server.uptime metric

blakehawkins:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/111)

### DIFF
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -201,7 +201,7 @@ func TestEmitMetricsEmptyBlacklist(t *testing.T) {
 		}
 	}
 
-	var seenMyCounter, seenResponseTimer, seenResponseSize, seenRequestSize, seenResponseError bool
+	var seenMyCounter, seenResponseTimer, seenResponseSize, seenRequestSize, seenResponseError, seenUptime bool
 	for _, metricLog := range metricLogs {
 		switch metricLog.MetricName {
 		case "my-counter":
@@ -262,6 +262,10 @@ func TestEmitMetricsEmptyBlacklist(t *testing.T) {
 			assert.NotNil(t, metricLog.Values["5m"])
 			assert.NotNil(t, metricLog.Values["15m"])
 			assert.NotNil(t, metricLog.Values["mean"])
+		case "server.uptime":
+			seenUptime = true
+			assert.Equal(t, "gauge", metricLog.MetricType, "server.uptime metric had incorrect type")
+			assert.NotZero(t, metricLog.Values["value"])
 		default:
 			assert.Fail(t, "unexpected metric encountered: %s", metricLog.MetricName)
 		}
@@ -271,6 +275,7 @@ func TestEmitMetricsEmptyBlacklist(t *testing.T) {
 	assert.True(t, seenRequestSize, "server.request.size metric was not emitted")
 	assert.True(t, seenResponseSize, "server.response.size metric was not emitted")
 	assert.True(t, seenResponseError, "server.response.error metric was not emitted")
+	assert.True(t, seenUptime, "server.uptime metric was not emitted")
 
 	select {
 	case err := <-serverErr:
@@ -322,7 +327,7 @@ func TestMetricTypeValueBlacklist(t *testing.T) {
 		}
 	}
 
-	var seenMyCounter, seenResponseTimer, seenResponseSize, seenRequestSize, seenResponseError bool
+	var seenMyCounter, seenResponseTimer, seenResponseSize, seenRequestSize, seenResponseError, seenUptime bool
 	for _, metricLog := range metricLogs {
 		switch metricLog.MetricName {
 		case "my-counter":
@@ -351,6 +356,10 @@ func TestMetricTypeValueBlacklist(t *testing.T) {
 			seenResponseError = true
 			assert.Equal(t, "meter", metricLog.MetricType, "server.response metric had incorrect type")
 			assert.NotNil(t, metricLog.Values["count"])
+		case "server.uptime":
+			seenUptime = true
+			assert.Equal(t, "gauge", metricLog.MetricType, "server.uptime metric had incorrect type")
+			assert.NotZero(t, metricLog.Values["value"])
 		default:
 			assert.Fail(t, "unexpected metric encountered: %s", metricLog.MetricName)
 		}
@@ -360,6 +369,7 @@ func TestMetricTypeValueBlacklist(t *testing.T) {
 	assert.True(t, seenRequestSize, "server.request.size metric was not emitted")
 	assert.True(t, seenResponseSize, "server.response.size metric was not emitted")
 	assert.True(t, seenResponseError, "server.response.error metric was not emitted")
+	assert.True(t, seenUptime, "server.uptime metric was not emitted")
 
 	select {
 	case err := <-serverErr:
@@ -385,6 +395,7 @@ func TestMetricsBlacklist(t *testing.T) {
 		server.WithMetricsBlacklist(map[string]struct{}{
 			"my-counter":          {},
 			"server.request.size": {},
+			"server.uptime":       {},
 		})
 		return server
 	})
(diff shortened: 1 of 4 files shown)


bmoylan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/111)* status: all files reviewed, 5 unresolved discussions (waiting on @blakehawkins)

*[a discussion](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpSMJuJ4R8VRhXmduI0:-LpSMJuJ4R8VRhXmduI1:bwpurdc) (no related file):*
Test is failing because we enumerate all the default metrics -- go ahead and add `server.uptime` to this switch statement https://github.com/palantir/witchcraft-go-server/blob/02fb84c/integration/metrics_test.go#L77

---
___
*[witchcraft/server_metrics.go, line 39 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpSMrN6DfeUuWGZ3JSW:-LpSMrN6DfeUuWGZ3JSX:bn8laq9) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ba0a0183cdec580a141802510be2860c9f3b0eef/witchcraft/server_metrics.go#L39)):*
> ```Go
> 	}
> 
> 	// start uptime metric
> ```

nit: Let's move all of this to a private `func initServerUptimeMetric(ctx context.Context, metricsRegistry metrics.Registry)` at the bottom of the file
___
*[witchcraft/server_metrics.go, line 40 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpSOEDMAyVnGGTla8Ct:-LpSOEDMAyVnGGTla8Cu:b-th5tgu) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ba0a0183cdec580a141802510be2860c9f3b0eef/witchcraft/server_metrics.go#L40)):*
> ```Go
> 
> 	// start uptime metric
> 	uptimeTimer := metricsRegistry.Timer("server.uptime")
> ```

I think you want a `Gauge` here, since things like p99 do not matter and all you care about is the point-in-time measure (it's also one metrics instead of 14)
___
*[witchcraft/server_metrics.go, line 51 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpSNT219ImISUmxoC3O:-LpSNT219ImISUmxoC3P:bi8e7lv) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ba0a0183cdec580a141802510be2860c9f3b0eef/witchcraft/server_metrics.go#L51)):*
> ```Go
> 				return
> 			case <-t.C:
> 				uptimeTimer.Update(time.Since(initTime) / time.Millisecond)
> ```

I'd actually prefer to lookup the timer each time rather than store it in a local variable. Something like `metricsRegistry.Gauge("server.uptime").Update(...)`
___
*[witchcraft/server_metrics.go, line 51 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpSNckg7U6eqU6D2r7q:-LpSNckg7U6eqU6D2r7r:bamz6y1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ba0a0183cdec580a141802510be2860c9f3b0eef/witchcraft/server_metrics.go#L51)):*
> ```Go
> 				return
> 			case <-t.C:
> 				uptimeTimer.Update(time.Since(initTime) / time.Millisecond)
> ```

Is the java one in milliseconds? In general all of our duration metrics are microseconds (though this one would be much longer than things like response times, so I'd understand using a coarser unit)

blakehawkins:
Thanks @bmoylan 

I haven't managed to login to Reviewable - but I've addressed your comments.

On microseconds vs milliseconds: I think the java version uses milliseconds, but it looks like a time.Duration wraps microseconds anyway. On balance, let's use the existing convention among witchcraft-go. Does that sound okay?

Now it looks like the 500 millisecond timeout is breaking the ratelimiter tests - I'll await your preference for how to proceed.

Easy way out would probably be to change the uptime metric to 150ms tick rate

bmoylan (commented):
Do you have a "Sign in with GitHub" button in the top right of https://reviewable.io/reviews/palantir/witchcraft-go-server/111? 

image

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-:-LpWm_Vy6Nou0Rh87Zsd:b-9yn2ug)* status: all files reviewed, 2 unresolved discussions (waiting on @blakehawkins)

___
*[witchcraft/server_metrics.go, line 39 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpWnDXLBwfr4mo7rn6L:-LpWnDXLBwfr4mo7rn6M:b-5wtep8) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7f70185a831e28fd639d7fb23d155c75a93d3435/witchcraft/server_metrics.go#L39)):*
> ```Go
> 	}
> 
> 	initServerUptimeMetric(ctx, metricsRegistry)
> ```

Let's move this into the `if !s.disableGoRuntimeMetrics` block, which is used in cases where we don't want any default metrics emitted without the server having done anything (normally tests, I don't know of a prod case that uses it)
___
*[witchcraft/server_metrics.go, line 67 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/111#-LpWmd7jCgpn2r3wdWpw:-LpWmd7jCgpn2r3wdWpx:bexjzne) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7f70185a831e28fd639d7fb23d155c75a93d3435/witchcraft/server_metrics.go#L67)):*
> ```Go
> 	go wapp.RunWithRecoveryLogging(ctx, func(ctx context.Context) {
> 		t := time.NewTicker(5.0 * time.Second)
> 		defer t.Stop()
> ```

if you update the metric for the first time up here rather than waiting for the ticker (literally copy the line called when the ticker fires), does that fix the test issue?

Unfortunately we end up doing this 90% of the time we use a ticker, so don't worry if it looks a little ugly

blakehawkins:
Hey @bmoylan - I do have that link, but I just get a "200 OK" when I click it. I wonder if it's being broken by `noscript`, firefox, or something else I'm doing that's atypical. Is it important to set up reviewable for purposes of this review?

I've applied the changes suggested - seems to have fixed the problem for other metrics, but `server.uptime` is still not seen. I can look into this more in the evening, unless you have any immediate thoughts as to what could be going wrong.

witchcraft/server_metrics.go
```diff
@@ -33,6 +38,8 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
 
 	// start routine that capture Go runtime metrics
 	if !s.disableGoRuntimeMetrics {
+		initServerUptimeMetric(ctx, metricsRegistry)
```
bmoylan: If you move this out of the if statement, it fixes the test

integration/metrics_test.go
```diff
@@ -62,6 +62,9 @@ func TestEmitMetrics(t *testing.T) {
 	// Allow the metric emitter to do its thing.
 	time.Sleep(150 * time.Millisecond)
 
+	// Allow the uptime metric to tick.
+	time.Sleep(5.0 * time.Second)
```
bmoylan: I don't think you'll need this anymore (try running the test without it)

gdearment:
@blakehawkins - whats the plan here?

blakehawkins:
> 
> 
> @blakehawkins - whats the plan here?

Hey @gdearment thanks for the ping. I'm going to try and carve out some time with @bmoylan this week to sort out the failing UT. I think Brad already knows what's wrong, so it should just be a matter of prioritisation.

bmoylan (approved):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/111)* status: :shipit: complete! all files reviewed, all discussions resolved

bmoylan:
Just needs a changelog!

bmoylan (approved):

PR #112, opened 2019-09-25
Added a sliding window based health check

### LINKED ISSUE #113: Add a default health check for a sliding window of events
For certain controllers one extremely common pattern for health checks is as follows:
The controller registers some events (example: OnSuccess, OnAttempt, OnFailure) and the health check is a function of all the events received within the last `x` minutes. In other words, there is a sliding window of size `x` that ends in the current time for events that we care.
Since we have 5 controllers using this pattern, we should have a first class API to deal with this.

splucs:
resolves https://github.com/palantir/witchcraft-go-server/issues/113
Sliding window based health checks are an extremely common pattern in Rubix controllers, so we have decided to add this to a shared library. Preferably witchcraft.

Tested by adding unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112)

### DIFF
--- a/status/health/window/window_event_storer.go
+++ b/status/health/window/window_event_storer.go
@@ -1,0 +1,82 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package window
+
+import (
+	"sync"
+	"time"
+
+	werror "github.com/palantir/witchcraft-go-error"
+)
+
+// Event is a struct that keeps a generic payload describing an event and a timestamp for when the event happened.
+type Event struct {
+	// Time is the time the event was submitted.
+	Time time.Time
+	// Payload is any generic information about this event.
+	Payload interface{}
+}
+
+// TimeWindowedEventStorer is a thread-safe struct that stores submitted events
+// and supports polling for all events submitted within the last windowSize period.
+// When any operation is made, all out-of-date events are pruned out of memory.
+type TimeWindowedEventStorer struct {
+	events      []Event
+	eventsMutex sync.Mutex
+	windowSize  time.Duration
+}
+
+// NewTimeWindowedEventStorer creates a new TimeWindowedEventStorer with the provided windowSize.
+func NewTimeWindowedEventStorer(windowSize time.Duration) (*TimeWindowedEventStorer, error) {
+	if windowSize <= 0 {
+		return nil, werror.Error("attempted to create a sliding window with non positive size")
+	}
+	return &TimeWindowedEventStorer{
+		windowSize: windowSize,
+	}, nil
+}
+
+func (t *TimeWindowedEventStorer) pruneOldEvents() {
+	currentTime := time.Now()
+	newStartIndex := 0
+	for index, entry := range t.events {
+		if currentTime.Sub(entry.Time) <= t.windowSize {
+			break
+		}
+		newStartIndex = index + 1
+	}
+	t.events = t.events[newStartIndex:]
+}
+
+// SubmitEvent prunes all out-of-date events out of memory and then adds a new one.
+func (t *TimeWindowedEventStorer) SubmitEvent(payload interface{}) {
+	t.eventsMutex.Lock()
+	defer t.eventsMutex.Unlock()
+
+	t.pruneOldEvents()
+	t.events = append(t.events, Event{
+		Time:    time.Now(),
+		Payload: payload,
+	})
+}
+
+// GetEventsInWindow prunes all out-of-date events out of memory and then returns all up-to-date events.
+func (t *TimeWindowedEventStorer) GetEventsInWindow() []Event {
+	t.eventsMutex.Lock()
+	defer t.eventsMutex.Unlock()
+
+	t.pruneOldEvents()
+	return t.events
+}
--- a/status/health/window/window_event_storer_test.go
+++ b/status/health/window/window_event_storer_test.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package window
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTimeWindowedEventStorer_ErrorOnCreate(t *testing.T) {
+	_, err := NewTimeWindowedEventStorer(0)
+	assert.Error(t, err)
+}
+
+func TestTimeWindowedEventStorer_NoEvents(t *testing.T) {
+	manager, err := NewTimeWindowedEventStorer(time.Millisecond)
+	assert.NoError(t, err)
+	errors := manager.GetEventsInWindow()
+	assert.Nil(t, errors)
+}
+
+func TestTimeWindowedEventStorer_AllEventsUpToDate(t *testing.T) {
+	manager, err := NewTimeWindowedEventStorer(time.Second)
+	assert.NoError(t, err)
+	manager.SubmitEvent("payload #1")
+	manager.SubmitEvent("payload #2")
+	manager.SubmitEvent("payload #3")
+	events := manager.GetEventsInWindow()
+	assert.EqualValues(t, len(events), 3)
+	assert.EqualValues(t, "payload #1", events[0].Payload)
+	assert.EqualValues(t, "payload #2", events[1].Payload)
+	assert.EqualValues(t, "payload #3", events[2].Payload)
+}
+
+func TestTimeWindowedEventStorer_AllEventsOutOfDate(t *testing.T) {
+	manager, err := NewTimeWindowedEventStorer(50 * time.Millisecond)
+	assert.NoError(t, err)
+	manager.SubmitEvent("payload #1")
+	manager.SubmitEvent("payload #2")
+	manager.SubmitEvent("payload #3")
+	<-time.After(100 * time.Millisecond)
+	events := manager.GetEventsInWindow()
+	assert.Empty(t, events)
+}
+
+func TestTimeWindowedEventStorer_SomeEventsOutOfDate(t *testing.T) {
+	manager, err := NewTimeWindowedEventStorer(500 * time.Millisecond)
+	assert.NoError(t, err)
+	manager.SubmitEvent("payload #1")
+	manager.SubmitEvent("payload #2")
+	<-time.After(time.Second)
+	manager.SubmitEvent("payload #3")
+	manager.SubmitEvent("payload #4")
+	events := manager.GetEventsInWindow()
+	assert.EqualValues(t, len(events), 2)
+	assert.EqualValues(t, "payload #3", events[0].Payload)
+	assert.EqualValues(t, "payload #4", events[1].Payload)
+}


nmiyake (changes requested):
Thanks for the PR -- some feedback.

Also, can you open an issue on the repository that describes the pain you're facing and how that is solved by this PR?

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-:-Lpe-i4O5qMbk1byZgzc:b-eb8mzt)* status: all files reviewed, 10 unresolved discussions (waiting on @splucs)

___
*[status/health/window/errors_to_check.go, line 21 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdsAaWB2vamb31abDJ:-LpdsAaWB2vamb31abDK:b-1kbleo) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/2c098470c8a29b55d3ec73269bc20a311871d3e9/status/health/window/errors_to_check.go#L21)):*
> ```Go
> 
> 	"github.com/palantir/witchcraft-go-server/conjure/witchcraft/api/health"
> 	health_util "github.com/palantir/witchcraft-go-server/status/health"
> ```

nit: generally prefer to avoid aliases with underscores. Can you alias the import above this as `conjurehealth` or alias this one as `whealth` (since main difference is witchcraft vs. Conjure)?
___
*[status/health/window/errors_to_check.go, line 37 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdsvANDm_D7l5QgLJ1:-LpdsvAO6191EQiqZcGm:b-p4o9vn) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/errors_to_check.go#L37)):*
> ```Go
> }
> 
> // HealthyIfNotAllErrors builds an ErrorsToCheckFn that returns unhealthy if there are only non-nil errors.
> ```

Would add that, if it does return unhealthy, it returns the first error as an unhealthy check.
___
*[status/health/window/sliding_window.go, line 27 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdt6cQ2joVl6AUcTtK:-Lpdt6cQ2joVl6AUcTtL:bt1or8e) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L27)):*
> ```Go
> }
> 
> // SlidingWindowManager is a thread safe struct where you can submit errors
> ```

nit: prefer avoiding documentation style that contains things like "you" or "we". Would prefer more neutral wording such as:

```
// SlidingWindowManager is a thread-safe struct that stores submitted errors
// and supports polling for all errors submitted within the last windowSize period.
```
___
*[status/health/window/sliding_window.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdtoIJ9tNjc3Mm8SI6:-LpdtoIJ9tNjc3Mm8SI7:bbp1tf3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L30)):*
> ```Go
> // and poll for all errors submitted within the last windowSize period.
> // It includes nil error submissions.
> type SlidingWindowManager struct {
> ```

Documentation on the struct should probably mention pruning behavior (that prune operation is run whenever functions are invoked on the struct)
___
*[status/health/window/sliding_window.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdu7HF9CunE16QMgT8:-Lpdu7HF9CunE16QMgT9:b5h5k5) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L30)):*
> ```Go
> // and poll for all errors submitted within the last windowSize period.
> // It includes nil error submissions.
> type SlidingWindowManager struct {
> ```

Would like a better name than `SlidingWindowManager` -- when seeing only the type name in isolation, it's not clear that this is a struct that stores errors.

Would something like `TimeWindowedErrors` or `TimeWindowedErrorStorer` make more sense? Not sure that either of those are perfect either, but would like to convey more meaning with name.
___
*[status/health/window/sliding_window.go, line 37 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdvj0s0-tqlDrRc9R9:-Lpdvj0s0-tqlDrRc9RA:b-rq0uob) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L37)):*
> ```Go
> 
> // NewSlidingWindowManager creates a new SlidingWindowManager with the provided windowSize.
> func NewSlidingWindowManager(windowSize time.Duration) SlidingWindowManager {
> ```

Should have a check for if this value is negative (if it is, probably makes sense to set it to 0).
___
*[status/health/window/sliding_window.go, line 46 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdw2pAEL6jFeB7qEyY:-Lpdw2pAEL6jFeB7qEyZ:b3j3oyo) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L46)):*
> ```Go
> 	var upToDateErrorEntries []errorEntry
> 	for _, entry := range s.errors {
> 		if time.Now().Sub(entry.time) <= s.windowSize {
> ```

`time.Now()` should be computed and stored in a variable once before the for loop so that the same value is used throughout the loop
___
*[status/health/window/sliding_window.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdwEQ54jhefJOALQw1:-LpdwEQ54jhefJOALQw2:bz8lnih) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L47)):*
> ```Go
> 	for _, entry := range s.errors {
> 		if time.Now().Sub(entry.time) <= s.windowSize {
> 			upToDateErrorEntries = append(upToDateErrorEntries, entry)
> ```

I would use a different strategy for pruning.

This struct always appends submitted errors to the end of `s.errors`, and since this is done under lock it's guaranteed that the time in each subsequent element is >= the previous element.

In that case, you just need to find the index of the first value that is within the window -- once you find that value, you can just set the slice to start at that position.

So, something like:

```go
if len(s.errors) == 0 {
   return
}
earliestValidTime := time.Now().Sub(s.windowSize)

// optimization: if latest error is before earliest valid time, there are no valid errors
if s.errors[len(s.errors)-1].time.Before(earliestValidTime) {
     s.errors = nil
     return
}

// otherwise, at least one error is within the window: find the index and update slice to start there
for i, entry := range s.errors {
       if !entry.time.Before(earliestValidTime) {
             s.errors = s.errors[i:]
       }
}
```
___
*[status/health/window/sliding_window.go, line 53 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdzxaF5sb7AOX0r0mj:-LpdzxaF5sb7AOX0r0mk:b-lghhyx) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L53)):*
> ```Go
> }
> 
> // SubmitError prunes all out-of-date errors out of memory and adds a new one.
> ```

Would rephrase documentation to indicate that this function records the provided error, and prunes any out-of-date errors before doing so. Even though the final content is the same, I think the documentation should emphasize the action of the function.
___
*[status/health/window/sliding_window.go, line 65 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpe-Jkt3hllHmCzWjM6:-Lpe-Jkt3hllHmCzWjM7:b-819em4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L65)):*
> ```Go
> }
> 
> // GetErrors prunes all out-of-date errors out of memory and returns all up-to-date errors.
> ```

Same as above for focus of documentation. Should also note that it returns a newly allocated slice.

ryanmcnamara:
Probably better to put these thoughts on the issue, but for now here:

# Attempt awareness
Often the desired health check desired on a controller is 
> If there have been attempts made have they succeeded?

The api / implementation as is currently misses some cases (considers them healthy)
* An attempt starts but never completes
* Attempts stop in an expected way but the last attempt made failed

An alternative API that wouldn't have these gaps would allow marking attempts and successes / failures and derive health from that.

# Cardinality issues
Often with a controller the operator wants to know which of the many resources (suppose hundreds) being managed are not succeeding, and the lifecycle of the attempt / success / failure of each resource is separate. The current api would require creating one check per resource which I don't think scales correctly given many resources

# Misc
I don't see the use case for `UnhealthyIfAtLeastOneError`. Practically a single error shouldn't be enough to be unhealthy

splucs (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112)* status: 1 of 5 files reviewed, 8 unresolved discussions (waiting on @nmiyake)

___
*[status/health/window/errors_to_check.go, line 37 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdsvANDm_D7l5QgLJ1:-Lpe5VAP65cTON6JRCf7:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/errors_to_check.go#L37)):*

Would add that, if it does return unhealthy, it returns the first error as an unhealthy check.
</blockquote></details>

Done.
___
*[status/health/window/sliding_window.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdtoIJ9tNjc3Mm8SI6:-Lpe5fZY5prDBJQP8Gen:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L30)):*

Documentation on the struct should probably mention pruning behavior (that prune operation is run whenever functions are invoked on the struct)
</blockquote></details>

Done.
___
*[status/health/window/sliding_window.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdu7HF9CunE16QMgT8:-Lpe5hkW7P6KBzzm1xM4:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L30)):*

Would like a better name than `SlidingWindowManager` -- when seeing only the type name in isolation, it's not clear that this is a struct that stores errors.

Would something like `TimeWindowedErrors` or `TimeWindowedErrorStorer` make more sense? Not sure that either of those are perfect either, but would like to convey more meaning with name.
</blockquote></details>

Done.
___
*[status/health/window/sliding_window.go, line 37 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdvj0s0-tqlDrRc9R9:-Lpe5j9Y1OqMhIy2yuvt:bih22re) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L37)):*

Should have a check for if this value is negative (if it is, probably makes sense to set it to 0).
</blockquote></details>

A size zero sliding window is always empty. I will make the constructor return an error instead.
___
*[status/health/window/sliding_window.go, line 46 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpdw2pAEL6jFeB7qEyY:-Lpe8GbH3VdonyCMWG-F:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L46)):*

`time.Now()` should be computed and stored in a variable once before the for loop so that the same value is used throughout the loop
</blockquote></details>

Done.
___
*[status/health/window/sliding_window.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdwEQ54jhefJOALQw1:-Lpe8R3_6OgpJ8_TwIw7:b-4t5j8t) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L47)):*

I would use a different strategy for pruning.

This struct always appends submitted errors to the end of `s.errors`, and since this is done under lock it's guaranteed that the time in each subsequent element is >= the previous element.

In that case, you just need to find the index of the first value that is within the window -- once you find that value, you can just set the slice to start at that position.

So, something like:

```go
if len(s.errors) == 0 {
   return
}
earliestValidTime := time.Now().Sub(s.windowSize)

// optimization: if latest error is before earliest valid time, there are no valid errors
if s.errors[len(s.errors)-1].time.Before(earliestValidTime) {
     s.errors = nil
     return
}

// otherwise, at least one error is within the window: find the index and update slice to start there
for i, entry := range s.errors {
       if !entry.time.Before(earliestValidTime) {
             s.errors = s.errors[i:]
       }
}
```

</blockquote></details>

I was wary of doing this since time.Now() is not guaranteed to work always. However, on second thought, even if there is some small deviation, it should be fine to have an error that is a few microseconds behind the window.
Optimization added.
___
*[status/health/window/sliding_window.go, line 53 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpdzxaF5sb7AOX0r0mj:-Lpe8oz5EUOeYvgfhp74:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L53)):*

Would rephrase documentation to indicate that this function records the provided error, and prunes any out-of-date errors before doing so. Even though the final content is the same, I think the documentation should emphasize the action of the function.
</blockquote></details>

Done.
___
*[status/health/window/sliding_window.go, line 65 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpe-Jkt3hllHmCzWjM6:-Lpe8pKA81vdF9563Y9L:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/83cf10caef88460ef606be9ffe835b53912bebda/status/health/window/sliding_window.go#L65)):*

Same as above for focus of documentation. Should also note that it returns a newly allocated slice.
</blockquote></details>

Done.

splucs:
@ryanmcnamara see the updates.
I have changed the health check to work with a generic payload type.
Given that you can put whatever you want in the payload (keys for resources + an enum for OnAttempt, OnSuccess, OnFailure + an error) and given that you can set up custom logic to derive health checks for events, you can easily extend this helper to support attempts with timeouts, multi resource checks or whatever custom logic needed.
I also added a EventsToCheckFn to deal with multiple resources in a single health check.
Since the attempts logic seemed pretty specific to your application, I did not include it, but extending this to support it should be trivial.

nmiyake (changes requested):
> Reviewed 1 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @nmiyake and @splucs)

___
*[status/health/window/sliding_window.go, line 42 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpiZvnq9ijNoF8MtDI0:-LpiZvnq9ijNoF8MtDI1:b-qzxqy) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/eff45de0e334f14159ea9cd8ca62e8f33657bfa9/status/health/window/sliding_window.go#L42)):*
> ```Go
> 
> // NewTimeWindowedEventStorer creates a new TimeWindowedEventStorer with the provided windowSize.
> func NewTimeWindowedEventStorer(windowSize time.Duration) (TimeWindowedEventStorer, error) {
> ```

Since all of the functions are on a pointer receiver, usability is improved if this function returns `*TimeWindowedEventStorer`.
___
*[status/health/window/source.go, line 33 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpiQvLL9kttx2GMVbdh:-LpiQvLL9kttx2GMVbdi:b-o0128v) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/eff45de0e334f14159ea9cd8ca62e8f33657bfa9/status/health/window/source.go#L33)):*
> ```Go
> 
> // NewHealthCheck creates a HealthCheckSource that polls a TimeWindowedEventStorer and returns a HealthStatus created using an EventsToCheckFn.
> func NewHealthCheck(timeWindowedEventStorer *TimeWindowedEventStorer, errorsToCheckFn EventsToCheckFn) status.HealthCheckSource {
> ```

I would prefer a safer API. To use this right now, it seems like you have to:

* Construct a `TimeWindowedEventStorer` externally
* Create/choose a `EventsToCheckFn` of a particular type
* Users will submit events to the `TimeWindowedEventStorer` using `SubmitEvent(interface{})`, and the type of event they submit must be compatible with the type expected by the specific `EventsToCheckFn` that is used

That makes this API quite dangerous -- given a specific `TimeWindowedEventStorer`, you have to trace the code back to creation to figure out what kind of event it accepts, and if you submit a bad event, the panic doesn't happen until a later time (the next time that `HealthStatus` is invoked and it tries to convert the submitted entry), and when that happens it's going to be very hard to figure out what submitted the invalid value.

My preference would be to keep `TimeWindowedEventStorer` as a helper struct that can be used to create specific checks, and have specific implementations create their own typed API using it.

What about something like this:

```go
type BaseTimeWindowHealthCheck struct {
	timeWindowedEventStorer *TimeWindowedEventStorer
	eventsToCheckFn         EventsToCheckFn
}

func (h *BaseTimeWindowHealthCheck) SubmitEvent(payload interface{}) {
	h.timeWindowedEventStorer.SubmitEvent(payload)
}

func (h *BaseTimeWindowHealthCheck) HealthStatus(ctx context.Context) health.HealthStatus {
	checkResult := h.eventsToCheckFn(ctx, h.timeWindowedEventStorer.GetEventsInWindow())
	return health.HealthStatus{
		Checks: map[health.CheckType]health.HealthCheckResult{
			checkResult.Type: checkResult,
		},
	}
}

var _ status.HealthCheckSource = &UnhealthyIfAtLeastOneErrorHealthCheck{}

type UnhealthyIfAtLeastOneErrorHealthCheck struct {
	baseCheck *BaseTimeWindowHealthCheck
}

func NewUnhealthyIfAtLeastOneErrorCheck(checkType health.CheckType, d time.Duration) (*UnhealthyIfAtLeastOneErrorHealthCheck, error) {
	eventStorer, err := NewTimeWindowedEventStorer(d)
	if err != nil {
		return nil, err
	}
	return &UnhealthyIfAtLeastOneErrorHealthCheck{
		baseCheck: &BaseTimeWindowHealthCheck{
			timeWindowedEventStorer: eventStorer,
			eventsToCheckFn: func(ctx context.Context, events []Event) health.HealthCheckResult {
				for _, event := range events {
					if event.Payload != nil {
						return whealth.UnhealthyHealthCheckResult(checkType, event.Payload.(error).Error())
					}
				}
				return whealth.HealthyHealthCheckResult(checkType)
			},
		},
	}, nil
}

func (h *UnhealthyIfAtLeastOneErrorHealthCheck) SubmitError(err error) {
	h.baseCheck.SubmitEvent(err)
}

func (h *UnhealthyIfAtLeastOneErrorHealthCheck) HealthStatus(ctx context.Context) health.HealthStatus {
	return h.baseCheck.HealthStatus(ctx)
}

// EventsToCheckFn is a function that constructs a HealthCheckResult from a set of events.
type EventsToCheckFn func(ctx context.Context, events []Event) health.HealthCheckResult

type healthCheckSource struct {
	slidingWindowManager *TimeWindowedEventStorer
	eventsToCheckFn      EventsToCheckFn
}

// NewHealthCheck creates a HealthCheckSource that polls a TimeWindowedEventStorer and returns a HealthStatus created using an EventsToCheckFn.
func NewHealthCheck(timeWindowedEventStorer *TimeWindowedEventStorer, errorsToCheckFn EventsToCheckFn) status.HealthCheckSource {
	return &healthCheckSource{
		slidingWindowManager: timeWindowedEventStorer,
		eventsToCheckFn:      errorsToCheckFn,
	}
}

func (h *healthCheckSource) HealthStatus(ctx context.Context) health.HealthStatus {
	checkResult := h.eventsToCheckFn(ctx, h.slidingWindowManager.GetEventsInWindow())
	return health.HealthStatus{
		Checks: map[health.CheckType]health.HealthCheckResult{
			checkResult.Type: checkResult,
		},
	}
}
```

This requires each new check type to declare a new type, provides for type safety of the function that adds events.

We don't have to take this exact approach, but would like something that provides a bit more safety, especially since my guess is that it's not critical to have a single common API signature from a consumption standpoint.

splucs (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[status/health/window/sliding_window.go, line 42 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpiZvnq9ijNoF8MtDI0:-LpmGsIj7MyrFm3h6IVh:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/eff45de0e334f14159ea9cd8ca62e8f33657bfa9/status/health/window/sliding_window.go#L42)):*

Since all of the functions are on a pointer receiver, usability is improved if this function returns `*TimeWindowedEventStorer`.
</blockquote></details>

Done.
___
*[status/health/window/source.go, line 33 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-LpiQvLL9kttx2GMVbdh:-LpmGCcB2XFBJJ8tuzys:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/eff45de0e334f14159ea9cd8ca62e8f33657bfa9/status/health/window/source.go#L33)):*

I would prefer a safer API. To use this right now, it seems like you have to:

* Construct a `TimeWindowedEventStorer` externally
* Create/choose a `EventsToCheckFn` of a particular type
* Users will submit events to the `TimeWindowedEventStorer` using `SubmitEvent(interface{})`, and the type of event they submit must be compatible with the type expected by the specific `EventsToCheckFn` that is used

That makes this API quite dangerous -- given a specific `TimeWindowedEventStorer`, you have to trace the code back to creation to figure out what kind of event it accepts, and if you submit a bad event, the panic doesn't happen until a later time (the next time that `HealthStatus` is invoked and it tries to convert the submitted entry), and when that happens it's going to be very hard to figure out what submitted the invalid value.

My preference would be to keep `TimeWindowedEventStorer` as a helper struct that can be used to create specific checks, and have specific implementations create their own typed API using it.

What about something like this:

```go
type BaseTimeWindowHealthCheck struct {
	timeWindowedEventStorer *TimeWindowedEventStorer
	eventsToCheckFn         EventsToCheckFn
}

func (h *BaseTimeWindowHealthCheck) SubmitEvent(payload interface{}) {
	h.timeWindowedEventStorer.SubmitEvent(payload)
}

func (h *BaseTimeWindowHealthCheck) HealthStatus(ctx context.Context) health.HealthStatus {
	checkResult := h.eventsToCheckFn(ctx, h.timeWindowedEventStorer.GetEventsInWindow())
	return health.HealthStatus{
		Checks: map[health.CheckType]health.HealthCheckResult{
			checkResult.Type: checkResult,
		},
	}
}

var _ status.HealthCheckSource = &UnhealthyIfAtLeastOneErrorHealthCheck{}

type UnhealthyIfAtLeastOneErrorHealthCheck struct {
	baseCheck *BaseTimeWindowHealthCheck
}

func NewUnhealthyIfAtLeastOneErrorCheck(checkType health.CheckType, d time.Duration) (*UnhealthyIfAtLeastOneErrorHealthCheck, error) {
	eventStorer, err := NewTimeWindowedEventStorer(d)
	if err != nil {
		return nil, err
	}
	return &UnhealthyIfAtLeastOneErrorHealthCheck{
		baseCheck: &BaseTimeWindowHealthCheck{
			timeWindowedEventStorer: eventStorer,
			eventsToCheckFn: func(ctx context.Context, events []Event) health.HealthCheckResult {
				for _, event := range events {
					if event.Payload != nil {
						return whealth.UnhealthyHealthCheckResult(checkType, event.Payload.(error).Error())
					}
				}
				return whealth.HealthyHealthCheckResult(checkType)
			},
		},
	}, nil
}

func (h *UnhealthyIfAtLeastOneErrorHealthCheck) SubmitError(err error) {
	h.baseCheck.SubmitEvent(err)
}

func (h *UnhealthyIfAtLeastOneErrorHealthCheck) HealthStatus(ctx context.Context) health.HealthStatus {
	return h.baseCheck.HealthStatus(ctx)
}

// EventsToCheckFn is a function that constructs a HealthCheckResult from a set of events.
type EventsToCheckFn func(ctx context.Context, events []Event) health.HealthCheckResult

type healthCheckSource struct {
	slidingWindowManager *TimeWindowedEventStorer
	eventsToCheckFn      EventsToCheckFn
}

// NewHealthCheck creates a HealthCheckSource that polls a TimeWindowedEventStorer and returns a HealthStatus created using an EventsToCheckFn.
func NewHealthCheck(timeWindowedEventStorer *TimeWindowedEventStorer, errorsToCheckFn EventsToCheckFn) status.HealthCheckSource {
	return &healthCheckSource{
		slidingWindowManager: timeWindowedEventStorer,
		eventsToCheckFn:      errorsToCheckFn,
	}
}

func (h *healthCheckSource) HealthStatus(ctx context.Context) health.HealthStatus {
	checkResult := h.eventsToCheckFn(ctx, h.slidingWindowManager.GetEventsInWindow())
	return health.HealthStatus{
		Checks: map[health.CheckType]health.HealthCheckResult{
			checkResult.Type: checkResult,
		},
	}
}
```

This requires each new check type to declare a new type, provides for type safety of the function that adds events.

We don't have to take this exact approach, but would like something that provides a bit more safety, especially since my guess is that it's not critical to have a single common API signature from a consumption standpoint.
</blockquote></details>

Done.

nmiyake (changes requested):
Thank you for updating the PR! In sync with the over-all approach (just some minor feedback on filenames).

Also, now that we've determined the structure we want to use, can you please break up this PR so that the review (and change) is more manageable? Ideally, should be:

* 1 PR that adds the base type and implementation
* Separate PRs for each implementation that is added (these can be independent, but each one based on the PR above)

This will make the PRs conceptually cleaner, and I'll be able to delve into them deeper to review all the code (including tests) more efficiently (and we can iterate on them in parallel).

I understand that this is blocking things on your end, so will commit to doing the reviews of these updated PRs quickly. Thanks for your patience!

> Reviewed 1 of 7 files at r4, 5 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-:-Lpo3aEFEp-JkCVX7Dat:b-r5k8h9)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @nmiyake and @splucs)

___
*[status/health/window/sliding_window.go, line 42 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpo3TxW0ubuQnRIYXSU:-Lpo3TxW0ubuQnRIYXSV:bxglai1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d6cff640307a97ad626753d8080339a2abba048c/status/health/window/sliding_window.go#L42)):*
> ```Go
> 
> // NewTimeWindowedEventStorer creates a new TimeWindowedEventStorer with the provided windowSize.
> func NewTimeWindowedEventStorer(windowSize time.Duration) (*TimeWindowedEventStorer, error) {
> ```

The filename (`sliding_window.go`) should be renamed to match
___
*[status/health/window/sliding_window_test.go, line 24 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/112#-Lpo3Y9e-QFb1CGf8sxY:-Lpo3Y9e-QFb1CGf8sxZ:b-4bgun6) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d6cff640307a97ad626753d8080339a2abba048c/status/health/window/sliding_window_test.go#L24)):*
> ```Go
> )
> 
> func TestTimeWindowedEventStorer_ErrorOnCreate(t *testing.T) {
> ```

The filename (`sliding_window_test.go`) should be renamed to match

PR #119, opened 2019-09-30
Add must constructors for sliding window health checks

splucs:
Most of these constructors will be used in initialization, which will mean their parameters will be statically defined.
Added the Must constructors to simplify code in such cases.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/119)


bmoylan (approved):
Confirmed that these contructors' errors are only input validation so known-good inputs should never panic

PR #120, opened 2019-09-30
Log init func invocation.

jamesross3:
This log line provides a useful timestamp when debugging any startup-related issues.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120)

### DIFF
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -560,6 +560,7 @@ func (s *Server) Start() (rErr error) {
 		}
 		ctx = wtracing.ContextWithTracer(ctx, tracer)
 
+		svc1log.FromContext(ctx).Info("Invoking server init func.")
 		cleanupFn, err := s.initFn(
 			ctx,
 			InitInfo{
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -560,7 +560,7 @@ func (s *Server) Start() (rErr error) {
 		}
 		ctx = wtracing.ContextWithTracer(ctx, tracer)
 
-		svc1log.FromContext(ctx).Debug("Invoking server init func.")
+		svc1log.FromContext(ctx).Debug("Running server initialization function.")
 		cleanupFn, err := s.initFn(
 			ctx,
 			InitInfo{
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -560,7 +560,7 @@ func (s *Server) Start() (rErr error) {
 		}
 		ctx = wtracing.ContextWithTracer(ctx, tracer)
 
-		svc1log.FromContext(ctx).Info("Invoking server init func.")
+		svc1log.FromContext(ctx).Debug("Invoking server init func.")
 		cleanupFn, err := s.initFn(
 			ctx,
 			InitInfo{


nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jamesross3)

___
*[witchcraft/witchcraft.go, line 563 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-Lq22YCY4qMvBz1ETd1n:-Lq22YCY4qMvBz1ETd1o:b-n951qk) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9d4a88f5ad2a80e66e56a51a07df460057854413/witchcraft/witchcraft.go#L563)):*
> ```Go
> 		ctx = wtracing.ContextWithTracer(ctx, tracer)
> 
> 		svc1log.FromContext(ctx).Info("Invoking server init func.")
> ```

I'm a bit hesitant about this -- we don't do a lot of logging in other parts of the general server, and not sure if this specific line contributes enough value to warrant including it by default.

Is there a cross-cutting issue which this will really help with? If you want to debug a particular service's issue, you can add logging equivalent to this to the first line of the `initFn` and it will be called.

If there is a strong case for adding this generically, then this level of logging should definitely be at the DEBUG level, since it's very much about an implementation detail (as opposed to something like "Server starting", which I could possibly see as a higher-level INFO message).

Would like to further understand the case this is meant to catch (and why it warrants adding it to the common layer for all consumers).

jamesross3:
I agree that if we add this log line then I should change it to `DEBUG`. When running witchcraft go applications in kubernetes, we rely on the liveness endpoint being served soon enough that the application's container doesn't get restarted due to a liveness probe failing. Debugging a liveness probe failure that we believe to be around the time of startup partly depends on having a reliable timestamp when the application starts doing things (as opposed to the container runtime doing things) - across our products, this is the method we pass to `WithInitFunc()`.

nmiyake (changes requested):
So currently we log `"Listening to https"` when the server is being started (which is after the init function has fully processed). Based on that, the only scenario where the extra debugging is useful is if the init function itself takes a long time/locks. Is this something that you are observing happening regularly? If so, then I'm not opposed to adding the logging, but just want to understand that part better.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-:-Lq2DxV80vvepVvwbaFK:b-8uvb4e)* status: all files reviewed, 1 unresolved discussion (waiting on @jamesross3)

___
*[witchcraft/witchcraft.go, line 563 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-Lq22YCY4qMvBz1ETd1n:-Lq2F3m_DySDk6NmAYwe:b3ebvnt) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9d4a88f5ad2a80e66e56a51a07df460057854413/witchcraft/witchcraft.go#L563)):*

I'm a bit hesitant about this -- we don't do a lot of logging in other parts of the general server, and not sure if this specific line contributes enough value to warrant including it by default.

Is there a cross-cutting issue which this will really help with? If you want to debug a particular service's issue, you can add logging equivalent to this to the first line of the `initFn` and it will be called.

If there is a strong case for adding this generically, then this level of logging should definitely be at the DEBUG level, since it's very much about an implementation detail (as opposed to something like "Server starting", which I could possibly see as a higher-level INFO message).

Would like to further understand the case this is meant to catch (and why it warrants adding it to the common layer for all consumers).
</blockquote></details>

If we do keep the message, would like it to be something like: "Running server initialization function" or "Initializing server"

jamesross3:
Yes, we have some products with init methods that are _usually_ quick but can potentially take a while.

nmiyake (commented):
Would like to update the content of the output as noted in the review comment (don't like the "server init func." look as part of a built-in log line)

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-:-Lq3WweRBu2TEFWtZ7Ft:bxo31ob)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jamesross3 and @nmiyake)

jamesross3 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[witchcraft/witchcraft.go, line 563 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-Lq22YCY4qMvBz1ETd1n:-Lq3XmLh9-TWXeRCJVtK:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9d4a88f5ad2a80e66e56a51a07df460057854413/witchcraft/witchcraft.go#L563)):*

If we do keep the message, would like it to be something like: "Running server initialization function" or "Initializing server"
</blockquote></details>

Done.

bmoylan:
@jamesross3 do we still want this?

jamesross3:
@bmoylan yes please

bmoylan (approved):

nmiyake (commented):
Can we add changelog for this? Otherwise content looks good.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120#-:-Lu3hj7S6PiWhsZ_ZaGC:b-4rd958)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @nmiyake)

nmiyake (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/120)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #122, opened 2019-10-14
Separate out trace sampling for application routes and management routes


k-simons:
- Right now these use the same trace sampling configuration and tracer
- These should be broken out into their own tracers and made configurable

PR #123, opened 2019-10-15
Add management tracer that gets registered to management routes

k-simons:
- Addressed https://github.com/palantir/witchcraft-go-server/issues/122
- Add identical configurability to the application and management tracer
- The management tracer defaults to no tracing
- Add `WithManagementTraceSamplerRate` and `WithTraceSamplerRate` helper functions. (not tied to this, but seemed like a natural extension). Can remove if you prefer
- Explicitly add a `alwaysTrace` option to the application tracer if no others are set. Before this relied on an implementation detail of wtracing
- Add a management trace rate to the install config and write a test for all on the install config
- Happy to add whatever additional testing you prefer, I captured the current testing coverage of the applicationTracer with the managementTracer, but there was not a ton there and can add additional tests if you have suggestions 

This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/123)

### DIFF
--- a/config/install.go
+++ b/config/install.go
@@ -21,11 +21,12 @@ import (
 // Install specifies the base install configuration fields that should be included in all witchcraft-go-server server
 // install configurations.
 type Install struct {
-	ProductName          string        `yaml:"product-name,omitempty"`
-	Server               Server        `yaml:"server,omitempty"`
-	MetricsEmitFrequency time.Duration `yaml:"metrics-emit-frequency,omitempty"`
-	TraceSampleRate      *float64      `yaml:"trace-sample-rate,omitempty"`
-	UseConsoleLog        bool          `yaml:"use-console-log,omitempty"`
+	ProductName               string        `yaml:"product-name,omitempty"`
+	Server                    Server        `yaml:"server,omitempty"`
+	MetricsEmitFrequency      time.Duration `yaml:"metrics-emit-frequency,omitempty"`
+	TraceSampleRate           *float64      `yaml:"trace-sample-rate,omitempty"`
+	ManagementTraceSampleRate *float64      `yaml:"management-trace-sample-rate,omitempty"`
+	UseConsoleLog             bool          `yaml:"use-console-log,omitempty"`
 }
 
 type Server struct {
--- a/config/install_test.go
+++ b/config/install_test.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2018 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
+)
+
+func TestInstallConfig(t *testing.T) {
+	conf := `
+product-name: productName
+trace-sample-rate: 0.5
+management-trace-sample-rate: 0.6
+use-console-log: true
+metrics-emit-frequency: 1s
+server:
+  address: address
+  port: 10
+  management-port: 11
+  context-path: ContextPath
+  client-ca-files:
+    - a
+    - b
+  cert-file: certFile
+  key-file: keyFile
+`
+	var install Install
+	err := yaml.Unmarshal([]byte(conf), &install)
+	assert.NoError(t, err)
+	assert.Equal(t, Install{
+		ProductName: "productName",
+		Server: Server{
+			Address:        "address",
+			Port:           10,
+			ManagementPort: 11,
+			ContextPath:    "ContextPath",
+			ClientCAFiles:  []string{"a", "b"},
+			CertFile:       "certFile",
+			KeyFile:        "keyFile",
+		},
+		MetricsEmitFrequency:      time.Second,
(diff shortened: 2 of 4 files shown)


nmiyake (changes requested):
Majority of the content/change looks good to me -- main thing is a question about what the behavior is when the application and management ports are the same.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-:-LrCZik8A54Nt4xRqc4T:b6drszk)* status: all files reviewed, 3 unresolved discussions (waiting on @k-simons)

___
*[config/install_test.go, line 1 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrCYLK62ndLOIaPYJkG:-LrCYLK62ndLOIaPYJkH:b-iehf6h) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dba35755067854b1491236b4a9f4d8fe5e7e84a7/config/install_test.go#L1)):*
> ```Go
> // Copyright (c) 2018 Palantir Technologies. All rights reserved.
> ```

nit: since this is a new file, the year should be 2019
___
*[witchcraft/witchcraft.go, line 554 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrC_B7l6NO1tcZCPb5J:-LrC_B7l6NO1tcZCPb5K:bjrmmri) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dba35755067854b1491236b4a9f4d8fe5e7e84a7/witchcraft/witchcraft.go#L554)):*
> ```Go
> 	// add middleware
> 	s.addMiddleware(router.RootRouter(), metricsRegistry, s.getApplicationTracingOptions(baseInstallCfg))
> 	if mgmtRouter != router {
> ```

What about the case when `mgmtRouter == router` (the case where the ports are the same)? Since I don't see any other code that references the management tracing options, seems like in that case it may use the application tracer?
___
*[witchcraft/witchcraft.go, line 827 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrCZnPN5cCQ_Nm1TqNq:-LrCZnPN5cCQ_Nm1TqNr:b-6wdbxg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ebd27c01f95e324154a3d59b5e97c9699e4b8c2c/witchcraft/witchcraft.go#L827)):*
> ```Go
> 	if sampleRate <= 0 {
> 		if s.svcLogger != nil {
> 			s.svcLogger.Info("Trace logging will be disabled due to trace-sample-rate <= 0")
> ```

Removing this logging does slightly modify behavior, but I am okay with the modification (especially since the message isn't entirely accurate, as if the incoming request has sampling headers then that will determine the behavior)

k-simons (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/123)* status: all files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[config/install_test.go, line 1 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrCYLK62ndLOIaPYJkG:-LrCaGoFELGCScal2Hkj:bb6un1j) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dba35755067854b1491236b4a9f4d8fe5e7e84a7/config/install_test.go#L1)):*

nit: since this is a new file, the year should be 2019
</blockquote></details>

Done
___
*[witchcraft/witchcraft.go, line 554 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrC_B7l6NO1tcZCPb5J:-LrCaUwB4uSpU8d95zfZ:b-rjlgsw) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dba35755067854b1491236b4a9f4d8fe5e7e84a7/witchcraft/witchcraft.go#L554)):*

What about the case when `mgmtRouter == router` (the case where the ports are the same)? Since I don't see any other code that references the management tracing options, seems like in that case it may use the application tracer?
</blockquote></details>

Yep that was intended, in the case that the routers were the same I thought that you would want the same tracer.  

Perhaps thats incorrect. I poked around the code a bit to explore the alternative, and it seems a bit confusing. Would the idea be to attach multiple middlewares to the application router and somehow do different things/register middleware for different routes? 
___
*[witchcraft/witchcraft.go, line 827 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrCZnPN5cCQ_Nm1TqNq:-LrCaLeP7533FMEJMfcj:bq6f7mo) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ebd27c01f95e324154a3d59b5e97c9699e4b8c2c/witchcraft/witchcraft.go#L827)):*

Removing this logging does slightly modify behavior, but I am okay with the modification (especially since the message isn't entirely accurate, as if the incoming request has sampling headers then that will determine the behavior)
</blockquote></details>

Ack. We could pass in a logger but it seemed like overkill

k-simons:
Would still like to test ete after CR but before merging just to make sure everything works as expected

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/123)* status: all files reviewed, 1 unresolved discussion (waiting on @k-simons)

___
*[witchcraft/witchcraft.go, line 554 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/123#-LrC_B7l6NO1tcZCPb5J:-LrGQ3az1Z_AFe7eJk7y:bhklvx3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dba35755067854b1491236b4a9f4d8fe5e7e84a7/witchcraft/witchcraft.go#L554)):*
<details><summary><i>Previously, k-simons wrote…</i></summary><blockquote>

Yep that was intended, in the case that the routers were the same I thought that you would want the same tracer.  

Perhaps thats incorrect. I poked around the code a bit to explore the alternative, and it seems a bit confusing. Would the idea be to attach multiple middlewares to the application router and somehow do different things/register middleware for different routes? 
</blockquote></details>

That's probably fine. I don't think the config is well-documented right now, but if we do add documentation to config (either as comments or something else), then this information is something we should probably add there.

PR #124, opened 2019-10-15
Add return statement for trace sampling

k-simons:
- Follow up of https://github.com/palantir/witchcraft-go-server/pull/123
- Adds a missing return statement to the `getSamplingTraceOption` function
- Add a table-test for `getSamplingTraceOption`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/124)

### DIFF
--- a/witchcraft/trace_test.go
+++ b/witchcraft/trace_test.go
@@ -90,7 +90,7 @@ func TestGetTracingOptions(t *testing.T) {
 	install := config.Install{
 		ProductName: "product",
 	}
-	port := 10
+	const port = 10
 	for _, test := range []struct {
 		name       string
 		sampler    wtracing.Sampler
--- a/witchcraft/trace_test.go
+++ b/witchcraft/trace_test.go
@@ -16,11 +16,11 @@ package witchcraft
 
 import (
 	"fmt"
-	"github.com/palantir/witchcraft-go-server/config"
-	"github.com/palantir/witchcraft-go-tracing/wtracing"
 	"math"
 	"testing"
 
+	"github.com/palantir/witchcraft-go-server/config"
+	"github.com/palantir/witchcraft-go-tracing/wtracing"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


nmiyake (approved):
Minor feedback, but since it's just a minor suggestion for test giving approval.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/124#-:-LrGiLnaB0UfcYLl9-MX:b-5ztrex)* status: all files reviewed, 1 unresolved discussion (waiting on @k-simons)

___
*[witchcraft/trace_test.go, line 93 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/124#-LrGi-n1Decdy1cNnKK7:-LrGi-n1Decdy1cNnKK8:b-1qs3re) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9da007d5e1bb550dab90b388e7e2663335bd14da/witchcraft/trace_test.go#L93)):*
> ```Go
> 		ProductName: "product",
> 	}
> 	port := 10
> ```

Is there a reason this is declared all the way up here?

If you don't want to use this directly in the call, then would change the declaration to `const port = 10` since this is a constant rather than a variable.

k-simons (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/124)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[witchcraft/trace_test.go, line 93 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/124#-LrGi-n1Decdy1cNnKK7:-LrGioE_0YI2iLnvf7nG:b-sfrr8y) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9da007d5e1bb550dab90b388e7e2663335bd14da/witchcraft/trace_test.go#L93)):*

Is there a reason this is declared all the way up here?

If you don't want to use this directly in the call, then would change the declaration to `const port = 10` since this is a constant rather than a variable.
</blockquote></details>

Done, changed to const. I could also inline this in the invocation, just wanted to make it clear it was not changing in each loop iteration.

I did not want to put a `const` called `port` outside the test and have not really seen `const` inside a function before, so didn't think of that.

PR #128, opened 2019-11-05
Feature: add support for blacklisting metric type values

### LINKED ISSUE #127: Default server/client metrics too verbose, no option to trim down
Some of the default request/server metrics being emitted break out into 12 distinct values (15m, 1m, 5m, Count, Max, Mean, meanRate, min, p50, p95, p99, stddev), which are probably not all equally valuable and are expensive, especially when emitted by daemons deployed on every single host. 

I propose we move these metrics to a whitelist model, where we have a default filter of values that we care about for a given metric (i.e if i only care about P99, i should be able to only get that) and allow products to override that filter. 

This gives us 1) ability for us to have a sane default that we can iterate on over time, depending on the value of metric vs cost trade-off 2) have products that have different constraints make different decisions (i.e daemons with health end points only might not care about any of those).

nmiyake:
Adds WithMetricTypeValuesBlacklist function to server that allows
specifying values that should not be emitted to the metric logs
on a per-metric basis.

Fixes #127

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/128)

### DIFF
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -118,3 +118,92 @@ func TestEmitMetrics(t *testing.T) {
 	default:
 	}
 }
+
+// TestMetricTypeValueBlacklist tests that if a metric type value is blacklisted, all metric of that type does not
+// contain any of the blacklisted keys.
+func TestMetricTypeValueBlacklist(t *testing.T) {
+	logOutputBuffer := &bytes.Buffer{}
+	port, err := httpserver.AvailablePort()
+	require.NoError(t, err)
+
+	// ensure that registry used in this test is unique/does not have any past metrics registered on it
+	metrics.DefaultMetricsRegistry = metrics.NewRootMetricsRegistry()
+	server, serverErr, cleanup := createAndRunCustomTestServer(t, port, port, func(ctx context.Context, info witchcraft.InitInfo) (deferFn func(), rErr error) {
+		ctx = metrics.AddTags(ctx, metrics.MustNewTag("key", "val"))
+		metrics.FromContext(ctx).Counter("my-counter").Inc(13)
+		return nil, info.Router.Post("/error", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			rw.WriteHeader(500)
+		}))
+	}, logOutputBuffer, func(t *testing.T, initFn witchcraft.InitFunc, installCfg config.Install, logOutputBuffer io.Writer) *witchcraft.Server {
+		installCfg.MetricsEmitFrequency = 100 * time.Millisecond
+		return createTestServer(t, initFn, installCfg, logOutputBuffer).WithMetricTypeValuesBlacklist(map[string]map[string]struct{}{
+			"histogram": {"count": {}},
+		})
+	})
+	defer func() {
+		require.NoError(t, server.Close())
+	}()
+	defer cleanup()
+
+	// Make POST that will 404 to trigger request size and error rate metrics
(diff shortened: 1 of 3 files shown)


bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/128)* status: all files reviewed, 2 unresolved discussions (waiting on @mahmoudm and @nmiyake)

___
*[witchcraft/server_metrics.go, line 27 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/128#-Lt9ykpmBQ3FPgGe3KK0:-Lt9ykpmBQ3FPgGe3KK1:br6lrc3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b35bfbd2bed299093f93a88cd7f3a5e6d1f186c8/witchcraft/server_metrics.go#L27)):*
> ```Go
> )
> 
> func defaultMetricTypeValuesBlacklist() map[string]map[string]struct{} {
> ```

Does this list match what our Java implementation does?
___
*[witchcraft/server_metrics.go, line 33 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/128#-Lt9yzB8-rwi7eaCw3pY:-Lt9yzB8-rwi7eaCw3pZ:b-2ax5) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b35bfbd2bed299093f93a88cd7f3a5e6d1f186c8/witchcraft/server_metrics.go#L33)):*
> ```Go
> 			"mean":   {},
> 			"stddev": {},
> 			"p50":    {},
> ```

Looks like this is called `median`, not `p50` (also applies to timer): https://github.com/palantir/go-metrics/blob/master/registry.go#L152

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/128)* status: all files reviewed, 2 unresolved discussions (waiting on @ashrayjain, @bmoylan, and @mahmoudm)

___
*[witchcraft/server_metrics.go, line 27 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/128#-Lt9ykpmBQ3FPgGe3KK0:-LtB0LRY15z1k15JpM7N:b66qxuk) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b35bfbd2bed299093f93a88cd7f3a5e6d1f186c8/witchcraft/server_metrics.go#L27)):*

Does this list match what our Java implementation does?
</blockquote></details>

@ashrayjain indicated that this does match the Java behavior.
___
*[witchcraft/server_metrics.go, line 33 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/128#-Lt9yzB8-rwi7eaCw3pY:-LtB0QQN7rag88ZzPLqs:b-2dbgl9) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/b35bfbd2bed299093f93a88cd7f3a5e6d1f186c8/witchcraft/server_metrics.go#L33)):*

Looks like this is called `median`, not `p50` (also applies to timer): https://github.com/palantir/go-metrics/blob/master/registry.go#L152
</blockquote></details>

The metric values that we use are translated through `github.com/palantir/pkg/metrics`, and for histograms we use `p50`: https://github.com/palantir/pkg/blob/master/metrics/value.go#L90

I've updated the tests to specifically test for the inclusion or omission of these values to verify this as well.

nmiyake (commented):
Addressed PR feedback, ready for review.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/128#-:-LtB5OuC89Qx_nxDnRAv:bphul25)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @ashrayjain, @bmoylan, and @mahmoudm)

mahmoudm (approved):

PR #129, opened 2019-11-05
Feature: add support for blacklisting metrics

### LINKED ISSUE #125: Allow selecting which go metrics you care about to emit
Currently, you can either have or not have the GoMetrics wholesale. This puts daemons in a tough spot where the cost of each additional metric is pretty high, but some of the GoMetrics are more useful than others. Can we allow selection of which GoMetrics we want vs not want?

nmiyake:
Adds WithMetricsBlacklist function to server that allows
specifying a set of metrics that should not be emitted.

Fixes #125

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)

### DIFF
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -440,7 +440,10 @@ func (s *Server) WithDisableGoRuntimeMetrics() *Server {
 	return s
 }
 
-// WithMetricsBlacklist sets the metric blacklist to the provided set of metrics. The provided input is copied.
+// WithMetricsBlacklist sets the metric blacklist to the provided set of metrics. The provided metrics should be the
+// name of the metric (for example, "server.response.size"). The blacklist only supports blacklisting at the metric
+// level: blacklisting an individual metric value (such as "server.response.size.count") will not have any effect. The
+// provided input is copied.
 func (s *Server) WithMetricsBlacklist(blacklist map[string]struct{}) *Server {
 	metricsBlacklist := make(map[string]struct{})
 	for k, v := range blacklist {
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -137,6 +137,9 @@ type Server struct {
 	// seconds if an interval is not specified in configuration).
 	disableGoRuntimeMetrics bool
 
+	// metricsBlacklist specifies the set of metrics that should not be emitted by the metric logger.
+	metricsBlacklist map[string]struct{}
+
 	// specifies the TLS client authentication mode used by the server. If not specified, the default value is
 	// tls.NoClientCert.
 	clientAuth tls.ClientAuthType
@@ -434,6 +437,16 @@ func (s *Server) WithDisableKeepAlives() *Server {
 // WithDisableGoRuntimeMetrics disables the server's enabled-by-default collection of runtime memory statistics.
 func (s *Server) WithDisableGoRuntimeMetrics() *Server {
 	s.disableGoRuntimeMetrics = true
+	return s
+}
+
+// WithMetricsBlacklist sets the metric blacklist to the provided set of metrics. The provided input is copied.
+func (s *Server) WithMetricsBlacklist(blacklist map[string]struct{}) *Server {
+	metricsBlacklist := make(map[string]struct{})
+	for k, v := range blacklist {
+		metricsBlacklist[k] = v
+	}
+	s.metricsBlacklist = metricsBlacklist
 	return s
 }
 
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -115,3 +115,85 @@ func TestEmitMetrics(t *testing.T) {
 	default:
 	}
 }
+
+// TestMetricsBlacklist verifies that blacklisted metrics are not emitted.
+func TestMetricsBlacklist(t *testing.T) {
+	logOutputBuffer := &bytes.Buffer{}
+	port, err := httpserver.AvailablePort()
+	require.NoError(t, err)
+	server, serverErr, cleanup := createAndRunCustomTestServer(t, port, port, func(ctx context.Context, info witchcraft.InitInfo) (deferFn func(), rErr error) {
+		ctx = metrics.AddTags(ctx, metrics.MustNewTag("key", "val"))
+		metrics.FromContext(ctx).Counter("my-counter").Inc(13)
+		return nil, info.Router.Post("/error", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			rw.WriteHeader(500)
+		}))
+	}, logOutputBuffer, func(t *testing.T, initFn witchcraft.InitFunc, installCfg config.Install, logOutputBuffer io.Writer) *witchcraft.Server {
+		installCfg.MetricsEmitFrequency = 100 * time.Millisecond
+		server := createTestServer(t, initFn, installCfg, logOutputBuffer)
+		server.WithMetricsBlacklist(map[string]struct{}{
+			"my-counter":          {},
+			"server.request.size": {},
+		})
+		return server
+	})
+	defer func() {
+		require.NoError(t, server.Close())
+	}()
+	defer cleanup()
+
+	// Make POST that will 404 to trigger request size and error rate metrics
+	_, err = testServerClient().Post(fmt.Sprintf("https://localhost:%d/%s/%s", port, basePath, "error"), "application/json", strings.NewReader("{}"))
+	require.NoError(t, err)
+
+	// Allow the metric emitter to do its thing.
+	time.Sleep(150 * time.Millisecond)
+
+	parts := strings.Split(logOutputBuffer.String(), "\n")
+	var metricLogs []logging.MetricLogV1
+	for _, curr := range parts {
+		if strings.Contains(curr, `"metric.1"`) {
+			var currLog logging.MetricLogV1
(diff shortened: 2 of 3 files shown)


mahmoudm (approved):

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[witchcraft/witchcraft.go, line 447 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-Lst2r815sfcH5lso0oC:-Lst2r815sfcH5lso0oD:bjzc470) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dbbc75d2b3b1c60b00f3bf509ddd04c9bff30808/witchcraft/witchcraft.go#L447)):*
> ```Go
> // WithMetricTypeValuesBlacklist sets the value of the metric type value blacklist to be the same as the provided value
> // (the content is copied).
> func (s *Server) WithMetricTypeValuesBlacklist(blacklist map[string]map[string]struct{}) *Server {
> ```

If we are expecting this to come from code rather than config, and we know the [exhaustive set](https://github.com/palantir/go-metrics/blob/master/registry.go#L127) of possible values, I wonder if we could provide a more foolproof API using structs with typed fields for the things we care about? Otherwise we should probably do validation like throwing an error when a metric type is not recognized

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 447 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-Lst2r815sfcH5lso0oC:-LswXGak4yrt06C25kNR:b-s6jsny) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dbbc75d2b3b1c60b00f3bf509ddd04c9bff30808/witchcraft/witchcraft.go#L447)):*

If we are expecting this to come from code rather than config, and we know the [exhaustive set](https://github.com/palantir/go-metrics/blob/master/registry.go#L127) of possible values, I wonder if we could provide a more foolproof API using structs with typed fields for the things we care about? Otherwise we should probably do validation like throwing an error when a metric type is not recognized
</blockquote></details>

Thanks for the feedback!

Agree with the sentiment, but main issue is figuring out how to expose the set of possible/supported values.

Technically, the emitter just deals with a `metrics.MetricVal` interface (https://github.com/palantir/pkg/blob/master/metrics/value.go#L11), so the set of possible types and values is infinite, but from a practical perspective I'm fine with ignoring that and just providing support for blacklisting "supported" metrics.

If we take an approach of only allowing blacklisting of type values we know about, I would like the set of possible values to come from the `github.com/palantir/pkg/metrics` package rather than hard-coding it in this project -- I wouldn't want to get in a situation where, if new metrics are added later, you would be forced to add them to `w-g-s`.

However, taking this approach would mean a few things:
* Adding a new exported function (API) to the `metrics` package that returns supported types and their values
* This would also require that all future metrics have a static set of possible keys for each metric type (this happens to be true right now and would likely be in the future, but there's nothing that currently requires this).

This would also require tagging a new version of `palantir/pkg`, and the new version of witchcraft would have a strict dependency on it (since it would require the new exported API).

This is *just* enough work that I'm on the fence about it... I definitely see the value here (can totally see a future situation where someone has a minor typo for a metric name/value key that makes the blacklist fail without warning), but all the workarounds assert some new state about the world as well (or requires adding code to w-g-s that creates an implicit version dependency).

Let me know what you think -- if you think the better validation with the extra dependencies makes sense, happy to open the PRs so we can evaluate that approach as well.




From a practical perspective, the set of metric types (and their values) are pretty much what is defined in https://github.com/palantir/pkg/blob/master/metrics/value.go.

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 447 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-Lst2r815sfcH5lso0oC:-LswhxtfD9-_Gux-Q65A:bxnmfuk) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dbbc75d2b3b1c60b00f3bf509ddd04c9bff30808/witchcraft/witchcraft.go#L447)):*

Thanks for the feedback!

Agree with the sentiment, but main issue is figuring out how to expose the set of possible/supported values.

Technically, the emitter just deals with a `metrics.MetricVal` interface (https://github.com/palantir/pkg/blob/master/metrics/value.go#L11), so the set of possible types and values is infinite, but from a practical perspective I'm fine with ignoring that and just providing support for blacklisting "supported" metrics.

If we take an approach of only allowing blacklisting of type values we know about, I would like the set of possible values to come from the `github.com/palantir/pkg/metrics` package rather than hard-coding it in this project -- I wouldn't want to get in a situation where, if new metrics are added later, you would be forced to add them to `w-g-s`.

However, taking this approach would mean a few things:
* Adding a new exported function (API) to the `metrics` package that returns supported types and their values
* This would also require that all future metrics have a static set of possible keys for each metric type (this happens to be true right now and would likely be in the future, but there's nothing that currently requires this).

This would also require tagging a new version of `palantir/pkg`, and the new version of witchcraft would have a strict dependency on it (since it would require the new exported API).

This is *just* enough work that I'm on the fence about it... I definitely see the value here (can totally see a future situation where someone has a minor typo for a metric name/value key that makes the blacklist fail without warning), but all the workarounds assert some new state about the world as well (or requires adding code to w-g-s that creates an implicit version dependency).

Let me know what you think -- if you think the better validation with the extra dependencies makes sense, happy to open the PRs so we can evaluate that approach as well.




From a practical perspective, the set of metric types (and their values) are pretty much what is defined in https://github.com/palantir/pkg/blob/master/metrics/value.go.


</blockquote></details>

@bmoylan sorry, I messed this up -- the code you reviewed in this PR was actually meant for the other PR (https://github.com/palantir/witchcraft-go-server/pull/128). Let's continue this discussion on that PR.

This PR is for blacklisting specific metrics by name, and I don't think that has the same issue here (since the metric names are an unbounded set since any metric can be added/logged at runtime, we don't have the same ability to validate).

witchcraft/witchcraft.go
```diff
@@ -437,6 +440,16 @@ func (s *Server) WithDisableGoRuntimeMetrics() *Server {
 	return s
 }
 
+// WithMetricsBlacklist sets the metric blacklist to the provided set of metrics. The provided input is copied.
```
bmoylan: Maybe it is worth documenting somehow that the metrics in the blacklist are those without the value suffix (`.count`, `.max`, etc)

bmoylan:
___
*[witchcraft/witchcraft.go, line 447 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-Lst2r815sfcH5lso0oC:-LtA0LTuC72-WVbeJIo6:b-oas5s6) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dbbc75d2b3b1c60b00f3bf509ddd04c9bff30808/witchcraft/witchcraft.go#L447)):*

@bmoylan sorry, I messed this up -- the code you reviewed in this PR was actually meant for the other PR (https://github.com/palantir/witchcraft-go-server/pull/128). Let's continue this discussion on that PR.

This PR is for blacklisting specific metrics by name, and I don't think that has the same issue here (since the metric names are an unbounded set since any metric can be added/logged at runtime, we don't have the same ability to validate).
</blockquote></details>

Thanks for the detailed explanation! I think I'm good with leaving it as a map since I agree plumbing the information out of go-metrics would be difficult and not the most valuable thing in the world.

bmoylan (commented):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

___
*[witchcraft/witchcraft.go, line 443 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-Lssv4IAavl9rG_alkWo-r2-443:-LtB4JRe8EGMbMIzC1vK:b-n9m51z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/aa65686862daa0518ed8da1bce1ede121c928c0b/witchcraft/witchcraft.go#L443)):*

Maybe it is worth documenting somehow that the metrics in the blacklist are those without the value suffix (`.count`, `.max`, etc)
</blockquote></details>

Good point -- will update.

nmiyake (commented):
Addressed PR feedback, ready for review.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/129#-:-LtB5I5_6srFXHn6zAbT:bphul25)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @bmoylan)

mahmoudm (approved):

mahmoudm:
@nmiyake can you fix conflicts please

PR #131, opened 2019-11-05
Add a configuration point for the servers metrics registry emitter

andybradshaw:
Allow users to provide a custom metrics visitor by writing a `MetricsVisitorProvider`. This would allow users to limit the values or tags emitted for metrics which might have prohibitively high cardinalities.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/131)

### DIFF
--- a/witchcraft/server_metrics.go
+++ b/witchcraft/server_metrics.go
@@ -24,7 +24,15 @@ import (
 	"github.com/palantir/witchcraft-go-server/config"
 )
 
-func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rRegistry metrics.RootRegistry, rDeferFn func(), rErr error) {
+type MetricsVisitorProvider func(s *Server) metrics.MetricVisitor
+
+func DefaultMetricEmitter(s *Server) metrics.MetricVisitor {
+	return func(metricID string, tags metrics.Tags, metricVal metrics.MetricVal) {
+		s.metricLogger.Metric(metricID, metricVal.Type(), metric1log.Values(metricVal.Values()), metric1log.Tags(tags.ToMap()))
+	}
+}
+
+func (s *Server) initMetrics(ctx context.Context, installCfg config.Install, visitor metrics.MetricVisitor) (rRegistry metrics.RootRegistry, rDeferFn func(), rErr error) {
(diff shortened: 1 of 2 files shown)


andybradshaw (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/131)* status: 0 of 2 files reviewed, all discussions resolved

___
*[witchcraft/server_metrics.go, line 29 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/131#-LsxOAyVEPI6wVuC1gFW:-LsxOAyVEPI6wVuC1gFX:bs91yt3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c502faf8596261b0a9b438b291440bebbf9d0efe/witchcraft/server_metrics.go#L29)):*
> ```Go
> type MetricsVisitorProvider func(s *Server) metrics.MetricVisitor
> 
> func DefaultMetricEmitter(s *Server) metrics.MetricVisitor {
> ```

I think it makes sense that this is the only emitter defined here. I was trying to think of a sane "metric emitter lite", but couldn't come up with something that worked in all scenarios. Thought it would be better to just let the consumer decide.

nmiyake:
Would #125 and #127 address this? Those allow trimming down emitted metrics (although they don't take tags into account). I have a slight preference for those approaches over this one since I feel like the other approaches allow for configuring/controlling output in a more standard manner.

andybradshaw:
Hmm... I think those PRs will effectively achieve the same goal, but I don't think they're necessarily incompatible approaches. You could give users a helper for creating a "metric blacklist" visitor or a "type value blacklist" visitor, even making those the default, while still letting people provide their own home-rolled thing in the cases where they need to off-road.

nmiyake:
Is there a concrete use case you have in mind that isn't covered by those configuration points? Trying to understand how this is intended to be used. If the other approaches address our current issues then would prefer not to add a generic customization point (adding configurability has a cost in terms of complexity/conceptual overhead). However, if there's a compelling use case then certainly open to doing this.

andybradshaw:
Don't have a concrete use case, but also don't think the complexity/overhead is that high and it gives people more control if they need it. Fine with using the other PRs and seeing how that goes, don't really have strong opinions here.

PR #134, opened 2019-11-15
Add validating refreshable

### LINKED ISSUE #105: Runtime config refreshable produces zero value if cannot parse yaml
If `var/conf/runtime.yml` starts as valid then is changed to something that cannot be parsed as YAML into the specified `RuntimeConfigType`, the runtime config refreshable provided in the `InitFunc` starts returning a zero value runtime config type.

Instead , it should return the previous value from the version of the file that was parseable.

Additionally, it might be worth adding a health check that verifies the current version of the config on disk is parseable and in use.

Here's a program that reproduces this behavior:

```go
package main

import (
	"context"
	"fmt"
	"github.com/palantir/witchcraft-go-server/witchcraft"
	"io/ioutil"
	"os"
	"time"
)

type RuntimeConfig struct {
	Foo int
}
/*
This program starts a simple withcraft server that prints the value of its runtime config's Foo field every second.

It starts with a valid config, then after three seconds changes the file on disk to be an invalid config, after which
the runtime config refreshable produces a zero value for the field rather than the previous valid value, which seems
wrong.

	$ go run main.go
	{"level":"INFO","time":"2019-09-09T19:19:39.165365Z","message":"Listening to https","type":"service.1","origin":"github.palantir.build/jhenke/repro-type-bug","params":{"address":"127.0.0.1:0","server":""}}
	Foo 42
	Foo 42
	Foo 42
	Wrote new config file
	{"level":"ERROR","time":"2019-09-09T19:19:42.100559Z","message":"Failed to unmarshal runtime configuration","type":"service.1","origin":"github.palantir.build/jhenke/repro-type-bug","stacktrace":"yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `some-ki...` into int"}
	Foo 0
	Foo 0
	Foo 0
	exit status 1
 */
func main() {
	if err := os.MkdirAll("var/conf", 0755); err != nil {
		panic(err)
	}
	if err := ioutil.WriteFile("var/conf/install.yml", []byte("use-console-log: true\nserver:\n  address: 127.0.0.1\n"), 0644); err != nil {
		panic(err)
	}
	if err := ioutil.WriteFile("var/conf/runtime.yml", []byte("foo: 42\n"), 0644); err != nil {
		panic(err)
	}
	go func() {
		time.Sleep(3 * time.Second)
		if err := ioutil.WriteFile("var/conf/runtime.yml", []byte("foo: some-kinda-text\n"), 0644); err != nil {
			panic(err)
		}
		fmt.Println("Wrote new config file")
		time.Sleep(3 * time.Second)
		os.Exit(1)
	}()
	if err := witchcraft.NewServer().
		WithSelfSignedCertificate().
		WithRuntimeConfigType(RuntimeConfig{}).
		WithInitFunc(func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
			go func() {
				for range time.Tick(1 * time.Second) {
					fmt.Println("Foo", info.RuntimeConfig.Current().(RuntimeConfig).Foo)
				}
			}()
			return nil, nil
		}).Start(); err != nil {
			panic(err)
	}
}
```

ashrayjain:
Fixes https://github.com/palantir/witchcraft-go-server/issues/105

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134)

### DIFF
--- a/changelog/@unreleased/pr-134.v2.yml
+++ b/changelog/@unreleased/pr-134.v2.yml
@@ -1,0 +1,5 @@
+type: break
+break:
+  description: Install and Runtime configuration is now unmarshalled using UnmarshalStrict as opposed to Unmarshal. Additionally, a health check is registered that indicates whether Runtime configuration reloading is blocked due to invalid configuration.
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/134
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -97,7 +97,7 @@ func createAndRunCustomTestServer(t *testing.T, port, managementPort int, initFn
 	require.NoError(t, err)
 	err = ioutil.WriteFile(installYML, installCfgYML, 0644)
 	require.NoError(t, err)
-	err = ioutil.WriteFile(runtimeYML, []byte(`number: 1`), 0644)
+	err = ioutil.WriteFile(runtimeYML, []byte(`logging:\n  level: info`), 0644)
 	require.NoError(t, err)
 
 	server = createServer(t, initFn, installCfg, logOutputBuffer)
--- a/integration/config_test.go
+++ b/integration/config_test.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/palantir/witchcraft-go-server/config"
 	"github.com/palantir/witchcraft-go-server/witchcraft"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -36,7 +37,8 @@ func TestEncryptedConfig(t *testing.T) {
 	)
 
 	type message struct {
-		Message string `yaml:"message"`
+		config.Runtime `yaml:",inline"`
+		Message        string `yaml:"message"`
 	}
 
 	for _, test := range []struct {
--- a/integration/config_test.go
+++ b/integration/config_test.go
@@ -41,6 +41,11 @@ func TestEncryptedConfig(t *testing.T) {
 		Message        string `yaml:"message"`
 	}
 
+	type messageInstall struct {
+		config.Install `yaml:",inline"`
+		Message        string `yaml:"message"`
+	}
+
 	for _, test := range []struct {
 		Name          string
 		ECVKeyContent string
@@ -56,7 +61,7 @@ func TestEncryptedConfig(t *testing.T) {
 			InstallConfig: fmt.Sprintf("message: %s\nuse-console-log: true\n", encryptedValue),
 			RuntimeConfig: fmt.Sprintf("message: %s\nlogging:\n  level: warn\n", encryptedValue),
 			Verify: func(info witchcraft.InitInfo) error {
-				if msg := info.InstallConfig.(*message).Message; msg != decryptedValue {
+				if msg := info.InstallConfig.(*messageInstall).Message; msg != decryptedValue {
 					return fmt.Errorf("expected %q got %q", decryptedValue, msg)
 				}
 				if msg := info.RuntimeConfig.Current().(*message).Message; msg != decryptedValue {
@@ -74,7 +79,7 @@ func TestEncryptedConfig(t *testing.T) {
 			InstallConfig: "message: hello install\nuse-console-log: true\n",
 			RuntimeConfig: "message: hello runtime\nlogging:\n  level: warn\n",
 			Verify: func(info witchcraft.InitInfo) error {
-				if msg := info.InstallConfig.(*message).Message; msg != "hello install" {
+				if msg := info.InstallConfig.(*messageInstall).Message; msg != "hello install" {
 					return fmt.Errorf("expected %q got %q", "hello install", msg)
 				}
 				if msg := info.RuntimeConfig.Current().(*message).Message; msg != "hello runtime" {
@@ -92,7 +97,7 @@ func TestEncryptedConfig(t *testing.T) {
 			InstallConfig: "message: hello install\nuse-console-log: true\n",
 			RuntimeConfig: fmt.Sprintf("message: %s\nlogging:\n  level: warn\n", encryptedValue),
 			Verify: func(info witchcraft.InitInfo) error {
-				if msg := info.InstallConfig.(*message).Message; msg != "hello install" {
+				if msg := info.InstallConfig.(*messageInstall).Message; msg != "hello install" {
 					return fmt.Errorf("expected %q got %q", "hello install", msg)
 				}
 				if msg := info.RuntimeConfig.Current().(*message).Message; msg != encryptedValue {
@@ -125,7 +130,7 @@ func TestEncryptedConfig(t *testing.T) {
 			server := witchcraft.NewServer().
 				WithECVKeyFromFile(ecvKeyFile).
 				WithInstallConfigFromFile(installFile).
-				WithInstallConfigType(&message{}).
+				WithInstallConfigType(&messageInstall{}).
 				WithRuntimeConfigFromFile(runtimeFile).
 				WithRuntimeConfigType(&message{}).
 				WithLoggerStdoutWriter(logOutputBuffer).
--- a/integration/health_test.go
+++ b/integration/health_test.go
@@ -427,7 +427,7 @@ invalid-key: invalid-value
 				Type:    health.CheckType("CONFIG_RELOAD"),
 				State:   health.HealthStateError,
 				Params:  make(map[string]interface{}),
-				Message: stringPtr("yaml: unmarshal errors:\n  line 2: field invalid-key not found in type config.Runtime"),
+				Message: stringPtr("Refreshable validation failed, please look at service logs for more information."),
 			},
 			health.CheckType("SERVER_STATUS"): {
 				Type:   health.CheckType("SERVER_STATUS"),
--- a/integration/health_test.go
+++ b/integration/health_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/palantir/witchcraft-go-server/status/health/periodic"
 	"github.com/palantir/witchcraft-go-server/status/reporter"
 	"github.com/palantir/witchcraft-go-server/witchcraft"
+	"github.com/palantir/witchcraft-go-server/witchcraft/refreshable"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -72,6 +73,11 @@ func TestAddHealthCheckSources(t *testing.T) {
 				State:   health.HealthStateHealthy,
 				Message: nil,
 				Params:  make(map[string]interface{}),
+			},
+			health.CheckType("CONFIG_RELOAD"): {
+				Type:   health.CheckType("CONFIG_RELOAD"),
+				State:  health.HealthStateHealthy,
+				Params: make(map[string]interface{}),
 			},
 			health.CheckType("SERVER_STATUS"): {
 				Type:    health.CheckType("SERVER_STATUS"),
@@ -187,6 +193,11 @@ func TestHealthReporter(t *testing.T) {
 				Message: &errString,
 				Params:  make(map[string]interface{}),
 			},
+			health.CheckType("CONFIG_RELOAD"): {
+				Type:   health.CheckType("CONFIG_RELOAD"),
+				State:  health.HealthStateHealthy,
+				Params: make(map[string]interface{}),
+			},
 			health.CheckType("SERVER_STATUS"): {
 				Type:   health.CheckType("SERVER_STATUS"),
 				State:  reporter.HealthyState,
@@ -235,6 +246,11 @@ func TestPeriodicHealthSource(t *testing.T) {
 			State:   health.HealthStateError,
 			Message: stringPtr("No successful checks during 1m0s grace period: something went wrong"),
 			Params:  map[string]interface{}{"foo": "bar"},
+		},
+		health.CheckType("CONFIG_RELOAD"): {
+			Type:   health.CheckType("CONFIG_RELOAD"),
+			State:  health.HealthStateHealthy,
+			Params: make(map[string]interface{}),
 		},
 		health.CheckType("SERVER_STATUS"): {
 			Type:    health.CheckType("SERVER_STATUS"),
@@ -314,6 +330,11 @@ func TestHealthSharedSecret(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, health.HealthStatus{
 		Checks: map[health.CheckType]health.HealthCheckResult{
+			health.CheckType("CONFIG_RELOAD"): {
+				Type:   health.CheckType("CONFIG_RELOAD"),
+				State:  health.HealthStateHealthy,
+				Params: make(map[string]interface{}),
+			},
 			health.CheckType("SERVER_STATUS"): {
 				Type:   health.CheckType("SERVER_STATUS"),
 				State:  reporter.HealthyState,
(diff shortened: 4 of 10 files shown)


ashrayjain:
@nmiyake @bmoylan could you please take a look at this?

nmiyake:
Will review by EOD Monday

nmiyake (changes requested):
Thanks for the PR! Over-all, definitely in sync with the functionality implemented. Some minor feedback/questions on implementation.

At a higher level, it seems that this validation is only added for runtime configuration. Should it be added for install configuration as well? I know that's not refreshable, but if we're performing strict YAML validation for runtime configuration, seems inconsistent not to do it for install configuration as well.

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-:-LtvTIvaAHCTRQwrT9mN:b-vk6v8f)* status: all files reviewed, 4 unresolved discussions (waiting on @ashrayjain)

___
*[witchcraft/refreshable/refreshable_validating.go, line 36 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvQFvy7pHTE2s_5ehY:-LtvQFvzBCS0z4Vi3tXm:b-i1n44g) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L36)):*
> ```Go
> 
> 	err := v.lastValidateErr.Load()
> 	if err != nil {
> ```

nit: inline err assignment and check so err is scoped more tightly:

```go
if err := v.lastValidateErr.Load(); err != nil {
```
___
*[witchcraft/refreshable/refreshable_validating.go, line 61 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvQRbKAAmOAwGcJRmf:-LtvQRbKAAmOAwGcJRmg:b-rc8rqj) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L61)):*
> ```Go
> // NewValidatingRefreshable returns a new Refreshable whose current value is the latest value that passes the provided
> // validatingFn successfully. This refreshable is also a HealthCheckSource that will be unhealthy whenever there are
> // updates that have failed validation.
> ```

Documentation should note that this function will return an error if the current value of the original refreshable is not valid according to the validation function at the time the function is called
___
*[witchcraft/refreshable/refreshable_validating.go, line 64 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvQe-pB6BCwaCtL7T7:-LtvQe-pB6BCwaCtL7T8:byvkqe3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L64)):*
> ```Go
> func NewValidatingRefreshable(origRefreshable Refreshable, healthCheckType health.CheckType, validatingFn func(interface{}) error) (*ValidatingRefreshable, error) {
> 	currentVal := origRefreshable.Current()
> 	if err := validatingFn(currentVal); err != nil {
> ```

This will panic if the provided `validatingFn` is `nil` -- since this function returns an error anyway, better to check if `validatingFn` is `nil` first and return an error if that is the case.
___
*[witchcraft/refreshable/refreshable_validating.go, line 77 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvRLIG44GTc1Q9IFnx:-LtvRLIG44GTc1Q9IFny:bpdubsl) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L77)):*
> ```Go
> 	}
> 
> 	_ = origRefreshable.Subscribe(func(i interface{}) {
> ```

A bit on the fence about ignoring the unsubscribe here. Understand that this is probably fine from a practical perspective, but from an API perspective I think it would be a bit cleaner to store the function and allow clients to unsubscribe when they know the validated refreshable is no longer needed (otherwise, I think it's also impossible for these objects to ever get cleaned up). However, even with this approach, there are questions about what the right behavior would be for functions like `Current()` etc. (since those functions don't typically return errors).

ashrayjain (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134)* status: 6 of 7 files reviewed, 3 unresolved discussions (waiting on @jdhenke and @nmiyake)

___
*[witchcraft/refreshable/refreshable_validating.go, line 61 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvQRbKAAmOAwGcJRmf:-Ltv_TqL9798s2UktK75:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L61)):*

Documentation should note that this function will return an error if the current value of the original refreshable is not valid according to the validation function at the time the function is called
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_validating.go, line 64 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvQe-pB6BCwaCtL7T7:-Ltv_UeE9RJgeGH2Cf39:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L64)):*

This will panic if the provided `validatingFn` is `nil` -- since this function returns an error anyway, better to check if `validatingFn` is `nil` first and return an error if that is the case.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_validating.go, line 77 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvRLIG44GTc1Q9IFnx:-Ltv_fuYAD0T40dlxxG4:bpd8y9x) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L77)):*

A bit on the fence about ignoring the unsubscribe here. Understand that this is probably fine from a practical perspective, but from an API perspective I think it would be a bit cleaner to store the function and allow clients to unsubscribe when they know the validated refreshable is no longer needed (otherwise, I think it's also impossible for these objects to ever get cleaned up). However, even with this approach, there are questions about what the right behavior would be for functions like `Current()` etc. (since those functions don't typically return errors).
</blockquote></details>

Yea, i was also not sure what to do here. I will point out that the DefaultRefreshable#Map() also does the same. 
The best i could come with here was to have the constructor return a cleanup function but it's unclear how a consumer would utilise this.

bmoylan (commented):
> Reviewed 5 of 7 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134)* status: all files reviewed, 6 unresolved discussions (waiting on @ashrayjain, @jdhenke, and @nmiyake)

___
*[witchcraft/witchcraft.go, line 748 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtyaWrx9reBxZicL0zK:-LtyaWry7QwE-2zQPYn-:b-ffgfu4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/98d526ca6e5e7015327d02911670a765127e5a2e/witchcraft/witchcraft.go#L748)):*
> ```Go
> 		runtimeConfigProvider,
> 		runtimeConfigReloadCheckType,
> 		func(cfgBytesVal interface{}) error {
> ```

It feels odd to have to duplicate this unmarshalling logic and run it twice per update (once to validate, again to map to a useful value)

I wonder if a better API for ValidatingRefreshable would be to add a `MapWithValidation(func (i interface{}) error)` or something whose map function can return an error to trigger the behavior when the validation fails
___
*[witchcraft/witchcraft.go, line 775 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-Ltyaf077OPNwa-dHVHg:-Ltyaf077OPNwa-dHVHh:b5ydecr) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/98d526ca6e5e7015327d02911670a765127e5a2e/witchcraft/witchcraft.go#L775)):*
> ```Go
> 		runtimeCfg := reflect.New(reflect.TypeOf(runtimeConfigStruct)).Interface()
> 		if err := yaml.UnmarshalStrict(cfgBytesVal.([]byte), *&runtimeCfg); err != nil {
> 			s.svcLogger.Error("Failed to unmarshal runtime configuration", svc1log.Stacktrace(err))
> ```

we could make this panic since we already validated we can unmarshal and it would be indicative of a programming bug in witchcraft
___
*[witchcraft/refreshable/refreshable_validating.go, line 26 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtyDlT262IA4UKbdM2v:-LtyDlT262IA4UKbdM2w:bc32o4a) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/98d526ca6e5e7015327d02911670a765127e5a2e/witchcraft/refreshable/refreshable_validating.go#L26)):*
> ```Go
> )
> 
> type ValidatingRefreshable struct {
> ```

Can we assert somewhere the two interfaces this struct implements? Could be as simple as
```go
var _ Refreshable = (*ValidatingRefreshable)(nil)
var _ status.HealthCheckSource = (*ValidatingRefreshable)(nil)
```
or we could change this to

```go
type ValidatingRefreshable interface {
	status.HealthCheckSource
	Refreshable
}

type validatingRefreshable struct { ... }
```

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134)* status: all files reviewed, 4 unresolved discussions (waiting on @ashrayjain, @jdhenke, and @nmiyake)

___
*[witchcraft/witchcraft.go, line 748 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtyaWrx9reBxZicL0zK:-Ltz07lr8cl_C5x7bVDD:buhwnrl) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/98d526ca6e5e7015327d02911670a765127e5a2e/witchcraft/witchcraft.go#L748)):*

It feels odd to have to duplicate this unmarshalling logic and run it twice per update (once to validate, again to map to a useful value)

I wonder if a better API for ValidatingRefreshable would be to add a `MapWithValidation(func (i interface{}) error)` or something whose map function can return an error to trigger the behavior when the validation fails
</blockquote></details>

Retracting this after discussing offline
___
*[witchcraft/refreshable/refreshable_validating.go, line 29 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-Ltz49qK5QbJ_mjypvOW:-Ltz49qK5QbJ_mjypvOX:bm0v3fv) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e4204b51220fba48c2ed24049f9abafccc65dffe/witchcraft/refreshable/refreshable_validating.go#L29)):*
> ```Go
> type ValidatingRefreshable struct {
> 	Refreshable
> 	status.HealthCheckSource
> ```

apparently this is an import cycle :frowning_face:

nmiyake:
@jdhenke would you mind taking a quick look at this sometime today?

bmoylan (commented):
> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134)* status: all files reviewed, 4 unresolved discussions (waiting on @ashrayjain, @jdhenke, and @nmiyake)

___
*[status/health/refreshable/source.go, line 37 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-Lu-GlKs9totz0MD4iz6:-Lu-GlKtDhplm41JxsWq:bd5ohwi) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a3477a78423de2ead6eb205868dac1353038b9cb/status/health/refreshable/source.go#L37)):*
> ```Go
> 
> 	if err := v.refreshable.LastValidateErr(); err != nil {
> 		healthCheckResult = whealth.UnhealthyHealthCheckResult(v.healthCheckType, err.Error())
> ```

:rotating_light:  Could this ever be unsafe? `failed to parse as struct 'secret-customer-name'` or something?
___
*[witchcraft/refreshable/refreshable_validating.go, line 77 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-LtvRLIG44GTc1Q9IFnx:-Lu-Fi5z82r5D8tOk14W:bgj8y1b) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/0ce21d673b577d2a93a9a1ee141d4f8dd4c5f76b/witchcraft/refreshable/refreshable_validating.go#L77)):*
<details><summary><i>Previously, ashrayjain (Ashray Jain) wrote…</i></summary><blockquote>

Yea, i was also not sure what to do here. I will point out that the DefaultRefreshable#Map() also does the same. 
The best i could come with here was to have the constructor return a cleanup function but it's unclear how a consumer would utilise this.
</blockquote></details>

A conversation for another PR, but I wonder if Refreshable should have a `Close` method (or `Unregister` or `Stop` or w/e) that unsubscribes it rather than forcing the caller to call it.

The docs can encourage `defer r.Close()` (replacing this unsusbcribe return value) but it is admittedly less forceful than a returned cleanup value.

bmoylan:
___
*[witchcraft/refreshable/refreshable_validating.go, line 79 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-Lu-IOxB1E6mNQoHbiCv:-Lu-IOxB1E6mNQoHbiCw:b-rrz7z2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/a3477a78423de2ead6eb205868dac1353038b9cb/witchcraft/refreshable/refreshable_validating.go#L79)):*
> ```Go
> 	_ = origRefreshable.Subscribe(func(i interface{}) {
> 		if err := validatingFn(i); err != nil {
> 			v.lastValidateErr.Store(errorWrapper{err})
> ```

Should we log the error here?

bmoylan (approved):
Ok, I think this is the best we can do on the logging front and not significantly different from what we do elsewhere :+1: 

> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-:-Lu2Nmgu8cGhk7BTf6K3:b-rcd2w8)* status: all files reviewed, 3 unresolved discussions (waiting on @jdhenke and @nmiyake)

bmoylan:
@nmiyake I've approved this; are you good with going to 2.0 due to the behavior break?

nmiyake:
Will review before 12pm Pacific today and plan to merge unless I have any major concerns (don't anticipate it, but would like to do final verification).

Since this changes only behavior, I would like to avoid a 2.0 if necessary, even if this means exposing a code-configuration option to disable this behavior on the server (or making this behavior opt-in until a 2.0 when we switch it to default).

nmiyake (approved):
Thanks for the PR -- I'm good with the content and implementation.

Good to merge, but would like to resolve plan for release. Options are:

(1) Merge as-is and bump minor version with documentation that default install/runtime config unmarshal behavior has changed
(2) Merge as-is and bump major version
(3) Merge this functionality but also add code-level configuration option to disable strict unmarshal and bump minor version with documentation
(4) Make this new functionality opt-in via code-level configuration option and bump minor version with documentation

(4) is the safest for consumers (guaranteed not to break any existing clients on bump). However, if we have strong conviction that strict unmarshal is generally correct (and that current clients don't violate this), we could move forward with (1) or (3). We could always do (1) first and introduce (3) only if we run into a consumer that requires it. However, for these options we should pay close attention to the health check value.

I don't think bumping the major version buys us much -- clients have to choose to update their dependency version either way. Also, even if we major bump, I don't want to back-port new fixes/features to 1.x, which means that as soon as there's an important bug we fix or feature we implement, consumers will have to move forward anyway. With the module transition being imminent as well, I would like to have major version bumps follow semantic version importing rules (so new import path), and don't think it's worth doing that for this change.

> Reviewed 2 of 4 files at r2, 2 of 4 files at r4, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/134#-:-Lu3iqHHDewNmOWQixeq:b8vvrvi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jdhenke)

ashrayjain:
I'm fine with all the options but have a preference for not making this optional. I think we should be opinionated here and avoid the spread of functionality we have to support.

bmoylan:
Yeah I would assert that (a) this behavior is more correct and prevents latent P0s  but also (b) there exist violations in the wild today that will P0 in the short term and/or cause upgrade failures which are automatically rolled back by the infra. I can't think of "legit" reasons for violation, so the remediation would be to fix the bad configuration rather than roll back the code change.

nmiyake:
Yup, in sync. We can always re-evaluate exposing knobs based on what we find from the field, but good with moving forward with merge + minor version bump with documentation and dealing with things from there (we can send out comms as well if we want to be extra careful).

nmiyake:
@bmoylan @ashrayjain, @jdhenke just took a look and flagged a possible race (which we should fix before tagging) and brought up some things to think about (don't think we need to change based on those things, but would like them to be properly considered by group here). He's going to comment here shortly.

nmiyake:
We synced up in person and agreed on the following:

* Moving to strict-by-default with no opt-out is likely too aggressive, especially when we don't have any concrete data on what impact this will have on stacks (and a product that fails strict config validation will fail to start)
* However, strict unmarshal is something we want to move towards: beyond just being a best practice, there are concrete bugs (especially in discovery) that can manifest due to incorrect values existing

Based on this, our plan is as follows:

* Keep this feature, but make it opt-in via code configuration
* Keep the health check in all scenarios. If configuration fails strict unmarhsal, note this as a WARN. This will help us collect data on what the impact of this change would actually be.

This allows consumers that want the feature to opt-in, while also allowing us to collect concrete data on the impact of this change (even for consumers that do not enable the feature).

Based on this data, we can decide whether it makes sense to change this feature to be opt-out (or enabled with no opt-out) in the future.

PR #141, opened 2019-11-19
Optimize window based health check source memory consumption

splucs:
## Before this PR
The "Healthy if not all errors" and "Unhealthy if at least one error" and its multi key versions health checks have a memory consumption of `O(number of events in the window)`. Although this is required for the generic BaseWindowHealthCheckSource, it is not required for these ones.

## After this PR
Optimized the mentioned health sources to work in `O(1)` memory for the single key versions and `O(min(number of events in the window, number of keys))` memory for the multi key versions.
Changed the "Unhealthy if at least one error" health checks sources to forward calls to "Healthy if not all errors" sources. This greatly simplifies the code.
Added a TimedKeyStore helper that can be used on custom implementations.
Changed the error message shown to be the last in the window instead of the first.
==COMMIT_MSG==
Optimize window based health check source memory consumption. Error message shown is now the last in the window and not the first.
==COMMIT_MSG==

## Possible downsides?
Higher overhead (complexity multiplicative constant) due to the double linked list.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/141)

### DIFF
--- a/status/health/window/error_source.go
+++ b/status/health/window/error_source.go
@@ -16,8 +16,10 @@ package window
 
 import (
 	"context"
+	"sync"
 	"time"
 
+	werror "github.com/palantir/witchcraft-go-error"
 	"github.com/palantir/witchcraft-go-server/conjure/witchcraft/api/health"
 	"github.com/palantir/witchcraft-go-server/status"
 	whealth "github.com/palantir/witchcraft-go-server/status/health"
@@ -38,8 +40,8 @@ type ErrorHealthCheckSource interface {
 // It returns the first non-nil error as an unhealthy check.
 // If there are no items, returns healthy.
 type unhealthyIfAtLeastOneErrorSource struct {
-	timeWindowedStore *TimeWindowedStore
-	checkType         health.CheckType
+	// unhealthyIfAtLeastOneErrorSource is a healthyIfNotAllErrorsSource that drops all successes.
+	source ErrorHealthCheckSource
 }
 
 // MustNewUnhealthyIfAtLeastOneErrorSource returns the result of calling NewUnhealthyIfAtLeastOneErrorSource, but panics if it returns an error.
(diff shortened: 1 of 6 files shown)


bmoylan (changes requested):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/141)* status: all files reviewed, 5 unresolved discussions (waiting on @splucs)

___
*[status/health/window/error_source.go, line 138 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1r2pl4qzcOnI76XHo:-Lu1r2pl4qzcOnI76XHp:b-dad3wj) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/error_source.go#L138)):*
> ```Go
> 	if !h.lastSuccessTime.IsZero() && curTime.Sub(h.lastSuccessTime) < h.windowSize {
> 		healthCheckResult = whealth.HealthyHealthCheckResult(h.checkType)
> 	} else if h.lastError != nil && !h.lastErrorTime.IsZero() && curTime.Sub(h.lastErrorTime) < h.windowSize {
> ```

is there any case where `h.lastError != nil && listErrorTime.IsZero()`? I don't think you need this middle case unless you think it helps readability
___
*[status/health/window/key_error_source.go, line 38 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1sXrI1LHV2DYQ2BH1:-Lu1sXrI1LHV2DYQ2BH2:b1bhtrg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L38)):*
> ```Go
> 
> // keyErrorPair is a struct that keeps a key as a string and an err.
> type keyErrorPair struct {
> ```

deadcode is failing because this is unused
___
*[status/health/window/key_error_source.go, line 140 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1toeHER3GaPhWiO8K:-Lu1toeHER3GaPhWiO8L:bqm031p) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L140)):*
> ```Go
> 	defer m.sourceMutex.Unlock()
> 
> 	m.pruneOldKeys(m.errorStore, m.lastError)
> ```

Why do we have to prune here if we're going to do it again in `HealthStatus`?
___
*[status/health/window/key_error_source.go, line 144 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1uBqx5NPDUypuk3bP:-Lu1uBqx5NPDUypuk3bQ:b-8uxppk) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L144)):*
> ```Go
> 
> 	if err == nil {
> 		m.successStore.Add(key)
> ```

Could we clear the the lastError for this key (`delete(m.lastError, key)`) so the garbage collector can get it? It seems that if there's a success we don't care about the old error anymore
___
*[status/health/window/timed_key_store.go, line 45 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1oe7I-cbojXTcFv_7:-Lu1oe7I-cbojXTcFv_8:bbq3xn5) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/timed_key_store.go#L45)):*
> ```Go
> 	// Add adds a new TimedKey to the end of the list with the timestamp set to the current time.
> 	// Adding an already present key will cause the current TimedKey to be updated to the current and to be sent to the end of the list.
> 	Add(key string)
> ```

nit: I might call this `Put` since it overwrites an existing value if it exists

splucs (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/141)* status: 1 of 6 files reviewed, 4 unresolved discussions (waiting on @bmoylan)

___
*[status/health/window/error_source.go, line 138 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1r2pl4qzcOnI76XHo:-Lu29fPMERZqr4hJ26Vk:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/error_source.go#L138)):*

is there any case where `h.lastError != nil && listErrorTime.IsZero()`? I don't think you need this middle case unless you think it helps readability
</blockquote></details>

Done.
___
*[status/health/window/key_error_source.go, line 38 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1sXrI1LHV2DYQ2BH1:-Lu29fcd7n0BBORzDIkY:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L38)):*

deadcode is failing because this is unused
</blockquote></details>

Done.
___
*[status/health/window/key_error_source.go, line 140 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1toeHER3GaPhWiO8K:-Lu29gPSE101BWiTbCWO:bjrmsja) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L140)):*

Why do we have to prune here if we're going to do it again in `HealthStatus`?
</blockquote></details>

If the consumer of the health check is offline, we will never prune.
___
*[status/health/window/key_error_source.go, line 144 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-Lu1uBqx5NPDUypuk3bP:-Lu2A1C6DHJv4aMOCbzJ:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fbc337a57b53299f325423c3c6f468fdc1654cc/status/health/window/key_error_source.go#L144)):*

Could we clear the the lastError for this key (`delete(m.lastError, key)`) so the garbage collector can get it? It seems that if there's a success we don't care about the old error anymore
</blockquote></details>

Done.

bmoylan (approved):
needs changelog

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/141#-:-Lu2HwNW7RUfDmn771pO:b-2w4vdg)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #142, opened 2019-11-22
Feature: add configuration to set unmarshal mode to strict

nmiyake:
When configuration is set to true, configuration unmarshal is
done in "strict" mode, which treats unknown keys as errors.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/142)

### DIFF
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -167,6 +167,10 @@ type Server struct {
 	// disableKeepAlives disables keep-alives.
 	disableKeepAlives bool
 
+	// if true, strict YAML unmarshaling is used when unmarshaling install and runtime configuration. This means that
+	// unknown keys will be treated as errors rather than being ignored.
+	strictUnmarshalConfig bool
+
 	// request logger configuration
 
 	// idsExtractor specifies the extractor used to extract identifiers (such as UID, SID, TokenID) from requests for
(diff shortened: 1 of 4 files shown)


witchcraft/witchcraft.go
```diff
@@ -440,6 +444,12 @@ func (s *Server) WithDisableKeepAlives() *Server {
 	return s
 }
 
+// WithStrictUnmarshalConfig configures the server to use the provided strict unmarshal configuration.
+func (s *Server) WithStrictUnmarshalConfig(strictUnmarshalConfig bool) *Server {
```
ashrayjain: can we make this be two methods `WithStrictUnmarshalConfig` and `WithoutStrictUnmarshalConfig` instead and no params? slight preference for APIs that don't involve boolean parameters.

witchcraft/witchcraft.go
```diff
@@ -744,6 +760,11 @@ func (s *Server) initRuntimeConfig(ctx context.Context) (rBaseCfg refreshableBas
 		return cfgBytes
 	})
 
+	unmarshalYAMLFn := yaml.Unmarshal
```
ashrayjain: can we add `unmarshalYAMLFn` as a field on witchcraft object instead of `strictUnmarshalConfig bool`? and then use that everywhere?

nmiyake: Sure. I don't think we should allow it to be customized beyond these values for now (without a more comprehensive change), but since this is fully controlled in code that should be fine.

witchcraft/witchcraft.go
nmiyake: Sure -- for now, just making this `WithStrictUnmarshalConfig`. If we change default to strict, we can introduce the `Without` variant.

nmiyake:
Addressed PR feedback

ashrayjain (approved):

nmiyake:
@bmoylan can you take a look when you get a chance?

bmoylan (approved):
> Reviewed 2 of 4 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/142)* status: all files reviewed, 2 unresolved discussions (waiting on @ashrayjain and @jdhenke)

PR #143, opened 2019-11-22
Fix: fix race condition in NewValidatingRefreshable

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/143)

### DIFF
--- a/changelog/@unreleased/pr-143.v2.yml
+++ b/changelog/@unreleased/pr-143.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix race condition in NewValidatingRefreshable
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/143
--- a/witchcraft/refreshable/refreshable_validating.go
+++ b/witchcraft/refreshable/refreshable_validating.go
@@ -74,7 +74,7 @@ func NewValidatingRefreshable(origRefreshable Refreshable, validatingFn func(int
 		lastValidateErr:      &lastValidateErr,
 	}
 
-	_ = origRefreshable.Subscribe(func(i interface{}) {
+	updateValueFn := func(i interface{}) {
 		if err := validatingFn(i); err != nil {
 			v.lastValidateErr.Store(errorWrapper{err})
 			return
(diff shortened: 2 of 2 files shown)


ashrayjain (changes requested):
Can we add a test for this race condition?

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/143)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bmoylan and @jdhenke)

___
*[witchcraft/refreshable/refreshable_validating.go, line 95 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/143#-LuG2SeA-PgXJFozC1a4:-LuG2SeA-PgXJFozC1a5:b-9qbb9z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/44272123c817937080257d4370290c081fbffaaf/witchcraft/refreshable/refreshable_validating.go#L95)):*
> ```Go
> 	// manually update value after performing subscription. This ensures that, if the current value changed between when
> 	// it was fetched earlier in the function and when the subscription was performed, it is properly captured.
> 	updateValueFn(origRefreshable.Current())
> ```

Would like to do so, but not sure how that can be done in a deterministic manner without modifying the code -- we need a way to be able to change the value between the initial set and the update set, but without adding channels/locks, there's no way to guarantee that an update happens at this time. All the operations here are also very quick/fast, so not sure that even if we hammered at a high rate we could guarantee seeing it.

We could verify manually locally by adding locks/channels (which I have not done), but that wouldn't be part of the code that's checked in.

Do you have any thoughts on how to effectively test for this case?

ashrayjain:
We can have a test implementation of a refreshable (or a mock, but I prefer a test impl) that we pass in, one where as soon as `Current()` is called once, we update the value of the refreshable to be invalid. Then the test can be to validate that the validatingrefreshable we get has a `validateErr` and still has a `Current()` value equal to the original value of our base refreshable.

nmiyake:
Makes sense; that sounds good to me. Would you mind picking this up? I have a few other things I'm looking at right now, so probably won't be able to get around to this for a bit.

ashrayjain:
I'm afraid I won't have time to push this through either anytime soon :(
Maybe @k-simons has a few spare cycles?

bmoylan:
Moved this over to the new pkg/refreshable and added a test: https://github.com/palantir/pkg/pull/209

PR #144, opened 2019-11-23
Feature: add module definition

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/144)


go.mod
```diff
@@ -0,0 +1,25 @@
...
+	github.com/palantir/pkg/safejson v1.0.0
+	github.com/palantir/pkg/safeyaml v1.0.0
+	github.com/palantir/pkg/signals v1.0.0
+	github.com/palantir/pkg/tlsconfig v1.0.0
+	github.com/palantir/witchcraft-go-error v1.2.0
+	github.com/palantir/witchcraft-go-logging v1.5.0
+	github.com/palantir/witchcraft-go-params v1.1.0
+	github.com/palantir/witchcraft-go-tracing v1.2.0
+	github.com/stretchr/testify v1.4.0
+	gopkg.in/fsnotify.v1 v1.4.7
```
tabboud: Should we apply a `replace` directive for this import to `github.com/fsnotify/fsnotify`? We could alternatively import it from `github.com` directly rather than `gopkg.in` here:
https://github.com/palantir/witchcraft-go-server/blob/develop/witchcraft/refreshable/refreshable_file.go#L27

nmiyake: Since this is meant to be a library, would like to avoid replace directives (since they're not inherited). Fine with moving this import over to `fsnotify/fsnotify` as long as we don't have other dependencies of this project that pull in `gopkg.in/fsnotify`. Not sure if one is considered more canonical over another -- for public .com repos, the "gopkg.in" path is imported by 1,201 packages (https://godoc.org/gopkg.in/fsnotify.v1) and and "fsnotify/fsnotify" is imported by 1,352 packages (https://godoc.org/github.com/fsnotify/fsnotify), so it seems to be fairly close. Couldn't see a definitive split internally.

It looks like "fsnotify/fsnotify" may be the more canonical path, so fine with switching with that. However, will note that this is not universally true -- for example, for `gopkg.in/yaml.v2`, the "gopkg.in" path is considered canonical, even though the repo itself is also in github.com.

tabboud: That makes sense. I was more just curious if that would be a better approach, but also didn't find consensus on what the best practice was here, which seems like there really isn't one currently. 
The reason I mentioned this, was because I noticed in other repositories that vendor this I was getting the following error when those repositories have the replace directive, and thought this could be an issue for consumers.

```
go: github.com/fsnotify/fsnotify@v1.4.7 used for two different module paths (github.com/fsnotify/fsnotify and gopkg.in/fsnotify.v1)
```

The downside to changing it to using the `github.com` is that now consumers will need the replace directive to get around the above issues.

tabboud (approved):
Overall lgtm 👍

nmiyake:
Great, thanks!

My plan is to perform an autorelease of all of the current changes once https://github.com/palantir/witchcraft-go-server/pull/146 has been merged so that we have a release that contains all of the work/bug fixes prior to this PR.

Once that release has been cut, I plan to merge this PR and perform another release that consists of only this change.

nmiyake:
@bmoylan let me know if this looks good/this plan sounds OK -- will merge once I get an ACK.

bmoylan (approved):
sounds good!

PR #151, opened 2019-12-11
Add configHealthCheckSource after the init function

k-simons:
## Before this PR

- We added the configHealthCheckSource before the init function. This caused it to get overridden by user health checks

## After this PR

- We add configHealthCheckSource after the init function like the SERVER_STATUS health check. This preserves it. 

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/151)

==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/151)

### DIFF
--- a/changelog/@unreleased/pr-151.v2.yml
+++ b/changelog/@unreleased/pr-151.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+feature:
+  fix: Add configHealthCheckSource after the init function
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/151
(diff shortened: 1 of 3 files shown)


nmiyake (approved):
Content looks good to me! Touched up the comment that was added a bit, but otherwise looks good.

changelog/@unreleased/pr-151.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+feature:
```
nmiyake, line 2: Ah, missed that there's a mismatch here: will open a follow-on PR to fix.

PR #154, opened 2019-12-16
Improvement: add Go metric tags to server.uptime metric

nmiyake:
Add tags that include Go version, Go OS and Go Arch to the
server.uptime metric.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/154)

### DIFF
--- a/changelog/@unreleased/pr-154.v2.yml
+++ b/changelog/@unreleased/pr-154.v2.yml
@@ -1,0 +1,9 @@
+type: improvement
+improvement:
+  description: |-
+    Add Go metric tags to server.uptime metric
+
+    Add tags that include Go version, Go OS and Go Arch to the
+    server.uptime metric.
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/154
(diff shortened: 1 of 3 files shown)


nmiyake:
This is analogous to the approach used on the Java side: see https://github.com/palantir/tritium/pull/516.

This will add new tags to the metric for every witchcraft-go-server service, so please flag if there are any concerns/issues. Cardinality of all of these should be fairly low.

The GOOS and GOARCH values are informational, but may not provide very valuable signal/information (would assume almost everything will be linux amd64) -- if adding tags has a non-negligible cost, then it may make sense to remove these. However, I think the version is valuable information that should be tracked/surfaced.

bmoylan (approved):
Yep, cardinality is not a concern since each process will only have one value at a time

nmiyake:
@mahmoudm can you verify that you're OK with this as well?

mahmoudm:
sgtm

mahmoudm (approved):

tabboud (approved):
👍

PR #155, opened 2020-01-06
Fix: Add default status code for rest error when none is provided.

jamesross3:
## Before this PR
It is possible for the httpStatusCode safe param to have a value of 0 when logged by [this line](https://github.com/palantir/witchcraft-go-server/blob/develop/rest/handlers.go#L105). This is because rest error creation ([here](https://github.com/palantir/witchcraft-go-server/blob/develop/rest/errors.go#L39)) can be called without a `StatusCode()` error parameter. The implementation of `rest.NewError` is such that the HTTP status code is the zero value (0) in this case. We always attach the httpStatusCode safe param in `rest.NewError`.

## After this PR
==COMMIT_MSG==
Add a default value of `500` for the `httpStatusCode` safe parameter. This aligns with the logic we have in the `rest.StatusCodeMapper` method [here](https://github.com/palantir/witchcraft-go-server/blob/develop/rest/handlers.go#L81). Perhaps we want to move this change to be there instead of in `rest.NewError`.
==COMMIT_MSG==

## Possible downsides?
This change only fixes a misleading log line.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)

### DIFF
--- a/changelog/@unreleased/pr-155.v2.yml
+++ b/changelog/@unreleased/pr-155.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Add default status code for rest error when none is provided.
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/155
+
--- a/changelog/@unreleased/pr-155.v2.yml
+++ b/changelog/@unreleased/pr-155.v2.yml
@@ -3,4 +3,3 @@ fix:
   description: Add default status code for rest error when none is provided.
   links:
     - https://github.com/palantir/witchcraft-go-server/pull/155
-
--- a/rest/errors.go
+++ b/rest/errors.go
@@ -37,7 +37,9 @@ func (f errorParamFunc) apply(err *errorMetadata) {
 }
 
 func NewError(err error, params ...ErrorParam) error {
-	e := errorMetadata{}
+	e := errorMetadata{
+		statusCode: StatusCodeMapper(err),
+	}
 	for _, param := range params {
 		param.apply(&e)
 	}
--- a/rest/errors.go
+++ b/rest/errors.go
@@ -16,6 +16,7 @@ package rest
 
 import (
 	werror "github.com/palantir/witchcraft-go-error"
+	"net/http"
 )
 
 const (
@@ -37,7 +38,9 @@ func (f errorParamFunc) apply(err *errorMetadata) {
 }
 
 func NewError(err error, params ...ErrorParam) error {
-	e := errorMetadata{}
+	e := errorMetadata{
+		statusCode: http.StatusInternalServerError,
+	}
 	for _, param := range params {
 		param.apply(&e)
 	}
--- a/rest/errors.go
+++ b/rest/errors.go
@@ -16,7 +16,6 @@ package rest
 
 import (
 	werror "github.com/palantir/witchcraft-go-error"
-	"net/http"
 )
 
 const (
@@ -38,9 +37,7 @@ func (f errorParamFunc) apply(err *errorMetadata) {
 }
 
 func NewError(err error, params ...ErrorParam) error {
-	e := errorMetadata{
-		statusCode: http.StatusInternalServerError,
-	}
+	e := errorMetadata{}
 	for _, param := range params {
 		param.apply(&e)
 	}
--- a/rest/handlers.go
+++ b/rest/handlers.go
@@ -85,10 +85,7 @@ func StatusCodeMapper(err error) int {
 		return http.StatusInternalServerError
 	}
 	statusCodeInt, ok := statusCode.(int)
-	if !ok {
-		return http.StatusInternalServerError
-	}
-	if statusCodeInt == 0 {
+	if !ok || statusCodeInt == 0 {
 		return http.StatusInternalServerError
 	}
 	return statusCodeInt
--- a/rest/handlers.go
+++ b/rest/handlers.go
@@ -88,6 +88,9 @@ func StatusCodeMapper(err error) int {
 	if !ok {
 		return http.StatusInternalServerError
 	}
+	if statusCodeInt == 0 {
+		return http.StatusInternalServerError
+	}
 	return statusCodeInt
 }
 
--- a/rest/handlers.go
+++ b/rest/handlers.go
@@ -88,9 +88,6 @@ func StatusCodeMapper(err error) int {
 	if !ok {
 		return http.StatusInternalServerError
 	}
-	if statusCodeInt == 0 {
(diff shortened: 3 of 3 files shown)


nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CRogers and @jamesross3)

___
*[rest/errors.go, line 19 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw2bQs-r_9E8E2XGa-:-Lxw2bQt5_ukjMNQLMBf:b-n93dwo) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/24b2655e9e0a56efe6b937c7fbd2ad8aaa863aaa/rest/errors.go#L19)):*
> ```Go
> import (
> 	werror "github.com/palantir/witchcraft-go-error"
> 	"net/http"
> ```

Import order is wrong -- need to run `format`

CRogers (commented):
I am not the right person to review this, but don't have perms to unrequest my review, so commenting to clear my list.

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CRogers and @jamesross3)

___
*[rest/errors.go, line 42 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw2ykc-HShwYs-sBwD:-Lxw2ykc-HShwYs-sBwE:b-25kdcw) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/24b2655e9e0a56efe6b937c7fbd2ad8aaa863aaa/rest/errors.go#L42)):*
> ```Go
> func NewError(err error, params ...ErrorParam) error {
> 	e := errorMetadata{
> 		statusCode: http.StatusInternalServerError,
> ```

Would it be more correct to have this use `StatusCodeMapper(err)` as the code?

nmiyake:
@CRogers I've seen this happen on multiple Go OSS repos (someone from the infra team added automatically for PR). We should probably track down what's causing this and modify the settings to reduce noise and allow the PRs to move forward more smoothly.

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jamesross3)

___
*[rest/handlers.go, line 79 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw6F8e3gnLwezO8koD:-Lxw6F8f4GSXI0ymn73b:b-b8zur0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/8ae1cb7b8e295f66a50497282e98b9f1a17cdf12/rest/handlers.go#L79)):*
> ```Go
> }
> 
> // StatusCodeMapper maps a provided error to an HTTP status code. If the provided error contains a status code added
> ```

Minor thing, but let's update comment to state:

"If the provided error contains a non-0 status code added using..."

This clarifies that, even if a status code of 0 is explicitly set using the StatusCode parameter, this function is still expected to return StatusInternalServerError.

jamesross3 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jamesross3 and @nmiyake)

___
*[rest/errors.go, line 42 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw2ykc-HShwYs-sBwD:-Lxw6U0RBHTj_OsXKS0z:b-ckcvmb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/24b2655e9e0a56efe6b937c7fbd2ad8aaa863aaa/rest/errors.go#L42)):*

Would it be more correct to have this use `StatusCodeMapper(err)` as the code?
</blockquote></details>

Yes I think that's better; done.

nmiyake (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jamesross3)

___
*[rest/handlers.go, line 88 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw7OVxCvTrEf2YUeCV:-Lxw7OVxCvTrEf2YUeCW:bnsfasb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/8f7c665acafcc52ae0c8295b7c243944328ef862/rest/handlers.go#L88)):*
> ```Go
> 	}
> 	statusCodeInt, ok := statusCode.(int)
> 	if !ok {
> ```

Nit that I missed the first time: let's combine:

```
if !ok || statusCodeInt == 0 {
   ...
}
```

nmiyake (commented):
Thanks, content of the change looks good to me! Can you add a generated changelog entry to satisfy the check?

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-:-Lxw8LJD2URpiRXBhYZw:b-ezeji0)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jamesross3 and @nmiyake)

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: all files reviewed, 3 unresolved discussions (waiting on @jamesross3)

___
*[changelog/@unreleased/pr-155.v2.yml, line 6 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/155#-Lxw8mGQ9hE-D8nFN944:-Lxw8mGQ9hE-D8nFN945:buz6fyg) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/8cc50a9bcdd287f1e1a1a627e809abab14e3781d/changelog/@unreleased/pr-155.v2.yml#L6)):*
> ```YAML
>   links:
>     - https://github.com/palantir/witchcraft-go-server/pull/155
> ```

Remove extra newline (assuming this was hand-generated/copy-pasted -- if not, please let me know how you generated it so I can fix whatever created the output)

nmiyake (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/155)* status: all files reviewed, 3 unresolved discussions (waiting on @jamesross3)

PR #158, opened 2020-02-14
Bump werror to 1.3.0 and move over Error/Wrap functions 

k-simons:
## Before this PR

- Bump werror to 1.3.0 and move over Error/Wrap functions
- The Error/Wrap functions without a context are deprecated 
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/158)

### DIFF
--- a/app_example/main.go
+++ b/app_example/main.go
@@ -50,7 +50,7 @@ func main() {
 			}
 
 			// register endpoint that uses runtime configuration
-			myNumRefreshable := refreshable.NewInt(info.RuntimeConfig.Map(func(in interface{}) interface{} {
+			myNumRefreshable := refreshable.NewInt(info.RuntimeConfig.Map(ctx, func(in interface{}) interface{} {
 				return in.(AppRuntimeConfig).MyNum
 			}))
 			if err := registerRuntimeNumEndpoint(info.Router, myNumRefreshable); err != nil {
--- a/changelog/@unreleased/pr-158.v2.yml
+++ b/changelog/@unreleased/pr-158.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Bump werror to 1.3.0 and move over Error/Wrap functions
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/158
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/palantir/pkg/safeyaml v1.0.0
 	github.com/palantir/pkg/signals v1.0.0
 	github.com/palantir/pkg/tlsconfig v1.0.0
-	github.com/palantir/witchcraft-go-error v1.2.0
+	github.com/palantir/witchcraft-go-error v1.3.0
 	github.com/palantir/witchcraft-go-logging v1.5.0
 	github.com/palantir/witchcraft-go-params v1.1.0
 	github.com/palantir/witchcraft-go-tracing v1.2.0
--- a/go.sum
+++ b/go.sum
@@ -67,6 +67,8 @@ github.com/palantir/pkg/transform v1.0.0 h1:21MzkUg9fQgIdadTYMM1Z1qrml2MVdpNY5ai
 github.com/palantir/pkg/transform v1.0.0/go.mod h1:YH2PQUzswoDayk4rTvKt6B+NcnUJgZRNr9MEqfAMCo0=
 github.com/palantir/witchcraft-go-error v1.2.0 h1:YFoZ8VC0ZLCGuhqM9iqdflUrTHGQmc3DC4GXGDZkhfY=
 github.com/palantir/witchcraft-go-error v1.2.0/go.mod h1:/cl2dMkuBbnfxDtFiC//8JfvZxmRkYRhgv3bBux9AD0=
+github.com/palantir/witchcraft-go-error v1.3.0 h1:1FUvIv+jS22t81uPI0Q5JEwzBDigClqGJJkLTuE8sZI=
+github.com/palantir/witchcraft-go-error v1.3.0/go.mod h1:/cl2dMkuBbnfxDtFiC//8JfvZxmRkYRhgv3bBux9AD0=
 github.com/palantir/witchcraft-go-logging v1.5.0 h1:LxmZ6XuhitMKmNrUQZ3UBU92q6PKPsawV94FlLVUui4=
 github.com/palantir/witchcraft-go-logging v1.5.0/go.mod h1:x2wqelmEPV2sqOgxnYpx7em44I2nzWuovl7d7cMv+pM=
 github.com/palantir/witchcraft-go-params v1.1.0 h1:siRqQv9TuJ0qY2JK5Svd3/rGQQCWvNnjI2OGAftm8gc=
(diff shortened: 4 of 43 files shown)


iamdanfox:
@k-simons I think i'm probably not best placed to review this, but the policy on this repo has auto-requested me.  Would you mind tuning it to some go-specific folks? https://github.com/palantir/witchcraft-go-server/blob/79c14183e4eb69cacc37db5b137557af92389ef7/.policy.yml#L26

k-simons:
@iamdanfox sure requested @nmiyake

nmiyake (changes requested):
Thanks for the PR!

The major high-level feedback is that, as of right now, I'm pretty opposed to the idea of adding a `context.Context` as a parameter just so that we can make the new error calls. My thought is that the first level/layer that has access to a context should use the new API calls to attach the parameters at that level -- adding a context just for the purpose of making the new call seems wrong (and also doesn't even add value in the scenarios where we're just adding `context.Background` to the call site).

This is also one of the reasons that I think, in the long term, the no-context version of the function shouldn't necessarily be deprecated -- there are valid instances in which a function doesn't have a context, but may want to wrap/return an error (although I'm still on-board with deprecating for now to force the new behavior in the common case).

I would have the above concern/objection even if the changes were just to non-exported functions, but for exported functions this applies doubly -- for a core project like this, we shouldn't be able to make breaking changes without a major version upgrade. Even though new compile errors should be simple to fix (just add a new context), this will still cause compile errors when current clients upgrade their dependencies, which violates the semver agreement. I know we're sometimes a bit loose on that, but with a core library like witchcraft-go-server I think it's more important to be stricter about semver contracts.

> Reviewed 42 of 42 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-:-M0ORbQ8BQAH1UeKD5Da:b-y9hhim)* status: all files reviewed, 7 unresolved discussions (waiting on @k-simons)

___
*[rest/errors.go, line 41 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0ORWoQ8InFV8jkY7zc:-M0ORWoQ8InFV8jkY7zd:boevk6l) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/rest/errors.go#L41)):*
> ```Go
> }
> 
> func NewError(ctx context.Context, err error, params ...ErrorParam) error {
> ```

Major API break
___
*[status/routes/routes.go, line 27 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0OPgC06jgBZ3PhapvN:-M0OPgC06jgBZ3PhapvO:bu5wl1r) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/status/routes/routes.go#L27)):*
> ```Go
> )
> 
> func AddLivenessRoutes(ctx context.Context, resource wresource.Resource, source status.Source) error {
> ```

All of these are API breaks (and these functions have a lot of callers).
___
*[witchcraft/ecv.go, line 46 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0OJWQY-Lnk00Lh3LEM:-M0OJWQY-Lnk00Lh3LEN:bjt2xbv) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/witchcraft/ecv.go#L46)):*
> ```Go
> }
> 
> func ECVKeyFromFile(ctx context.Context, path string) ECVKeyProvider {
> ```

This breaks public API. Additionally, even if it didn't, it would be unnatural/strange to have `ECVKeyFromFile` take a `Context` while `WithECVKeyProvider` doesn't.

Also, these functions perform the actual "work" when the provider function is invoked. If we really do want the context, then the provider interface should be modified to take the context rather than being passed in as part of these calls.

If we really want the context errors, then we should:

1. Define new version of the `ConfigBytesProvider` interface that accepts a `context.Context` as part of the function invocation
2. In places where interface isn't exposed (like `ECVKeyFromFile`), use new interface
3. Where interface is exposed, create wrapper that provides an empty context
4. Update call sites to pass in context

I think that part of the reason why the function doesn't accept a context is that it's used before reading in the install config, so there isn't a context that has things like loggers configured on it. This doesn't totally prevent the usage of them, but that was the original thinking.

For this specific ECV scenario, not sure that it's worth piping this through.
___
*[witchcraft/witchcraft.go, line 338 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0OPUld-_wIcNC6DEsb:-M0OPUle5hP__LE4tEvA:b-3xmq78) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/witchcraft/witchcraft.go#L338)):*
> ```Go
> // WithECVKeyFromFile configures the server to use the ECV key in the file at the specified path as the ECV key for
> // decrypting ECV values in configuration.
> func (s *Server) WithECVKeyFromFile(ctx context.Context, fPath string) *Server {
> ```

See other comment on this
___
*[witchcraft/refreshable/refreshable.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0ORIsO94g2C6LpnC2N:-M0ORIsPEDcXurbfL6UE:boevk6l) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/witchcraft/refreshable/refreshable.go#L30)):*
> ```Go
> 
> 	// Map returns a new Refreshable based on the current one that handles updates based on the current Refreshable.
> 	Map(context.Context, func(interface{}) interface{}) Refreshable
> ```

Major API break
___
*[witchcraft/refreshable/refreshable_default.go, line 44 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0ORMJT1laBKtIVH8wz:-M0ORMJU6q7iN7ezArmG:boevk6l) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/witchcraft/refreshable/refreshable_default.go#L44)):*
> ```Go
> }
> 
> func (d *DefaultRefreshable) Update(ctx context.Context, val interface{}) error {
> ```

Major API break
___
*[witchcraft/wresource/resource.go, line 71 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/158#-M0OQNKWE6MDsqLB1mbo:-M0OQNKWE6MDsqLB1mbp:b-fa062i) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/20da5eb342b25f16fca8f37087f2f20f833d1740/witchcraft/wresource/resource.go#L71)):*
> ```Go
> }
> 
> func (r *resourceImpl) Register(ctx context.Context, endpointName, method, path string, handler http.Handler, params ...wrouter.RouteParam) error {
> ```

This is a major API break.

I would rather not force the addition of contexts in instances like this, and depend on the first caller with a context to add the params from context when they wrap.

k-simons:
Closed for https://github.com/palantir/witchcraft-go-server/pull/161 instead

PR #163, opened 2020-02-27
Add anchored variant of NewHealthyIfNotAllErrorsSource

EdwinRobbins:
## Before this PR

Existing healthyIfNotAllErrorsSource implementation allows single error to dominate starting window and windows after gaps which are larger than the grace period. This change allows for enforcing the grace period on these windows as well.

## After this PR
==COMMIT_MSG==

- Adds healthyIfNotAllErrorsSource option to have grace period respected for initial window and those after gaps
==COMMIT_MSG==

## Possible downsides?

Purely additive, no expected potential negative user impact

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/163)


fawind:
Generally looks good but I'm not super familiar with witchcraft-go. I would suggest a second look from @dansanduleac maybe?

jamesross3 (approved):
Thanks for the documented tests 👍

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/163)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @EdwinRobbins, @fawind, and @splucs)

___
*[status/health/window/error_source.go, line 127 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/163#-M1C6v6x-CsMo9wIr1Lz:-M1C6v6x-CsMo9wIr1M-:b-fs0r7e) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/589c1ba09c187e232c858ad53852abc4e389d9e6/status/health/window/error_source.go#L127)):*
> ```Go
> // NewAnchoredHealthyIfNotAllErrorsSource creates an healthyIfNotAllErrorsSource
> // with supplied checkType, using sliding window of size windowSize, which will
> // anchor healthy for the initial window or after gaps greater than windowSize
> ```

What does it mean to "anchor healthy"? I don't fully understand, so if you could explain it a bit more that would be great!

EdwinRobbins:
@nmiyake I've updated the godoc to make this more clear. In this case means to make the window evaluation forced to be at lease grace period window specified before it can report as unhealthy. It is "anchored" by a healthy check that is inserted at the start of new windows.

status/health/window/error_source_test.go
```diff
@@ -54,9 +57,9 @@ func TestUnhealthyIfAtLeastOneErrorSource(t *testing.T) {
 			name: "unhealthy when there is at least one err",
 			errors: []error{
 				nil,
-				werror.Error("Error #1"),
+				werror.ErrorWithContextParams(context.Background(), "Error #1"),
```
splucs, line 60: `context.Background` has no params, what's the point of adding these?

nmiyake (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/163)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fawind and @splucs)

status/health/window/error_source_test.go
EdwinRobbins: the Error call is deprecated, this is just to remove the deprecated calls. Under the hood this is what Error is doing anyway.

nmiyake: Edwin is correct. We had this discussion on the thread that added the `ErrorWithContextParams` function. Technically, `werror.Error` isn't fully deprecated (it's fine to use in cases where no context is available), but from a practical perspective most callers should prefer the `ErrorWithContextParams` function, so we marked the other function as "deprecated" for now to provide better signal for current developers. Once we get a critical mass of usage of the newer approach, we may un-deprecate.

ashrayjain:
If i have errors (and only errors) at a freq equal to the window size, doesn't this make my health check always be green (incorrectly)?

EdwinRobbins:
> If i have errors (and only errors) at a freq equal to the window size, doesn't this make my health check always be green (incorrectly)?

There is a sampling behavior at work here. You'd only want to use a grace period that is at least 2x your health reporting rate, a la Nyquist sampling. It is possible for a user to configure this health check source in a way that may not alert them, but this is just to provide another option and it's up to users to choose and configure these as appropriate for their service

PR #165, opened 2020-03-02
Enforce full initial elapses window before HealthyIfNotAllErrorsSource reports unhealthy

EdwinRobbins:
## Before this PR

The default behavior for all HealthyIfNotAllErrorsSource should intuitively be to require the first initial window to elapse before it can become unhealthy, regardless if it is anchored or not. Anchoring should only be required if desired in post startup windows 

## After this PR
==COMMIT_MSG==

HealthyIfNotAllErrorsSources require the window period elapse after startup before becoming unhealthy whether anchored or not.
==COMMIT_MSG==

## Possible downsides?

Breaking change in existing behavior for those who count on initial window to become unhealthy on single error, but in truth such uses should likely have been using NewUnhealthyIfAtLeastOneErrorSource actually anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/165)

### DIFF
--- a/status/health/window/error_source.go
+++ b/status/health/window/error_source.go
@@ -58,12 +58,19 @@ func MustNewUnhealthyIfAtLeastOneErrorSource(checkType health.CheckType, windowS
 // with a sliding window of size windowSize and uses the checkType.
 // windowSize must be a positive value, otherwise returns error.
 func NewUnhealthyIfAtLeastOneErrorSource(checkType health.CheckType, windowSize time.Duration) (ErrorHealthCheckSource, error) {
-	source, err := NewHealthyIfNotAllErrorsSource(checkType, windowSize)
-	if err != nil {
-		return nil, err
+	if windowSize <= 0 {
+		return nil, werror.Error("windowSize must be positive", werror.SafeParam("windowSize", windowSize))
 	}
+
+	underlyingSource := &healthyIfNotAllErrorsSource{
+		timeProvider:       newOrdinaryTimeProvider(),
+		useAnchoredWindows: false,
+		windowSize:         windowSize,
+		checkType:          checkType,
+	}
+
 	return &unhealthyIfAtLeastOneErrorSource{
-		source: source,
+		source: underlyingSource,
 	}, nil
 }
 
@@ -114,6 +121,8 @@ func NewHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.
 // MustNewAnchoredHealthyIfNotAllErrorsSource returns the result of calling
 // NewAnchoredHealthyIfNotAllErrorsSource but panics if that call returns an error
 // Should only be used in instances where the inputs are statically defined and known to be valid.
+// Care should be taken in considering health submission rate and window size when using anchored
+// windows. Windows too close to service emission frequency may cause errors to not surface
 func MustNewAnchoredHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.Duration) ErrorHealthCheckSource {
 	source, err := newHealthyIfNotAllErrorsSource(checkType, windowSize, true, newOrdinaryTimeProvider())
 	if err != nil {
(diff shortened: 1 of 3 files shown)


EdwinRobbins:
@ashrayjain, @splucs, and @ryanmcnamara  who have context around these health changes

status/health/window/error_source.go
```diff
@@ -58,12 +58,19 @@ func MustNewUnhealthyIfAtLeastOneErrorSource(checkType health.CheckType, windowS
...
 // windowSize must be a positive value, otherwise returns error.
 func NewUnhealthyIfAtLeastOneErrorSource(checkType health.CheckType, windowSize time.Duration) (ErrorHealthCheckSource, error) {
-	source, err := NewHealthyIfNotAllErrorsSource(checkType, windowSize)
-	if err != nil {
-		return nil, err
+	if windowSize <= 0 {
+		return nil, werror.Error("windowSize must be positive", werror.SafeParam("windowSize", windowSize))
+	}
+
+	underlyingSource := &healthyIfNotAllErrorsSource{
```
ryanmcnamara, line 65: i think you can use the contructor func

ryanmcnamara (approved):

EdwinRobbins (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/165)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @fawind and @ryanmcnamara)

___
*[status/health/window/error_source.go, line 65 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/165#-M1SKRwHCq9IfV09rXya-r3-65:-M1X3wo-35YSgowLavHC:bbgsxy0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/bc69100e92170ef680ea2e2c5446d6c7e67e020c/status/health/window/error_source.go#L65)):*
<details><summary><i>Previously, ryanmcnamara (Ryan ) wrote…</i></summary><blockquote>

i think you can use the contructor func
</blockquote></details>

The UnhealthyIfAtLeastOneErrorSource behaviorally shouldn't grace period its first window as this would make it behave differently than subsequent windows. Before it could leverage the same constructor func but now that the constructor always requires first full window this needs its own logic here.

status/health/window/error_source.go
```diff
@@ -143,9 +154,8 @@ func newHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.
 		checkType:          checkType,
 	}
 
-	if useAnchoredWindows {
-		retVal.lastSuccessTime = retVal.timeProvider.Now()
-	}
+	// Always require the first initial window to elapse before reporting unhealthy
+	retVal.lastSuccessTime = retVal.timeProvider.Now()
```
ashrayjain, line 158: i thought we discussed that we won't do this via using the last success "hack" but instead via a first class that like "startUpTime" or something like that.
not super strongly opposed to this but this is again way too magical imo.

EdwinRobbins (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/165)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @ashrayjain, @fawind, and @ryanmcnamara)

___
*[status/health/window/error_source.go, line 158 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/165#-M1SKRwHCq9IfV09rXya-r3-158:-M1XAANI5L5PxFcxJfGg:b5knvz7) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/bc69100e92170ef680ea2e2c5446d6c7e67e020c/status/health/window/error_source.go#L158)):*
<details><summary><i>Previously, ashrayjain (Ashray Jain) wrote…</i></summary><blockquote>

i thought we discussed that we won't do this via using the last success "hack" but instead via a first class that like "startUpTime" or something like that.
not super strongly opposed to this but this is again way too magical imo.
</blockquote></details>

I took a stab at this but the UnhealthyIfAtLeastOneErrorSource complicates it because it wraps this source and relies on the underlying HealthStatus func. You'd have to refactor its implementation otherwise it also inherits the initial first window grace period which is counter what its name suggests. So this was the most concise change that produces the desired behavior.

EdwinRobbins (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/165)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @ashrayjain, @fawind, and @ryanmcnamara)

___
*[status/health/window/error_source.go, line 158 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/165#-M1SKRwHCq9IfV09rXya-r3-158:-M1XO4Tn8WNy3xcyB8xh:b-ix4n8q) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/bc69100e92170ef680ea2e2c5446d6c7e67e020c/status/health/window/error_source.go#L158)):*
<details><summary><i>Previously, EdwinRobbins wrote…</i></summary><blockquote>

I took a stab at this but the UnhealthyIfAtLeastOneErrorSource complicates it because it wraps this source and relies on the underlying HealthStatus func. You'd have to refactor its implementation otherwise it also inherits the initial first window grace period which is counter what its name suggests. So this was the most concise change that produces the desired behavior. 
</blockquote></details>

And with the comment I think the behavior is clear to readers, so shouldn't seem too magical.

EdwinRobbins:
@nmiyake for help with merge and release

nmiyake (approved):

PR #166, opened 2020-03-06
Change refreshable_file.go to use a ticker instead of fsnotify such that symlinks work

k-simons:
## Before this PR

- The current refreshable_file does not support symlinks correctly. This can be verified by reverting the changes to `refreshable_file.go` and keeping the tests that were also added in this PR
- Our use case revolves quite heavily in symlinks. It is how Kubernetes updates configuration files for a pod
- This changes the behavior (file-checking+checksumming) to occur on a regular (configurable) interval as opposed to using fs-notify which seems unable to process symlink events
- If those are curious, here is a bit more color on how the updates need to be tracked K8s side:
Lets say you mount a config map at var/conf. Every "->" equals a symlink. What actually happens is:

`var/conf/file` -> `var/conf/..data/file`
`var/conf/..data/file`  -> `var/conf/.$TIME_STAMP_OF_CONFIG_UPDATE_1/file`
So double symlink. Then if the configuration updates, what happens is a new directory gets written with:
`var/conf/.$TIME_STAMP_OF_CONFIG_UPDATE_2/file`
And then they re-point
`var/conf/..data/file`  -> `var/conf/.$TIME_STAMP_OF_CONFIG_UPDATE_1/file`
to
`var/conf/..data/file`  -> `var/conf/.$TIME_STAMP_OF_CONFIG_UPDATE_2/file`

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/166)


k-simons:
@nmiyake  for Review

bmoylan:
Instead could we `filepath.EvalSymlinks` on the provided path, so that we watch the underlying file? This would not work if the symlink target itself changed, but I guess we could re-evaluate them every time we get an event.

If we go the ticker route, I think witchcraft-java uses a 1 second interval so might as well match it.

witchcraft/refreshable/refreshable_file.go
```diff
@@ -35,73 +34,58 @@ type fileRefreshable struct {
 }
 
 // NewFileRefreshable returns a new Refreshable whose current value is the bytes of the file at the provided path.
+// The duration configures how often the refreshable check the file
 // Calling this function also starts a goroutine which updates the value of the refreshable whenever the specified file
 // is changed. The goroutine will terminate when the provided context is done or when the returned cancel function is
 // called.
-func NewFileRefreshable(ctx context.Context, filePath string) (r Refreshable, rErr error) {
+func NewFileRefreshable(ctx context.Context, filePath string, duration time.Duration) (Refreshable, error) {
```
nmiyake: Would prefer not to break this public API. Can we leave the arguments the same and use a default refresh duration (per Brad's suggestion, 1s to match Java sounds fine) and then add another `New` function that specifies the duration?

witchcraft/refreshable/refreshable_file_test.go
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) 2018 Palantir Technologies. All rights reserved.
```
nmiyake: nit: since this file was newly created, copyright year should be 2020

nmiyake (commented):
Focusing on high-level until we reach consensus on approach (once we do that, I will do a more thorough review).

I think I'm good with this approach -- only real down-side is that we introduce config reads at this interval, but I think that reading the runtime configurations should generally not be too long, and if the default cadence is what we're suggesting (once a minute or so), I don't anticipate too much load. This does mean that the amount of time it might take for the runtime config to be read may be longer, but that's probably something that we can live with. Removing the `fsnotify` dependency is also a nice side benefit as well.

Any concerns from others with regards to moving from the model of notifications to polling once a minute? If not, then I think proceeding with this approach makes sense to me.

k-simons:
@bmoylan I attempted to get the code working with `filepath.EvalSymlinks` and it is a bit tricky. You need to have the `TestRefreshableCanFollowMovingSymLink` test pass because that is how Kubernetes follows symlinks. 

So you actually have to watch multiple files, the original symlink, and the file requested. If that symlink changes you need to watch that also :/ 

I don't personally think this complexity is worth it

k-simons:
@nmiyake you said 

"Any concerns from others with regards to moving from the model of notifications to polling once a minute? If not, then I think proceeding with this approach makes sense to me."

And also 

"Can we leave the arguments the same and use a default refresh duration (per Brad's suggestion, 1s to match Java sounds fine)"

Updated to 1 second but wanted to clarify you are fine with that

nmiyake:
Yes, I'm good with this. Sounds like you and I are OK with this and Brad seemed to be OK as well, given that he made the 1s suggestion. Think this is a pretty good cross-section, but maybe @ryanmcnamara or @ashrayjain can weigh in as well?

witchcraft/refreshable/refreshable_file.go
```diff
@@ -35,73 +34,58 @@ type fileRefreshable struct {
...
-func (d *fileRefreshable) watchForChanges(ctx context.Context) error {
-	watcher, err := fsnotify.NewWatcher()
+func (d *fileRefreshable) watchForChangesAsync(ctxOuter context.Context, duration time.Duration) {
+	go wapp.RunWithRecoveryLogging(wparams.ContextWithSafeParam(ctxOuter, "filePath", d.filePath), func(ctx context.Context) {
+		d.watchForChanges(ctx, duration)
+	})
+}
+
+func (d *fileRefreshable) watchForChanges(ctx context.Context, duration time.Duration) {
+	gcIntervalTicker := time.NewTicker(duration)
```
bmoylan, line 71: Let's `defer gcIntervalTicker.Stop()`

witchcraft/refreshable/refreshable_file.go
```diff
@@ -35,73 +34,58 @@ type fileRefreshable struct {
...
-	if err := fRefreshable.watchForChanges(ctx); err != nil {
-		return nil, err
-	}
+	fRefreshable.watchForChangesAsync(ctx, duration)
 	return fRefreshable, nil
 }
 
-func (d *fileRefreshable) watchForChanges(ctx context.Context) error {
-	watcher, err := fsnotify.NewWatcher()
+func (d *fileRefreshable) watchForChangesAsync(ctxOuter context.Context, duration time.Duration) {
```
bmoylan: I don't think you should need to use `ctxOuter`, scoping should do the right thing with `ctx`

witchcraft/refreshable/refreshable_file_test.go
```diff
@@ -0,0 +1,155 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package refreshable
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"path"
```
bmoylan: Replace path with filepath so the test works for windows devs

bmoylan (commented):
Yep I am good with this approach if handling the symlinks based on fsnotify is not simple. Thanks!

witchcraft/refreshable/refreshable_file_test.go
k-simons: Done

witchcraft/refreshable/refreshable_file.go
k-simons: Updated! Sorry didn't realize the FileRefreshable was considered public as well

witchcraft/refreshable/refreshable_file.go
k-simons: Interesting! Have not seen this before. What is the benefit? 
Done.

witchcraft/refreshable/refreshable_file.go
k-simons: Just for my own readability. Done

witchcraft/refreshable/refreshable_file_test.go
k-simons: Done

nmiyake (changes requested):
Thanks! Looks good over-all, but some feedback.

> Reviewed 282 of 284 files at r2, 1 of 3 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-:-M1l26hRAkXSaNjy0_AE:bdeswjm)* status: all files reviewed, 13 unresolved discussions (waiting on @bmoylan and @k-simons)

___
*[witchcraft/witchcraft.go, line 329 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1kxtMt2DmU_XeGzTl4:-M1kxtMt2DmU_XeGzTl5:bv7xfny) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/witchcraft.go#L329)):*
> ```Go
> // WithRuntimeConfigFromFileAndDuration is identical to WithRuntimeConfigFromFile except a duration is provided
> // If provided, this duration is the frequency in which the file refreshable is checked
> func (s *Server) WithRuntimeConfigFromFileAndDuration(fpath string, duration time.Duration) *Server {
> ```

Would prefer that we don't expose this, as it's not something that most consumers should ever adjust.

We already have a `WithRuntimeConfigProvider` function, which allows for setting the runtime config provider generically. For the test usage that currently exists, we can call `WithRuntimeConfigProvider` with a `refreshable.NewFileRefreshableWithDuration(ctx, fpath, duration)` created for the test (and you can decomp this as a function in the test if needed)
___
*[witchcraft/refreshable/refreshable_file.go, line 62 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1hsa8--bV_Lay_REjl-r3-62:-M1kzCNgEBdZDUDy56z4:b-6n88xa) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/dfd0ebd020aea9ac06c92022d4a14d030dc4c596/witchcraft/refreshable/refreshable_file.go#L62)):*
<details><summary><i>Previously, k-simons wrote…</i></summary><blockquote>

Interesting! Have not seen this before. What is the benefit? 
Done. 
</blockquote></details>

Mostly a matter of clean-up -- although it doesn't technically stop the other routine, it does ensure that this ticker will no longer fire once called (which is more conceptually correct).
___
*[witchcraft/refreshable/refreshable_file.go, line 37 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1kylJ3EjQSQbBHh44u:-M1kylJ3EjQSQbBHh44v:bpt7wqu) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file.go#L37)):*
> ```Go
> 
> const (
> 	refreshableSyncPeriod = time.Second
> ```

nit: can we name this `defaultRefreshableSyncPeriod`? 
___
*[witchcraft/refreshable/refreshable_file.go, line 41 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1ky_0vCNRqf8jpifBV:-M1ky_0vCNRqf8jpifBW:b-b7iose) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file.go#L41)):*
> ```Go
> 
> // NewFileRefreshable returns a new Refreshable whose current value is the bytes of the file at the provided path.
> // The file is checked every refreshableSyncPeriod
> ```

Comment is now out-of-date. Might make sense to have the `NewFileRefreshableWithDuration` function have the "primary" comment and have the comment for this function say that it calls `NewFileRefreshableWithDuration` with a duration of 1s.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 33 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l0JFU0gi07TbFr5qu:-M1l0JFU0gi07TbFr5qv:b-g1ppai) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L33)):*
> ```Go
> )
> 
> func TestRefreshableChanges(t *testing.T) {
> ```

Test name should be `TestFileRefreshableChanges` since this is specifically testing file `FileRefreshable`.

Can you add brief top-level comments to tests to explain what is being verified? For example, here would be helpful to have something like:

"Verifies that creating a FileRefreshable with a refresh duration of 50ms, updating value and reading the value after 80ms works".

I've found that adding this level of detail that slightly more detailed than the name alone but that doesn't require reading through the actual test code/comments is very helpful when debugging test failures later.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 34 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l-1079XDlSi-QLsTr:-M1l-1079XDlSi-QLsTs:bynvqa) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L34)):*
> ```Go
> 
> func TestRefreshableChanges(t *testing.T) {
> 	tempDir, err := ioutil.TempDir("", "")
> ```

nit: generally prefer following pattern for creating and cleaning up temp directories for tests:

```
tempDir, cleanup, err := dirs.TempDir("", "")
require.NoError(t, err)
defer cleanup()
```
___
*[witchcraft/refreshable/refreshable_file_test.go, line 50 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1CKH8YuUgo4dnj0d:-M1l1CKH8YuUgo4dnj0e:bqmjdh4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L50)):*
> ```Go
> }
> 
> func TestRefreshableSubscribes(t *testing.T) {
> ```

Similar feedback to above: would add `FileRefreshable` to name and top-level summary comment.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 59 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1OKH3rhKS06otLeV:-M1l1OKH3rhKS06otLeW:b-te12em) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L59)):*
> ```Go
> 	str := getStringFromRefreshable(t, r)
> 	assert.Equal(t, str, "renderConf1")
> 	count := 0
> ```

Since `count` is being written and read from different routines, safer to use `atomic.AddInt32` and `atomic.LoadInt32`
___
*[witchcraft/refreshable/refreshable_file_test.go, line 74 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1g2D7fn70akFv34c:-M1l1g2D7fn70akFv34d:bcvhpni) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L74)):*
> ```Go
> }
> 
> func TestRefreshableCanFollowSymLink(t *testing.T) {
> ```

Add "File" to name, and top-level comment.

This is the kind of test where the top-level comment really helps -- test is involved enough that name alone doesn't quite capture it, and although the test is well-commented, it mostly describes *what* is being done, and it's not super easy to zoom out and figure out the intent/expectation that's being tested.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 99 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l2-TIB_jSTzBvQxgU:-M1l2-TIB_jSTzBvQxgV:br3nl2y) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L99)):*
> ```Go
> }
> 
> func TestRefreshableCanFollowMultipleSymLinks(t *testing.T) {
> ```

Same comment for this testing function and the next one.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 166 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l-RNi1Ftoa1WAtNLp:-M1l-RNi1Ftoa1WAtNLq:bu9f3n0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L166)):*
> ```Go
> 
> func writeFileHelper(t *testing.T, path, value string) {
> 	err := ioutil.WriteFile(path, []byte(value), 0777)
> ```

nit: for default file, 0644 should be sufficient permissions

k-simons (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/166)* status: 281 of 284 files reviewed, 8 unresolved discussions (waiting on @bmoylan and @nmiyake)

___
*[witchcraft/witchcraft.go, line 329 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1kxtMt2DmU_XeGzTl4:-M1l8vOg-_4grsB4NWZF:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/witchcraft.go#L329)):*

Would prefer that we don't expose this, as it's not something that most consumers should ever adjust.

We already have a `WithRuntimeConfigProvider` function, which allows for setting the runtime config provider generically. For the test usage that currently exists, we can call `WithRuntimeConfigProvider` with a `refreshable.NewFileRefreshableWithDuration(ctx, fpath, duration)` created for the test (and you can decomp this as a function in the test if needed)
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file.go, line 37 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1kylJ3EjQSQbBHh44u:-M1ldG-mARWQ3LAhS345:bb6un1j) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file.go#L37)):*

nit: can we name this `defaultRefreshableSyncPeriod`? 
</blockquote></details>

Done
___
*[witchcraft/refreshable/refreshable_file.go, line 41 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1ky_0vCNRqf8jpifBV:-M1ldxHHEXnJleVPZzXA:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file.go#L41)):*

Comment is now out-of-date. Might make sense to have the `NewFileRefreshableWithDuration` function have the "primary" comment and have the comment for this function say that it calls `NewFileRefreshableWithDuration` with a duration of 1s.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 33 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l0JFU0gi07TbFr5qu:-M1lcO1K2g7za4_xueXs:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L33)):*

Test name should be `TestFileRefreshableChanges` since this is specifically testing file `FileRefreshable`.

Can you add brief top-level comments to tests to explain what is being verified? For example, here would be helpful to have something like:

"Verifies that creating a FileRefreshable with a refresh duration of 50ms, updating value and reading the value after 80ms works".

I've found that adding this level of detail that slightly more detailed than the name alone but that doesn't require reading through the actual test code/comments is very helpful when debugging test failures later.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 34 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l-1079XDlSi-QLsTr:-M1l9IH28Y8etj7Ychuz:bb6un1j) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L34)):*

nit: generally prefer following pattern for creating and cleaning up temp directories for tests:

```
tempDir, cleanup, err := dirs.TempDir("", "")
require.NoError(t, err)
defer cleanup()
```
</blockquote></details>

Done
___
*[witchcraft/refreshable/refreshable_file_test.go, line 50 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1CKH8YuUgo4dnj0d:-M1lcNGF2VTNot4tihuM:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L50)):*

Similar feedback to above: would add `FileRefreshable` to name and top-level summary comment.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 59 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1OKH3rhKS06otLeV:-M1lcNjkCwCgrQj60zZj:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L59)):*

Since `count` is being written and read from different routines, safer to use `atomic.AddInt32` and `atomic.LoadInt32`
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 74 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l1g2D7fn70akFv34c:-M1ld6bz5BLv8nJ63QOV:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L74)):*

Add "File" to name, and top-level comment.

This is the kind of test where the top-level comment really helps -- test is involved enough that name alone doesn't quite capture it, and although the test is well-commented, it mostly describes *what* is being done, and it's not super easy to zoom out and figure out the intent/expectation that's being tested.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 99 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l2-TIB_jSTzBvQxgU:-M1ld6sX4zXz5UleKP9N:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L99)):*

Same comment for this testing function and the next one.
</blockquote></details>

Done.
___
*[witchcraft/refreshable/refreshable_file_test.go, line 166 at r5](https://reviewable.io/reviews/palantir/witchcraft-go-server/166#-M1l-RNi1Ftoa1WAtNLp:-M1l9LuU-JsKF8V8Oq6t:bb6un1j) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/409f5e9586ea9343f227caf6db3fdd6ab178f1f3/witchcraft/refreshable/refreshable_file_test.go#L166)):*

nit: for default file, 0644 should be sufficient permissions
</blockquote></details>

Done

nmiyake (approved):
> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/166)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan)

nmiyake (approved):

PR #168, opened 2020-03-16
Require first window to elapse for single and multi-key HealthyIfNotAllErrorsSource to become unhealthy

EdwinRobbins:
## Before this PR

This aligns the multi key version of HealthyIfNotAllErrors source with the single error source in requiring the first full window to elapse before it will report unhealthy. This makes the check behave more intuitively as otherwise one initial error can dominate the window and make it report unhealthy immediately, which is behavior more in line with MultiKeyUnhealthyIfAtLeastOneErrorSource

## After this PR
==COMMIT_MSG==

MultiKeyUnhealthyIfAtLeastOneErrorSource requires the first full window to elapse before it will report unhealthy.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/168)


EdwinRobbins:
@splucs who is familiar with these health checks

EdwinRobbins:
@ashrayjain who may also be interested

status/health/window/key_error_source.go
```diff
@@ -149,27 +160,31 @@ func (m *multiKeyHealthyIfNotAllErrorsSource) HealthStatus(ctx context.Context)
...
-	m.pruneOldKeys(m.errorStore, m.lastError)
-	m.pruneOldKeys(m.successStore, nil)
-
-	params := make(map[string]interface{})
-	for _, key := range m.errorStore.List().Keys() {
-		if _, hasSuccess := m.successStore.Get(key); hasSuccess {
-			continue
+	var healthCheckResult health.HealthCheckResult
+	if m.requireFirstFullWindow && m.timeProvider.Now().Sub(m.startTime) < m.windowSize {
+		healthCheckResult = whealth.HealthyHealthCheckResult(m.checkType)
```
splucs: Shouldn't we return REPAIRING in this case?

EdwinRobbins: Yes that's more intuitive this was a holdover from the anchored windows used in the single source version of this. I'll update both so that non-anchored versions of both that require the first full windows will report REPAIRING for only errors in their initial windows.

status/health/window/error_source.go
```diff
@@ -139,23 +135,27 @@ func MustNewAnchoredHealthyIfNotAllErrorsSource(checkType health.CheckType, wind
 // considering health submission rate and window size when using anchored windows.
 // Windows too close to service emission frequency may cause errors to not surface
 func NewAnchoredHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.Duration) (ErrorHealthCheckSource, error) {
-	return newHealthyIfNotAllErrorsSource(checkType, windowSize, true, newOrdinaryTimeProvider())
+	return newHealthyIfNotAllErrorsSource(checkType, windowSize, true, true, newOrdinaryTimeProvider())
 }
 
-func newHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.Duration, useAnchoredWindows bool, timeProvider timeProvider) (ErrorHealthCheckSource, error) {
+func newHealthyIfNotAllErrorsSource(checkType health.CheckType, windowSize time.Duration, useAnchoredWindows bool, requireFirstFullWindow bool, timeProvider timeProvider) (ErrorHealthCheckSource, error) {
```
splucs, line 141: Why make this a configurable boolean if all external constructors set this to true? This should be configurable externally also.

EdwinRobbins: They don't all set it as true; NewUnhealthyIfAtLeastOneErrorSource should be come unhealthy for its initial window but uses a wrapped HealthyIfNotAllErrorsSource for its core logic, so calls this with false.

splucs:
👍

EdwinRobbins:
@nmiyake or @bmoylan can you help with merge and release, please?

bmoylan (approved):

PR #170, opened 2020-03-23
Reduce default trace sample rate from 100% -> 1%

### LINKED ISSUE #169: Reduce default trace rate to 1%
## What happened?

Currently, the default trace rate for the tracer on the main server context is 100%. This is rarely the right tracing rate, and accidentally using it can have a significant fan out effect downstream.

We should reduce the default trace rate to `0.01` (1%), which is a reasonable default for most services. We could do this either as a default in the Install config yaml annotations or by changing the fallback sampler to one with 1% sample rate [here](https://github.com/palantir/witchcraft-go-server/blob/b8e9db114ad73c6d1d490cf8107bf4f9bf5b8f6d/witchcraft/witchcraft.go#L906).

asanderson15:
## Before this PR
Fixes https://github.com/palantir/witchcraft-go-server/issues/169

## After this PR
==COMMIT_MSG==
Reduce default trace sample rate from 100% -> 1%
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/170)

### DIFF
--- a/changelog/@unreleased/pr-170.v2.yml
+++ b/changelog/@unreleased/pr-170.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    Reduce default trace sample rate from 100% -> 1%
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/170
+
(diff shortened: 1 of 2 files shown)


bmoylan (approved):
Will give @nmiyake a few hours to NACK but sgtm

PR #172, opened 2020-04-03
Introduce ReadinessReporter and ReadinessComponent

Neverblade:
## Before this PR
We currently have no support for defining custom readiness like we do for health. Services can have specific criteria for accepting or not accepting traffic and there should be a standard way of implementing that.

## After this PR
==COMMIT_MSG==
Introduce ReadinessReporter and ReadinessComponent, which allows for the defining and aggregation of custom readiness checks. Rename the HealthReporter and HealthComponent files for clarity.
==COMMIT_MSG==

## Possible downsides?
This PR is only additive, so no (unless it's possible to depend on the name of a file).

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172)


status/reporter/readinessreporter.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return worstRespStatus, aggregatedMetadata
+}
+
+func (r *readinessReporter) InitializeReadinessComponent(ctx context.Context, name ComponentName) (Component, error) {
+	var component Component
+	component = &readinessComponent{
+		name: name,
+		// Initialize to not ready, mirroring how health components are initialized to REPAIRING.
+		status: http.StatusInternalServerError,
```
willdeuschle, line 75: initialize to REPAIRING for consistency?

k-simons:
@Neverblade I don't think I follow, why can't you just do:

```
info.Router.WithReadiness(getSource(ctx))
```
where 
`getSource` is defined internally in your repo

status/reporter/readinessreporter.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	if _, exists := r.readinessComponents[name]; exists {
+		return nil, werror.ErrorWithContextParams(ctx, "Readiness component already exists.", werror.SafeParam("name", name))
```
willdeuschle: nit: lowercase, no punc

nmiyake (changes requested):
I've left feedback on the current implementation, but would like to sort out whether what @k-simons flagged is applicable before proceeding.

> Reviewed 6 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-:-M40B7CZ2r-auheokaHi:bu4gspn)* status: 6 of 7 files reviewed, 10 unresolved discussions (waiting on @Neverblade)

___
*[status/reporter/readinessreporter.go, line 36 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M404t4G1mcbcGi2ckBJ:-M404t4G1mcbcGi2ckBK:b-pwkmxb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L36)):*
> ```Go
> 
> type readinessReporter struct {
> 	// In the code below, `mutex` solely exists to protect access to `readinessComponents`.
> ```

Can simplify documentation here -- comment of the form "mutex for readinessComponents" or "protects access to readinessComponents" is fine
___
*[status/reporter/readinessreporter.go, line 51 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M405UFC4WOiscaol4m5:-M405UFC4WOiscaol4m6:b-21i2xl) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L51)):*
> ```Go
> 	defer r.mutex.RUnlock()
> 
> 	// If there are no readiness components, we return ready by default.
> ```

nit: prefer neutral style for comments, so omit "we" here:

"if there are no readiness components, return ready by default"
___
*[status/reporter/readinessreporter.go, line 57 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M405GZp0OsxkMwWRU9b:-M405GZp0OsxkMwWRU9c:b-pkxqb6) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L57)):*
> ```Go
> 		respStatus, metadata := component.Status()
> 		// Response codes within [200, 399] are considered ready.
> 		//   See https://github.palantir.build/deployability/sls-spec/blob/develop/docs/readiness.md#readiness.
> ```

Should be changed to "Refer to readiness section of the SLS specification" (don't use internal URL links)
___
*[status/reporter/readinessreporter.go, line 59 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M405tZm4vyLniyqz7v5:-M405tZm4vyLniyqz7v6:b-pg5mvc) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L59)):*
> ```Go
> 		//   See https://github.palantir.build/deployability/sls-spec/blob/develop/docs/readiness.md#readiness.
> 		if respStatus < 200 || respStatus >= 400 {
> 			worstRespStatus = respStatus
> ```

A bit confused here -- the variable name is "worstRespStatus", but based on implementation it looks like it stores the last non-success response status and returns that. So, if multiple components are reporting failure with different response statuses, it will just take the last one encountered.

That might be the most reasonable behavior, but in that case would do 2 things:
* Document this behavior
* Make the result deterministic for a given set of inputs (right now, it's random since map iteration order is random)
  *  Can either sort the inputs so the "last" element will remain steady, or do something like track the largest/smallest failure status
___
*[status/reporter/readinessreporter.go, line 77 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M407Y1G0vg5vkjGl3AW:-M407Y1G0vg5vkjGl3AX:bbzxmxn) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L77)):*
> ```Go
> 	defer r.mutex.Unlock()
> 	if _, exists := r.readinessComponents[name]; exists {
> 		return nil, werror.ErrorWithContextParams(ctx, "Readiness component already exists.", werror.SafeParam("name", name))
> ```

Unlike logger messages, error messages should not be capitalized and should not have ending punctuation since they may be chained with `:`. So, should be:

```
readiness component already exists
```
___
*[status/reporter/readinessreporter.go, line 81 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M408J-62PcwBHU_Hl6X:-M408J-731-9ceY1dpgd:bwpmk3u) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L81)):*
> ```Go
> 
> 	r.readinessComponents[name] = component
> 	svc1log.FromContext(ctx).Info("Registered new readiness component.", svc1log.SafeParam("readinessComponent", name))
> ```

In the error, the param name is "name", but here it's "readinessComponent" -- should be consistent
___
*[status/reporter/readinessreporter.go, line 95 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M408WI372WBYSc2Jjuy:-M408WI372WBYSc2Jjuz:bijncoq) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L95)):*
> ```Go
> 	r.mutex.Lock()
> 	defer r.mutex.Unlock()
> 	if _, exists := r.readinessComponents[name]; exists {
> ```

nit: it's a bit more idiomatic Go to have the "main logic" at the higher level of indentation. So, invert:

```go
if _, exists := r.readinessComponents[name]; !exists {
    return false
}
... // logic
return true
```
___
*[status/reporter/readinessreporter_test.go, line 99 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M408z8ZBtVuthusnlpA:-M408z8ZBtVuthusnlpB:b8k14kt) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter_test.go#L99)):*
> ```Go
> 				"unready-component": "unready-metadata",
> 			},
> 		},
> ```

Add test that has multiple components with failures with different response codes -- that will test/enforce that the logic that chooses the failure response code is stable (per the CR comment, this should sometimes fail right now since right now it's dependent on map iteration order)

Neverblade:
@k-simons I think you can do that (and in some situations could be the right call), and this change won't preclude us from doing so. But a similar argument could be used to argue against the existing HealthReporter/HealthComponent since users can directly feed in custom HealthCheckSources, so I think it boils down to whether the benefits outweigh the cost of adding more code. I think in this case it does because the reporter provides the following two benefits:
1. Response caching, so the user doesn't have to explicitly design their readiness check for that
2. Aggregation of multiple readiness checks. The behavior is somewhat unintuitive so it'd be nice to have a standard version to use/reference.

Neverblade (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @nmiyake and @willdeuschle)

___
*[status/reporter/readinessreporter.go, line 36 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M404t4G1mcbcGi2ckBJ:-M40NJpq11xAAntsogt5:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L36)):*

Can simplify documentation here -- comment of the form "mutex for readinessComponents" or "protects access to readinessComponents" is fine
</blockquote></details>

Done.
___
*[status/reporter/readinessreporter.go, line 57 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M405GZp0OsxkMwWRU9b:-M40N_EW2AW1br7DjZuD:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L57)):*

Should be changed to "Refer to readiness section of the SLS specification" (don't use internal URL links)
</blockquote></details>

Done.
___
*[status/reporter/readinessreporter.go, line 59 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M405tZm4vyLniyqz7v5:-M40Pdz36NYJUBc9v7_Q:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L59)):*

A bit confused here -- the variable name is "worstRespStatus", but based on implementation it looks like it stores the last non-success response status and returns that. So, if multiple components are reporting failure with different response statuses, it will just take the last one encountered.

That might be the most reasonable behavior, but in that case would do 2 things:
* Document this behavior
* Make the result deterministic for a given set of inputs (right now, it's random since map iteration order is random)
  *  Can either sort the inputs so the "last" element will remain steady, or do something like track the largest/smallest failure status
</blockquote></details>

Done.
___
*[status/reporter/readinessreporter.go, line 71 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M4042Kq-FB_-dDlD1J0-r1-71:-M40PnRTAGsroysFfFQ5:b-rnazkt) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L71)):*
<details><summary><i>Previously, willdeuschle (Will Deuschle) wrote…</i></summary><blockquote>

initialize to REPAIRING for consistency?
</blockquote></details>

Not sure I follow, is there a http response code equivalent to health's REPAIRING?
___
*[status/reporter/readinessreporter.go, line 77 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M407Y1G0vg5vkjGl3AW:-M40SXSG5hZnqhroIfwF:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L77)):*

Unlike logger messages, error messages should not be capitalized and should not have ending punctuation since they may be chained with `:`. So, should be:

```
readiness component already exists
```
</blockquote></details>

Done.
___
*[status/reporter/readinessreporter.go, line 81 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M408J-62PcwBHU_Hl6X:-M40QWra89AmdfWQRa_f:b-dwrbho) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L81)):*

In the error, the param name is "name", but here it's "readinessComponent" -- should be consistent
</blockquote></details>

Done - added it as a logger param to enforce consistency
___
*[status/reporter/readinessreporter_test.go, line 99 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M408z8ZBtVuthusnlpA:-M40SMnwDrcPq0rN4LwE:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter_test.go#L99)):*

Add test that has multiple components with failures with different response codes -- that will test/enforce that the logic that chooses the failure response code is stable (per the CR comment, this should sometimes fail right now since right now it's dependent on map iteration order)
</blockquote></details>

Done.

willdeuschle (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @Neverblade and @nmiyake)

___
*[status/reporter/readinessreporter.go, line 71 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M4042Kq-FB_-dDlD1J0-r1-71:-M40ks7_0DvUgkJkKgJx:blqc60f) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L71)):*
<details><summary><i>Previously, Neverblade (James Lin) wrote…</i></summary><blockquote>

Not sure I follow, is there a http response code equivalent to health's REPAIRING?
</blockquote></details>

520 (as is done for repairing https://github.palantir.build/deployability/health-status-alerter#values), but this may not be idiomatic for readiness

Neverblade:
___
*[status/reporter/readinessreporter.go, line 71 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-M4042Kq-FB_-dDlD1J0-r1-71:-M4164_O3xkXRBOxnUA6:br18qz0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5913729f4dfb2e713e229cccff76b47a30f1513e/status/reporter/readinessreporter.go#L71)):*
<details><summary><i>Previously, willdeuschle (Will Deuschle) wrote…</i></summary><blockquote>

520 (as is done for repairing https://github.palantir.build/deployability/health-status-alerter#values), but this may not be idiomatic for readiness
</blockquote></details>

Based off of the health section of the SLS spec I think that might be out of date or historical; the current expectation for health is that a 200 is returned regardless of the status. For readiness it doesn't look like we make a distinction beyond ready/unready so I don't have a strong opinion on what specific code to use

Neverblade (commented):
@nmiyake Thank you for the review! I've updated the PR based off of your feedback - I also checked in w/ Kevin, who's now on-board w/ the intent of this PR. Let me know if there are other steps to pursue here (i.e. if I should bring in another reviewer)

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172#-:-M54-IW_AJ9YKaRs9ujA:bv1p64j)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @nmiyake and @willdeuschle)

nmiyake (approved):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/172)* status: all files reviewed, 1 unresolved discussion (waiting on @willdeuschle)

nmiyake:
@willdeuschle have your concerns been addressed?

@Neverblade can you add a changelog to the PR?

willdeuschle:
yes I'm a +1

PR #175, opened 2020-04-16
Return repairing for errors inside anchored window

splucs:
## Before this PR
Currently the health check simply swallows all errors for all errors than happen after a long period of inactivity for up to one time window. Instead it should return repairing and expose the error message.
Also there is a bug where if a small (less than a time window) stream of error happens, the health check reports unhealthy for a period of the stream one widow after the start of the stream. 

## After this PR
Now errors that happen immediately after long periods of inactivity are not ignored and cause the health check to report repairing.
The mentioned bug is now fixed: the health check will report repairing for the whole time.
==COMMIT_MSG==
Return repairing for errors inside anchored window
==COMMIT_MSG==

## Possible downsides?
No that I'm aware of.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/175)

PR #176, opened 2020-04-17
Return repairing for errors inside multi key anchored window

splucs:
This PR is analogous to https://github.com/palantir/witchcraft-go-server/pull/175, but for multi key window checks.

## Before this PR
Currently the health check does not support "anchoring", which leads to initial errors after long periods of inactivity to cause the health check to report ERROR. We should report REPAIRING for an initial period after idle time.

## After this PR
==COMMIT_MSG==
Return repairing for errors inside multi key anchored window
==COMMIT_MSG==

## Possible downsides?
Not that I'm aware of.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/176)

### DIFF
--- a/changelog/@unreleased/pr-176.v2.yml
+++ b/changelog/@unreleased/pr-176.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    Return repairing for errors inside multi key anchored window
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/176
+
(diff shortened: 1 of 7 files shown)


nmiyake (approved):
> Reviewed 6 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/176)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #177, opened 2020-04-23
improvement: Add health check source registry

splucs:
## Before this PR
One of the problems with the w-g-s is that after the health check sources are passed on to the server, it is not possible to add new ones or remove existing ones.

## After this PR
Now we have a dynamic registry of health checks that can be passed to the server. If we need to add or remove health checks, we can just register or unregister them, respectively.
I'm doing this because we already have 3 products with dynamic health checks.

==COMMIT_MSG==
Add health check source registry
==COMMIT_MSG==

## Possible downsides?
Nothing that I'm aware of. Some internal services implement a health check registry and it works fine.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/177)


bmoylan:
___
*[status/registry/source.go, line 39 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/177#-MZYBfP28OSY9A0yYarF:-MZYBfP28OSY9A0yYarG:b-aoshr) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4dddc52c7db795dbac31daa97bdf06be19701c4a/status/registry/source.go#L39)):*
> ```Go
> 
> type timedHealthCheckSource struct {
> 	timestamp time.Time
> ```

Hey @splucs sorry this PR has languished so long. Can you explain the usefulness of the time tracking? This feels relatively complicated compared to protecting a slice/map with a lock

PR #178, opened 2020-04-27
Add health check source tree.

jamesross3:
## Before this PR
It is typical for some of our services to implement a health check ensuring connectivity/ability to authenticate with another service while also implementing health checks for more specific interactions with that service. A common pattern in our kubernetes controllers, for example is to have two different checks:
1. Can the controller authenticate with the kubernetes API server?
2. Is the controller successfully processing whatever resource events it is receiving from the API server?

Note here that if (1) fails, (2) necessarily fails, as processing resource events is predicated on being able to authenticate with the API server. If (1) is failing, we do not care about debugging (2). In this case, we really just care about the health check result for (1).

This PR introduces a health check source tree that enables consumers to determine when the tree should traverse to health check sources further down in the tree based on the health check result at the current tree node.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/178)

### DIFF
--- a/changelog/@unreleased/pr-178.v2.yml
+++ b/changelog/@unreleased/pr-178.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    It is typical for some of our services to implement a health check ensuring connectivity/ability to authenticate with another service while also implementing health checks for more specific interactions with that service. This PR introduces a health check source tree that enables consumers to determine when the tree should traverse to health check sources further down in the tree based on the health check result at the current tree node.
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/178
+
--- a/status/health/tree/tree.go
+++ b/status/health/tree/tree.go
@@ -22,6 +22,9 @@ import (
 )
 
 type (
+	// TraverseForHealthStatus accepts the health status result of the current health check source node
+	// and returns whether the current health check source tree node should traverse to child health check
+	// source nodes or not.
 	TraverseForHealthStatus func(status health.HealthStatus) bool
 )
 
@@ -50,8 +53,8 @@ func WithTraverseForHealthStatus(fn TraverseForHealthStatus) CheckSourceTreePara
 }
 
 // NewHealthCheckSourceTree returns a new health check source tree node that uses the result of its own health check
-// to determine if it should invoke the provided child health checks. the default behavior is to only traverse to child
-// checks is if the most severe health state of the current node's health status is health.HealthStateHealthy.
+// to determine if it should invoke the provided child health checks. The default behavior is to only traverse to child
+// checks if the most severe health state of the current node's health status is health.HealthStateHealthy.
 func NewHealthCheckSourceTree(
 	ownHealthCheckSource status.HealthCheckSource,
 	childrenHealthCheckSources []status.HealthCheckSource,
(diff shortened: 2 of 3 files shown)


jamesross3:
cc @bmoylan and @nmiyake

nmiyake (changes requested):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/178)* status: all files reviewed, 4 unresolved discussions (waiting on @jamesross3)

___
*[status/health/tree/tree.go, line 25 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61vUst5hfH-5AZLZra:-M61vUst5hfH-5AZLZrb:b-3hdao) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L25)):*
> ```Go
> 
> type (
> 	TraverseForHealthStatus func(status health.HealthStatus) bool
> ```

What's the contract for this function?
___
*[status/health/tree/tree.go, line 53 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61u6U49ZVIhyrbq2r9:-M61u6U49ZVIhyrbq2rA:b-h0ai7o) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L53)):*
> ```Go
> 
> // NewHealthCheckSourceTree returns a new health check source tree node that uses the result of its own health check
> // to determine if it should invoke the provided child health checks. the default behavior is to only traverse to child
> ```

nit: capitalize "The" after period
___
*[status/health/tree/tree.go, line 54 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61uDBD7xNKYD8Xq2RJ:-M61uDBD7xNKYD8Xq2RK:bxxi1go) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L54)):*
> ```Go
> // NewHealthCheckSourceTree returns a new health check source tree node that uses the result of its own health check
> // to determine if it should invoke the provided child health checks. the default behavior is to only traverse to child
> // checks is if the most severe health state of the current node's health status is health.HealthStateHealthy.
> ```

Sentence doesn't seem correct ("is if" with the prior part)
___
*[status/health/tree/tree.go, line 76 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61w-TgDNVItvycGmqq:-M61w-TgDNVItvycGmqr:blxdza3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L76)):*
> ```Go
> 		return ownHealthStatus
> 	}
> 	healthStatusFromChildSources := n.combinedChildrenHealthCheckSource.HealthStatus(ctx)
> ```

nit: even if the functionality is the same, I think it would. be easier. to follow if you ranged over the child source checks and add that into the ownHealthStatus (rather than ranging over the ownHealthStatus checks and adding it into the child sources one)

jamesross3 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/178)* status: all files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[status/health/tree/tree.go, line 25 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61vUst5hfH-5AZLZra:-M61yZLK8Cr0YgPydpFO:btf32t) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L25)):*

What's the contract for this function?
</blockquote></details>

Added documentation
___
*[status/health/tree/tree.go, line 54 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/178#-M61uDBD7xNKYD8Xq2RJ:-M61ygh--BRbYoQukyR3:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/4fab1d2254536f91040af166b7281cd152834ff5/status/health/tree/tree.go#L54)):*

Sentence doesn't seem correct ("is if" with the prior part)
</blockquote></details>

Done.

nmiyake (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/178)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #179, opened 2020-05-04
Improvement: use logging.sls metric to track logger usage

ch-plattner:
## Before this PR
Witchcraft-go-servers do not currently emit metrics about the SLS logs produced.

As part of the go infra tooling week, we are adding this functionality.

## After this PR
==COMMIT_MSG==
This PR adds a metric-recording svclogger to the WGS.
==COMMIT_MSG==

## Possible downsides?
Given that we will be recording metrics with each log, it may be possible that we see a slight perf decrease in logging. We will monitor for this both in units tests and in production when we wire in all of the metric-recording loggers.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/179)

PR #180, opened 2020-05-05
improvement: add benchmarking tests for server loggers

ch-plattner:
## After this PR
This PR adds benchmark tests for all SLS loggers, in preparation for adding metrics about logging in https://github.com/palantir/witchcraft-go-server/pull/179.

==COMMIT_MSG==
Adds a benchmark test for sls loggers
==COMMIT_MSG==

## Possible downsides?
None that I can see.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180)

### DIFF
--- a/changelog/@unreleased/pr-180.v2.yml
+++ b/changelog/@unreleased/pr-180.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add benchmarking tests for server loggers
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/180
--- a/witchcraft/witchcraft_test.go
+++ b/witchcraft/witchcraft_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"net"
 	"os"
 	"strconv"
@@ -108,84 +109,77 @@ func TestFatalErrorLogging(t *testing.T) {
 	}
 }
 
-// BenchmarkServer_Loggers benchmarks the time for Server loggers to log 100 lines.
+// BenchmarkServer_Loggers benchmarks the time for Server loggers to log a fixed number of lines.
 func BenchmarkServer_Loggers(b *testing.B) {
 	for _, test := range []struct {
-		Name   string
-		InitFn witchcraft.InitFunc
+		Name     string
+		InitBody func(context.Context, int)
 	}{
 		{
 			Name: "svc1log",
-			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
-				for i := 0; i < b.N; i++ {
-					for j := 0; j < 100; j++ {
-						svc1log.FromContext(ctx).Info("info!")
-					}
+			InitBody: func(ctx context.Context, numLinesToLog int) {
+				for j := 0; j < numLinesToLog; j++ {
+					svc1log.FromContext(ctx).Info("info!")
 				}
-				return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
 			},
 		},
 		{
 			Name: "evt2log",
-			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
-				for i := 0; i < b.N; i++ {
-					for j := 0; j < 100; j++ {
-						evt2log.FromContext(ctx).Event("event!")
-					}
+			InitBody: func(ctx context.Context, numLinesToLog int) {
+				for j := 0; j < numLinesToLog; j++ {
+					evt2log.FromContext(ctx).Event("event!")
 				}
-				return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
 			},
 		},
 		{
 			Name: "metric1log",
-			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
-				for i := 0; i < b.N; i++ {
-					for j := 0; j < 100; j++ {
-						metric1log.FromContext(ctx).Metric("metric!", metric1log.MetricTypeKey)
-					}
+			InitBody: func(ctx context.Context, numLinesToLog int) {
+				for j := 0; j < numLinesToLog; j++ {
+					metric1log.FromContext(ctx).Metric("metric!", metric1log.MetricTypeKey)
 				}
-				return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
 			},
 		},
 		{
 			Name: "trc1log",
-			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
-				for i := 0; i < b.N; i++ {
-					for j := 0; j < 100; j++ {
-						trc1log.FromContext(ctx).Log(wtracing.SpanModel{})
-					}
+			InitBody: func(ctx context.Context, numLinesToLog int) {
+				for j := 0; j < numLinesToLog; j++ {
+					trc1log.FromContext(ctx).Log(wtracing.SpanModel{})
 				}
-				return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
 			},
 		},
 		{
 			Name: "audit2log",
-			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
-				for i := 0; i < b.N; i++ {
-					for j := 0; j < 100; j++ {
-						audit2log.FromContext(ctx).Audit("audit!", audit2log.AuditResultSuccess)
-					}
+			InitBody: func(ctx context.Context, numLinesToLog int) {
+				for j := 0; j < numLinesToLog; j++ {
+					audit2log.FromContext(ctx).Audit("audit!", audit2log.AuditResultSuccess)
 				}
-				return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
 			},
 		},
 	} {
-		b.Run(test.Name, func(b *testing.B) {
-			logOutputBuffer := &bytes.Buffer{}
-			err := witchcraft.NewServer().
-				WithInitFunc(test.InitFn).
-				WithInstallConfig(config.Install{UseConsoleLog: true}).
-				WithRuntimeConfig(config.Runtime{}).
-				WithLoggerStdoutWriter(logOutputBuffer).
-				WithECVKeyProvider(witchcraft.ECVKeyNoOp()).
-				WithDisableGoRuntimeMetrics().
-				WithMetricsBlacklist(map[string]struct{}{"server.uptime": {}}).
-				WithSelfSignedCertificate().
-				Start()
+		for _, count := range []int{1, 10, 100} {
+			b.Run(fmt.Sprintf("%s-count=%d", test.Name, count), func(b *testing.B) {
+				numLinesToLog := 100
+				logOutputBuffer := &bytes.Buffer{}
+				err := witchcraft.NewServer().
+					WithInitFunc(func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
+						for i := 0; i < b.N; i++ {
+							test.InitBody(ctx, numLinesToLog)
+						}
+						return nil, werror.ErrorWithContextParams(ctx, "must error to get Start to return!")
+					}).
+					WithInstallConfig(config.Install{UseConsoleLog: true}).
+					WithRuntimeConfig(config.Runtime{}).
+					WithLoggerStdoutWriter(logOutputBuffer).
+					WithECVKeyProvider(witchcraft.ECVKeyNoOp()).
+					WithDisableGoRuntimeMetrics().
+					WithMetricsBlacklist(map[string]struct{}{"server.uptime": {}}).
+					WithSelfSignedCertificate().
+					Start()
 
-			// Requires an error so that `Start` will return
-			require.Error(b, err)
-		})
+				// Requires an error so that `Start` will return
+				require.Error(b, err)
+			})
+		}
 	}
 }
 
--- a/witchcraft/witchcraft_test.go
+++ b/witchcraft/witchcraft_test.go
@@ -156,9 +156,8 @@ func BenchmarkServer_Loggers(b *testing.B) {
 			},
 		},
 	} {
-		for _, count := range []int{1, 10, 100} {
-			b.Run(fmt.Sprintf("%s-count=%d", test.Name, count), func(b *testing.B) {
-				numLinesToLog := 100
+		for _, numLinesToLog := range []int{1, 10, 100} {
+			b.Run(fmt.Sprintf("%s-count=%d", test.Name, numLinesToLog), func(b *testing.B) {
 				logOutputBuffer := &bytes.Buffer{}
 				err := witchcraft.NewServer().
 					WithInitFunc(func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
--- a/witchcraft/witchcraft_test.go
+++ b/witchcraft/witchcraft_test.go
@@ -112,66 +112,56 @@ func TestFatalErrorLogging(t *testing.T) {
 // BenchmarkServer_Loggers benchmarks the time for Server loggers to log a fixed number of lines.
 func BenchmarkServer_Loggers(b *testing.B) {
 	for _, test := range []struct {
-		Name     string
-		InitBody func(context.Context, int)
+		Name        string
+		LoggingBody func(context.Context)
 	}{
 		{
 			Name: "svc1log",
-			InitBody: func(ctx context.Context, numLinesToLog int) {
-				for j := 0; j < numLinesToLog; j++ {
-					svc1log.FromContext(ctx).Info("info!")
-				}
+			LoggingBody: func(ctx context.Context) {
+				svc1log.FromContext(ctx).Info("info!")
 			},
 		},
 		{
 			Name: "evt2log",
-			InitBody: func(ctx context.Context, numLinesToLog int) {
-				for j := 0; j < numLinesToLog; j++ {
-					evt2log.FromContext(ctx).Event("event!")
-				}
+			LoggingBody: func(ctx context.Context) {
+				evt2log.FromContext(ctx).Event("event!")
 			},
 		},
 		{
 			Name: "metric1log",
-			InitBody: func(ctx context.Context, numLinesToLog int) {
-				for j := 0; j < numLinesToLog; j++ {
-					metric1log.FromContext(ctx).Metric("metric!", metric1log.MetricTypeKey)
-				}
+			LoggingBody: func(ctx context.Context) {
+				metric1log.FromContext(ctx).Metric("metric!", metric1log.MetricTypeKey)
 			},
 		},
 		{
 			Name: "trc1log",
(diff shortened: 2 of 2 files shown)


nmiyake (changes requested):
Some feedback, but looks good overall!

What does the actual benchmark output look like? Is it broken out per test case (logger type) or not? Just wanted to check, as that would inform whether this should be a single benchmark or separate ones (I think the `b.Run` will make these separate, but just wanted to verify).

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-:-M6aIpsp1aI92sIOoaEL:bwvtr9r)* status: all files reviewed, 2 unresolved discussions (waiting on @ch-plattner)

___
*[witchcraft/witchcraft_test.go, line 111 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aHw5y5dnkPEdwLEan:-M6aHw5y5dnkPEdwLEao:bxhjsd0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6ba936c3737561969d58f02d7364e50569dc6290/witchcraft/witchcraft_test.go#L111)):*
> ```Go
> }
> 
> // BenchmarkServer_Loggers benchmarks the time for Server loggers to log 100 lines.
> ```

Can we extract 100 out to a constant?
___
*[witchcraft/witchcraft_test.go, line 120 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aIFSl6oCvq-mS1vRr:-M6aIFSl6oCvq-mS1vRs:b-gc2v4z) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6ba936c3737561969d58f02d7364e50569dc6290/witchcraft/witchcraft_test.go#L120)):*
> ```Go
> 			Name: "svc1log",
> 			InitFn: func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
> 				for i := 0; i < b.N; i++ {
> ```

Repetition can be reduced a bit here -- would recommend moving the `InitFn` logic to the body of the test and updating the test struct to just take a `func(ctx context.Context`). Then each of the test cases can just define the logic that's actually separate (`svc1log.FromContext(ctx).Info(..,)` etc.)

ch-plattner (commented):
The benchmarking output looks like this:
```
BenchmarkServer_Loggers/svc1log-numLinesToLog=1-16         	  508044	      2032 ns/op
BenchmarkServer_Loggers/svc1log-numLinesToLog=10-16        	   65530	     20192 ns/op
BenchmarkServer_Loggers/svc1log-numLinesToLog=100-16       	    6422	    209397 ns/op
BenchmarkServer_Loggers/evt2log-numLinesToLog=1-16         	  842160	      1532 ns/op
BenchmarkServer_Loggers/evt2log-numLinesToLog=10-16        	   81696	     15295 ns/op
BenchmarkServer_Loggers/evt2log-numLinesToLog=100-16       	    8635	    150312 ns/op
BenchmarkServer_Loggers/metric1log-numLinesToLog=1-16      	  713745	      1786 ns/op
BenchmarkServer_Loggers/metric1log-numLinesToLog=10-16     	   65976	     20416 ns/op
BenchmarkServer_Loggers/metric1log-numLinesToLog=100-16    	    6849	    181950 ns/op
BenchmarkServer_Loggers/trc1log-numLinesToLog=1-16         	  693789	      1905 ns/op
BenchmarkServer_Loggers/trc1log-numLinesToLog=10-16        	   70176	     24920 ns/op
BenchmarkServer_Loggers/trc1log-numLinesToLog=100-16       	    5924	    185683 ns/op
BenchmarkServer_Loggers/audit2log-numLinesToLog=1-16       	  727717	      1805 ns/op
BenchmarkServer_Loggers/audit2log-numLinesToLog=10-16      	   71988	     17627 ns/op
BenchmarkServer_Loggers/audit2log-numLinesToLog=100-16     	    7414	    178835 ns/op
PASS
```

So its broken down by each `b.Run`.

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-:-M6aZoJX5xpzjvWkhO8R:bktb3zy)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[witchcraft/witchcraft_test.go, line 111 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aHw5y5dnkPEdwLEan:-M6aZy6V0yFKGXLzq2jC:b7id5zs) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6ba936c3737561969d58f02d7364e50569dc6290/witchcraft/witchcraft_test.go#L111)):*

Can we extract 100 out to a constant?
</blockquote></details>

Yep, actually added a loop to try a few values 
___
*[witchcraft/witchcraft_test.go, line 120 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aIFSl6oCvq-mS1vRr:-M6a_-od0qdRi6fedRmJ:b-e55spp) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6ba936c3737561969d58f02d7364e50569dc6290/witchcraft/witchcraft_test.go#L120)):*

Repetition can be reduced a bit here -- would recommend moving the `InitFn` logic to the body of the test and updating the test struct to just take a `func(ctx context.Context`). Then each of the test cases can just define the logic that's actually separate (`svc1log.FromContext(ctx).Info(..,)` etc.)
</blockquote></details>

Great, done!

nmiyake (changes requested):
Great, thanks!

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-:-M6aa2Hr1sXzJ3rXRYeD:b-6mhyx0)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ch-plattner and @nmiyake)

___
*[witchcraft/witchcraft_test.go, line 129 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aao0KAh5eORUNff7R:-M6aao0KAh5eORUNff7S:b5ay4x) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L129)):*
> ```Go
> 			Name: "evt2log",
> 			InitBody: func(ctx context.Context, numLinesToLog int) {
> 				for j := 0; j < numLinesToLog; j++ {
> ```

I think it makes sense to move this loop out into the test body as well given that it's the same for all test cases (and we'd expect it to stay the same for any other cases we add later)
___
*[witchcraft/witchcraft_test.go, line 160 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ab7Pm5Qs_SQ0SOkTc:-M6ab7Pm5Qs_SQ0SOkTd:b-u9uzvr) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L160)):*
> ```Go
> 	} {
> 		for _, numLinesToLog := range []int{1, 10, 100} {
> 			b.Run(fmt.Sprintf("%s-numLinesToLog=%d", test.Name, numLinesToLog), func(b *testing.B) {
> ```

Can we append `-benchmark` to the test name? I know it's redundant (and this isn't a common pattern or idiom), but I think it would make the output clearer since the benchmarks will sometimes append a hyphen and a number as well:

```
BenchmarkServer_Loggers/svc1log-numLinesToLog=1-16  
```

This output was a bit confusing to me at first glance, since my brain parsed it as `1-16`. The rename would resolve the issue:

```
BenchmarkServer_Loggers/svc1log-numLinesToLog=1-benchmark-16  
```

Doesn't have to be exactly this construction, but anything that will make the output more readable would be helpful.
___
*[witchcraft/witchcraft_test.go, line 164 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6adHNT4LFlGc8GAyH6:-M6adHNT4LFlGc8GAyH7:bmvhjl6) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L164)):*
> ```Go
> 				err := witchcraft.NewServer().
> 					WithInitFunc(func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
> 						for i := 0; i < b.N; i++ {
> ```

We should probably call `b.ResetTimer()` before the for loop so that the benchmark only considers the time it takes to execute the function -- otherwise, the time to get to the `InitFunc` is also part of the benchmark, and that could introduce variability that's just noise.

I think that should be fine for now, although this also has me wondering whether we should make the benchmark much more targeted (just instantiate loggers with and without metric logger wrapper and benchmark those operations so that it's fully isolated on just the piece we care about)
___
*[witchcraft/witchcraft_test.go, line 174 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ad3xQBKPRHGjGDm-e:-M6ad3xQBKPRHGjGDm-f:b-yt2d8y) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L174)):*
> ```Go
> 					WithECVKeyProvider(witchcraft.ECVKeyNoOp()).
> 					WithDisableGoRuntimeMetrics().
> 					WithMetricsBlacklist(map[string]struct{}{"server.uptime": {}}).
> ```

Is there a reason that we're blacklisting this metric for this test?

ch-plattner (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @nmiyake)

___
*[witchcraft/witchcraft_test.go, line 129 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6aao0KAh5eORUNff7R:-M6aiSQS316e1FoW0DjS:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L129)):*

I think it makes sense to move this loop out into the test body as well given that it's the same for all test cases (and we'd expect it to stay the same for any other cases we add later)
</blockquote></details>

Done.
___
*[witchcraft/witchcraft_test.go, line 160 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ab7Pm5Qs_SQ0SOkTc:-M6ah0cZDumz4lqrn4L6:bqvque6) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L160)):*

Can we append `-benchmark` to the test name? I know it's redundant (and this isn't a common pattern or idiom), but I think it would make the output clearer since the benchmarks will sometimes append a hyphen and a number as well:

```
BenchmarkServer_Loggers/svc1log-numLinesToLog=1-16  
```

This output was a bit confusing to me at first glance, since my brain parsed it as `1-16`. The rename would resolve the issue:

```
BenchmarkServer_Loggers/svc1log-numLinesToLog=1-benchmark-16  
```

Doesn't have to be exactly this construction, but anything that will make the output more readable would be helpful.
</blockquote></details>

Ah, yeah I'm actually not sure why there is a 16. I'll go ahead and add `-benchmark` to make this more accessible
___
*[witchcraft/witchcraft_test.go, line 164 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6adHNT4LFlGc8GAyH6:-M6aiabI2y7j0DSz7VXb:b-icx3ql) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L164)):*

We should probably call `b.ResetTimer()` before the for loop so that the benchmark only considers the time it takes to execute the function -- otherwise, the time to get to the `InitFunc` is also part of the benchmark, and that could introduce variability that's just noise.

I think that should be fine for now, although this also has me wondering whether we should make the benchmark much more targeted (just instantiate loggers with and without metric logger wrapper and benchmark those operations so that it's fully isolated on just the piece we care about)
</blockquote></details>

I was under the impression that it only benchmarked what's inside the `for i := 0; i < b.N; i++ {}` loop? I would definitely be open to isolating the testing though
___
*[witchcraft/witchcraft_test.go, line 174 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ad3xQBKPRHGjGDm-e:-M6ahnQO6JNA8KdbZp6W:b-eqkd9d) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L174)):*

Is there a reason that we're blacklisting this metric for this test?
</blockquote></details>

Nope! Will delete

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ch-plattner and @nmiyake)

___
*[witchcraft/witchcraft_test.go, line 164 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6adHNT4LFlGc8GAyH6:-M6avjtm9GZzJthU3cqM:b-upnl16) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L164)):*
Based on my reading of https://golang.org/pkg/testing/#hdr-Benchmarks, I believe it's for the whole function:

> The benchmark function must run the target code b.N times. During benchmark execution, b.N is adjusted until the benchmark function lasts long enough to be timed reliably. 

Note how it says it's for the entire function. The next section is where they also state:

> If a benchmark needs some expensive setup before running, the timer may be reset
___
*[witchcraft/witchcraft_test.go, line 154 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ay0hg1esvq-6A1a99:-M6ay0hg1esvq-6A1a9A:br3a8pp) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ec5ca6973b66e9f0c5b6cbfa228fe4794c615913/witchcraft/witchcraft_test.go#L154)):*
> ```Go
> 					WithInitFunc(func(ctx context.Context, info witchcraft.InitInfo) (cleanup func(), rErr error) {
> 						for i := 0; i < b.N; i++ {
> 							for j := 0; j < numLinesToLog; j++ {
> ```

Sorry for the late high-level feedback here, but this just clicked for me after reading through the above benchmarks documentation more closely.

We shouldn't need to iterate over `numLinesToLog` or specify that as a parameter, because that's exactly what the benchmarks framework is designed to do -- it will automatically keep adjusting `N` until it can get a good reading.

There are cases where we want to run multiple benchmark tests, but that's usually if we're modifying the input to the function in some manner over the tests. This can be done to verify that an algorithm scales as expected (linear etc.), but if we're just adjusting the number of runs then that can be handled by the benchmarks framework.
___
*[witchcraft/witchcraft_test.go, line 162 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6anmnaENVZGWR_18Iq:-M6anmnb6BF345itoN29:b-o6wylz) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ec5ca6973b66e9f0c5b6cbfa228fe4794c615913/witchcraft/witchcraft_test.go#L162)):*
> ```Go
> 					WithInstallConfig(config.Install{UseConsoleLog: true}).
> 					WithRuntimeConfig(config.Runtime{}).
> 					WithLoggerStdoutWriter(&bytes.Buffer{}).
> ```

Shouldn't really matter, but if we're not examining the output anyway, can use `ioutil.Discard` so that output is a no-op (otherwise this in-memory buffer will just continue to grow)

ch-plattner (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @nmiyake)

___
*[witchcraft/witchcraft_test.go, line 164 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6adHNT4LFlGc8GAyH6:-M6azlyB8QGXFonneCYp:b-dnmfk5) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/7dc6e137808a7d82231fdf4dd3b5e2c52ed1bc98/witchcraft/witchcraft_test.go#L164)):*

Based on my reading of https://golang.org/pkg/testing/#hdr-Benchmarks, I believe it's for the whole function:

> The benchmark function must run the target code b.N times. During benchmark execution, b.N is adjusted until the benchmark function lasts long enough to be timed reliably. 

Note how it says it's for the entire function. The next section is where they also state:

> If a benchmark needs some expensive setup before running, the timer may be reset
</blockquote></details>

Ah, thanks for pointing that out. I've added calls to `ResetTimer` and `StopTimer` around the loop in question.
___
*[witchcraft/witchcraft_test.go, line 154 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6ay0hg1esvq-6A1a99:-M6azy7c9rsNfUXB2hA2:bpad9e1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ec5ca6973b66e9f0c5b6cbfa228fe4794c615913/witchcraft/witchcraft_test.go#L154)):*

Sorry for the late high-level feedback here, but this just clicked for me after reading through the above benchmarks documentation more closely.

We shouldn't need to iterate over `numLinesToLog` or specify that as a parameter, because that's exactly what the benchmarks framework is designed to do -- it will automatically keep adjusting `N` until it can get a good reading.

There are cases where we want to run multiple benchmark tests, but that's usually if we're modifying the input to the function in some manner over the tests. This can be done to verify that an algorithm scales as expected (linear etc.), but if we're just adjusting the number of runs then that can be handled by the benchmarks framework.
</blockquote></details>

I see. Yep, since we just want the time to run one log output call, it makes sense to just rely on the built in `N`. Removed all reference to that!
___
*[witchcraft/witchcraft_test.go, line 162 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-M6anmnaENVZGWR_18Iq:-M6azQANBkxTinYGCABV:bclirs) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/ec5ca6973b66e9f0c5b6cbfa228fe4794c615913/witchcraft/witchcraft_test.go#L162)):*

Shouldn't really matter, but if we're not examining the output anyway, can use `ioutil.Discard` so that output is a no-op (otherwise this in-memory buffer will just continue to grow)
</blockquote></details>

That's a good idea. The actual log output isn't helpful

nmiyake (approved):
Great, thanks for iterating on this -- looks good!

> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180#-:-M6bAjGN4PzZ4OO7in5j:bpjjzpw)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nmiyake)

nmiyake (approved):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/180)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #187, opened 2020-05-12
Return conjure NotFound error for 404s

### LINKED ISSUE #186: 404s do not produce request logs
When a request path does not match a registered route, RouteHandlerMiddlewares do not run. This means that request logs are not emitted for the request and it is not included in server.response metrics.

As part of fixing this, we should also return a conjure NotFound error body rather than using the default text response provided by the upstream router implementations. I've opened a draft change at #187 for this.

bmoylan:
Fixes #186 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187)

### DIFF
--- a/wrouter/router_root.go
+++ b/wrouter/router_root.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"net/http"
 	"sort"
+
+	"github.com/palantir/pkg/metrics"
 )
 
 type RootRouter interface {
@@ -27,6 +29,8 @@ type RootRouter interface {
 	AddRequestHandlerMiddleware(handlers ...RequestHandlerMiddleware)
 	AddRouteHandlerMiddleware(handlers ...RouteHandlerMiddleware)
 
+	// RegisterNotFoundHandler registers a handler to produce 404 responses.
+	// It should be called after all middlewares are added to the router.
 	RegisterNotFoundHandler(handler http.Handler)
 }
 
@@ -65,6 +69,7 @@ func New(impl RouterImpl, params ...RootRouterParam) RootRouter {
 		p.configure(r)
 	}
 	r.updateCachedHandler()
+
 	return r
 }
 
@@ -205,7 +210,25 @@ func (r *rootRouter) AddRouteHandlerMiddleware(handlers ...RouteHandlerMiddlewar
 }
 
 func (r *rootRouter) RegisterNotFoundHandler(handler http.Handler) {
-	r.impl.RegisterNotFoundHandler(handler)
+	wrappedHandlerFn := createRouteRequestHandler(func(rw http.ResponseWriter, r *http.Request, reqVals RequestVals) {
+		handler.ServeHTTP(rw, r)
+	}, r.routeHandlers)
+
+	r.impl.RegisterNotFoundHandler(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		wrappedHandlerFn(rw, r, RequestVals{
+			Spec: RouteSpec{
+				Method:       r.Method,
+				PathTemplate: "/*",
+			},
+			PathParamVals: map[string]string{},
+			ParamPerms: &requestParamPermsImpl{
+				pathParamPerms:   newParamPerms(nil, nil),
+				queryParamPerms:  newParamPerms(nil, nil),
+				headerParamPerms: newParamPerms(nil, nil),
+			},
+			MetricTags: metrics.Tags{},
+		})
+	}))
 }
 
 type requestHandlerWithNext struct {
--- a/wrouter/router_root.go
+++ b/wrouter/router_root.go
@@ -26,6 +26,8 @@ type RootRouter interface {
 
 	AddRequestHandlerMiddleware(handlers ...RequestHandlerMiddleware)
 	AddRouteHandlerMiddleware(handlers ...RouteHandlerMiddleware)
+
+	RegisterNotFoundHandler(handler http.Handler)
 }
 
 type rootRouter struct {
@@ -200,6 +202,10 @@ func (r *rootRouter) AddRequestHandlerMiddleware(handlers ...RequestHandlerMiddl
 
 func (r *rootRouter) AddRouteHandlerMiddleware(handlers ...RouteHandlerMiddleware) {
 	r.routeHandlers = append(r.routeHandlers, handlers...)
+}
+
+func (r *rootRouter) RegisterNotFoundHandler(handler http.Handler) {
+	r.impl.RegisterNotFoundHandler(handler)
 }
 
 type requestHandlerWithNext struct {
--- a/wrouter/routerimpl.go
+++ b/wrouter/routerimpl.go
@@ -44,5 +44,7 @@ type RouterImpl interface {
 	// in the "pathVarNames" slice.
 	PathParams(req *http.Request, pathVarNames []string) map[string]string
 
+	// RegisterNotFoundHandler registers a handler which generates Not Found (404) responses.
+	// If not provided, the implementation's default behavior (typically http.Error) will be used.
 	RegisterNotFoundHandler(handler http.Handler)
 }
--- a/wrouter/routerimpl.go
+++ b/wrouter/routerimpl.go
@@ -43,4 +43,6 @@ type RouterImpl interface {
 	// the number of parameters in the path. The variable names of the parameters in the path are provided in order
 	// in the "pathVarNames" slice.
 	PathParams(req *http.Request, pathVarNames []string) map[string]string
+
+	RegisterNotFoundHandler(handler http.Handler)
 }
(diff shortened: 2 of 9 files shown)


wrouter/routerimpl.go
```diff
@@ -43,4 +43,6 @@ type RouterImpl interface {
 	// the number of parameters in the path. The variable names of the parameters in the path are provided in order
 	// in the "pathVarNames" slice.
 	PathParams(req *http.Request, pathVarNames []string) map[string]string
+
+	RegisterNotFoundHandler(handler http.Handler)
```
bmoylan, line 49: This is technically a break in the case that someone implemented a third router type, but that's pretty unlikely right?

wrouter/router_root.go
```diff
@@ -202,6 +209,28 @@ func (r *rootRouter) AddRouteHandlerMiddleware(handlers ...RouteHandlerMiddlewar
...
+func (r *rootRouter) RegisterNotFoundHandler(handler http.Handler) {
+	wrappedHandlerFn := createRouteRequestHandler(func(rw http.ResponseWriter, r *http.Request, reqVals RequestVals) {
+		handler.ServeHTTP(rw, r)
+	}, r.routeHandlers)
+
+	r.impl.RegisterNotFoundHandler(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		wrappedHandlerFn(rw, r, RequestVals{
+			Spec: RouteSpec{
+				Method:       r.Method,
+				PathTemplate: "/*",
```
bmoylan, line 221: I tested this on a java service, and they are able to inject the subrouter's context path if one matches. E.g. `/foo/api/bad/path` would generate `/foo/api/*`. This is difficult for us because the subrouters are flattened when provided to the router impl, so I think this is fine.

nmiyake (changes requested):
Thanks for the PR!

Core implementation looks good -- some minor feedback.

> Reviewed 6 of 8 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-:-M7DnwOR9ihjR5jBk_Cs:b6temy7)* status: all files reviewed, 3 unresolved discussions (waiting on @bmoylan)

___
*[integration/server_test.go, line 361 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M7DmEYz8Ki_AL55bq_9:-M7DmEZ-COjPWiN-rjNG:b-t1gw1) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/integration/server_test.go#L361)):*
> ```Go
> 
> 	t.Run("request log", func(t *testing.T) {
> 		//t.Skip("404s do not correctly produce request logs: https://github.com/palantir/witchcraft-go-server/issues/186")
> ```

remove commented-out `t.Skip`
___
*[wrouter/router_root.go, line 221 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M78jmUfNJPhtmXnHjgQ-r2-221:-M7DmtwXDAVA7EqYLYWr:b-trk1n4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/wrouter/router_root.go#L221)):*

I tested this on a java service, and they are able to inject the subrouter's context path if one matches. E.g. `/foo/api/bad/path` would generate `/foo/api/*`. This is difficult for us because the subrouters are flattened when provided to the router impl, so I think this is fine.
</blockquote></details>

Thanks for the investigation -- agree this is fine for now. If we find that this is a major issue, we can look into what it would take to support this later (but as you mentioned, think it would be hard for us to do generically given that we consider the router layer pluggable, and the internal mechanisms they use can vary).
___
*[wrouter/routerimpl.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M78jmUfNJPhtmXnHjgR-r2-47:-M7DnFEhEZLBvK5fJS--:bwkepp2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/wrouter/routerimpl.go#L47)):*

This is technically a break in the case that someone implemented a third router type, but that's pretty unlikely right?
</blockquote></details>

Can you add a comment to document the behavior like for the other functions?

Yes, although it's technically not the best API practice, I think that adding to the interface is fine. If we wanted to be really stringent about it, we could make this a separate interface for now and then do a type assertion for whether or not the router supports a "NotFoundHandler" to determine whether or not we support it and then only add this function to this interface when we release a new major version, but given the realities of how this library is used I don't think the extra overhead is worth it.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[integration/server_test.go, line 361 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M7DmEYz8Ki_AL55bq_9:-M7DoLhKDEB8y1umS0xb:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/integration/server_test.go#L361)):*

remove commented-out `t.Skip`
</blockquote></details>

Done.
___
*[wrouter/routerimpl.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M78jmUfNJPhtmXnHjgR-r2-47:-M7Dp8euAo4lQQhLQmx7:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/wrouter/routerimpl.go#L47)):*

Can you add a comment to document the behavior like for the other functions?

Yes, although it's technically not the best API practice, I think that adding to the interface is fine. If we wanted to be really stringent about it, we could make this a separate interface for now and then do a type assertion for whether or not the router supports a "NotFoundHandler" to determine whether or not we support it and then only add this function to this interface when we release a new major version, but given the realities of how this library is used I don't think the extra overhead is worth it.
</blockquote></details>

Done.

nmiyake (changes requested):
Request for minor modification of added documentation

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-:-M7Dt-gZ5j_5i_YclLuG:b7x8sxp)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[wrouter/routerimpl.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M78jmUfNJPhtmXnHjgR-r2-47:-M7DqiLYC6B5ZQOavBYm:bs50ebu) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/wrouter/routerimpl.go#L47)):*

Done.
</blockquote></details>

Can we make the documentation a bit more precise? Doesn't have to be exactly this, but something along the lines of:

```
RegisterNotFoundHandler registers a handler that is used to handle any requests that do not match any registered routes on the router. If not provided, the implementation's default behavior is used (typically returns an http.Error with a 404 response code).
```

Basically, want to highlight that the contract is that the handler is invoked for non-registered routes -- although returning a 404 is likely the most common/correct behavior, it's not technically part of the contract (someone could register a NotFoundHandler that returns a 200 and echoes the provided URL, for example).

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @nmiyake)

___
*[wrouter/routerimpl.go, line 47 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-M78jmUfNJPhtmXnHjgR-r2-47:-M7Dtb7j0efC6TLGwoBQ:b-ik8b4) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/6cc72c4d34f9d3731e6c165538d273303fa0f5b9/wrouter/routerimpl.go#L47)):*

Can we make the documentation a bit more precise? Doesn't have to be exactly this, but something along the lines of:

```
RegisterNotFoundHandler registers a handler that is used to handle any requests that do not match any registered routes on the router. If not provided, the implementation's default behavior is used (typically returns an http.Error with a 404 response code).
```

Basically, want to highlight that the contract is that the handler is invoked for non-registered routes -- although returning a 404 is likely the most common/correct behavior, it's not technically part of the contract (someone could register a NotFoundHandler that returns a 200 and echoes the provided URL, for example).
</blockquote></details>

Sure thing, done.

nmiyake (approved):
Great, thanks for implementing this fix!

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/187#-:-M7DuAfB2EgZ_gemuYdu:b2ktmqn)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #192, opened 2020-06-11
fix: do not pass nil registry to loggers

ch-plattner:
## Before this PR
@mruck flagged that errors in the server `Start` function that occur before the metrics registry is initialized cause the instrumented svclogger to panic. This makes debugging install config issues quite difficult.

This happens when `initLoggers` is called in the following two situations:
* https://github.com/palantir/witchcraft-go-server/blob/develop/witchcraft/witchcraft.go#L522
* https://github.com/palantir/witchcraft-go-server/blob/develop/witchcraft/witchcraft.go#L532

## After this PR
==COMMIT_MSG==
Do not instrument loggers if metrics registry is nil
==COMMIT_MSG==

## Possible downsides?
None

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/192)


k-simons:
@ch-plattner Would it be possible to handle this instead by never passing in a nil Registry? I believe there are 2 places with `witchcraft.go` in which this happens?

Not sure if that is cleaner, but it would be nicer to not have to worry about nil registries if possible, defer to you

ch-plattner:
> @ch-plattner Would it be possible to handle this instead by never passing in a nil Registry? I believe there are 2 places with `witchcraft.go` in which this happens?
> 
> Not sure if that is cleaner, but it would be nicer to not have to worry about nil registries if possible, defer to you

Sure, let me give that a try! I think you're right that makes the API for `initLoggers` more crisp

ch-plattner:
@k-simons good suggestion :D

nmiyake (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/192)* status: all files reviewed, 1 unresolved discussion (waiting on @ch-plattner)

___
*[changelog/@unreleased/pr-192.v2.yml, line 3 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/192#-M9anCmbEPYs6bmz6qd_:-M9anCmcAQDeu4sP20gu:b32s1wp) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d712c17c8f61afb75f2fe7095d507e136c909e80/changelog/@unreleased/pr-192.v2.yml#L3)):*
> ```YAML
> type: fix
> fix:
>   description: Do not instrument loggers if metrics registry is nil
> ```

Changelog should be updated  to reflect the new approach.

ch-plattner (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/192#-:-M9cr1yc6_7vd2G3yGYZ:bnfp4nl)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-192.v2.yml, line 3 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/192#-M9anCmbEPYs6bmz6qd_:-M9cr1ZV1mP4KRT6ubDB:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/d712c17c8f61afb75f2fe7095d507e136c909e80/changelog/@unreleased/pr-192.v2.yml#L3)):*

Changelog should be updated  to reflect the new approach.
</blockquote></details>

Done.

nmiyake (approved):

PR #80, opened 2019-05-29
Default to console logging even when running in non-Docker containers

ashrayjain:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/80)

PR #205, opened 2020-08-27
Improvement: Expand the logToStdoutBasedOnEnv to take into account environment variables

k-simons:
## Before this PR

- logToStdoutBasedOnEnv would look at a file on disk called `"/.dockerenv"` to determine it if was in a container
- This will not work for container runtimes other than docker
- Add an additional configuration point to opt into this based on an env var

## After this PR

- If the environment variable called `CONTAINER` is present and the logger is not explicitly configured, logToStdoutBasedOnEnv will now return true
==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/205)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ traffic itself is still encrypted.
 `witchcraft-server` is configured with service, event, metric, request and trace loggers from the 
 `witchcraft-go-logging` project and emits structured JSON logs using [`zap`](https://github.com/uber-go/zap) as the
 logger implementation. The default behavior emits logs to the `var/log` directory (`var/log/service.log`, 
-`var/log/request.log`, etc.) unless the server is run in a Docker container, in which case the logs are always emitted 
+`var/log/request.log`, etc.) unless the server is run in a Docker container or a container that has an environment variable called `$CONTAINER` set, in which case the logs are always emitted
 to `stdout`. The `use-console-log` property in the install configuration can also be set to "true" to always output logs 
 to `stdout`. The runtime configuration supports configuring the log output level for service logs.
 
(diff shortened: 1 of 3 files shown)


bmoylan:
Going to go with this in favor of #80. This has been requested multiple times over the years, and is consistent with the Java behavior that has been stable for a while. We don't currently have the infra to automatically set this whenever we build a docker container, but some internal teams are looking at tackling that in T3.

Overall I think this is a relatively low-risk change (additive behavior) but I don't want to continue to accumulate tech debt by adding more logic to this check. So I hope we stick with the env var until we can make this the default.

Will give @tabboud a few hours to chime in before merging.

README.md
```diff
@@ -88,7 +88,7 @@ traffic itself is still encrypted.
 `witchcraft-server` is configured with service, event, metric, request and trace loggers from the 
 `witchcraft-go-logging` project and emits structured JSON logs using [`zap`](https://github.com/uber-go/zap) as the
 logger implementation. The default behavior emits logs to the `var/log` directory (`var/log/service.log`, 
-`var/log/request.log`, etc.) unless the server is run in a Docker container, in which case the logs are always emitted 
+`var/log/request.log`, etc.) unless the server is run in a Docker container or a container that has an environment variable called `$CONTAINER` set, in which case the logs are always emitted
```
tabboud: nit: This would still be true if the `CONTAINER` env var is set on a non-container. So how about rewording slightly?
```
... run in a Docker container or has the environment variable `$CONTAINER` set, ...
```

tabboud (commented):
This looks sane to me.
@bmoylan thanks for tagging me, as I am going to begin looking at the generation side in T3, so this will be sure to come up.

k-simons:
@tabboud done

tabboud (approved):

bmoylan (approved):

PR #210, opened 2020-10-15
Improvement: add timed keyed health source

jhowarth-pal:
## Before this PR
There is no easy way to have a health check that will only remain unhealthy for X amount of time or unless a new non-nil error is submitted to renew the timer. The key can also be cleared to healthy if a nil error is submitted.

## After this PR

MultiKeyHealthyIfNoRecentErrorsSource fulfills the above. More information on use case here: https://palantir.quip.com/UMGBAi1IhwRH
==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?

Adds another health source that could confuse people when trying to pick one

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/210)

### DIFF
--- a/status/health/window/key_error_source.go
+++ b/status/health/window/key_error_source.go
@@ -172,8 +172,6 @@ func newMultiKeyHealthyIfNoRecentErrorsSource(checkType health.CheckType, messag
 	}, nil
 }
 
-///////////////// Remove Before Review Joe ////////////////////
-
 // multiKeyHealthyIfNotAllErrorsSource is a HealthCheckSource that keeps the latest errors
 // for multiple keys submitted within the last windowSize time frame.
 // It returns unhealthy if there is at least one key with only non-nil errors within the last windowSize time frame.
--- a/status/health/window/key_error_source.go
+++ b/status/health/window/key_error_source.go
@@ -138,7 +138,7 @@ func (m *multiKeyHealthyIfNoRecentErrorsSource) Submit(key string, err error) {
 	m.sourceMutex.Lock()
 	defer m.sourceMutex.Unlock()
 
-	pruneOldKeys(m.errorStore, m.windowSize, m.timeProvider)
+	m.errorStore.PruneOldKeys(m.windowSize, m.timeProvider)
 	m.errorStore.Put(key, err)
 }
 
@@ -147,7 +147,7 @@ func (m *multiKeyHealthyIfNoRecentErrorsSource) HealthStatus(ctx context.Context
 	m.sourceMutex.Lock()
 	defer m.sourceMutex.Unlock()
 
-	pruneOldKeys(m.errorStore, m.windowSize, m.timeProvider)
+	m.errorStore.PruneOldKeys(m.windowSize, m.timeProvider)
 
 	var healthCheckResult health.HealthCheckResult
 	params := make(map[string]interface{})
@@ -266,9 +266,9 @@ func (m *multiKeyHealthyIfNotAllErrorsSource) Submit(key string, err error) {
 	m.sourceMutex.Lock()
 	defer m.sourceMutex.Unlock()
 
-	pruneOldKeys(m.errorStore, m.windowSize, m.timeProvider)
-	pruneOldKeys(m.successStore, m.windowSize, m.timeProvider)
-	pruneOldKeys(m.gapEndTimeStore, m.repairingGracePeriod+m.windowSize, m.timeProvider)
+	m.errorStore.PruneOldKeys(m.windowSize, m.timeProvider)
+	m.successStore.PruneOldKeys(m.windowSize, m.timeProvider)
+	m.gapEndTimeStore.PruneOldKeys(m.repairingGracePeriod+m.windowSize, m.timeProvider)
 
 	_, hasError := m.errorStore.Get(key)
 	_, hasSuccess := m.successStore.Get(key)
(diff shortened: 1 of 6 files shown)


status/health/window/key_error_source.go
```diff
@@ -243,8 +333,8 @@ func (m *multiKeyHealthyIfNotAllErrorsSource) HealthStatus(ctx context.Context)
 	}
 }
 
-func (m *multiKeyHealthyIfNotAllErrorsSource) pruneOldKeys(store TimedKeyStore, maxAge time.Duration) {
-	curTime := m.timeProvider.Now()
+func pruneOldKeys(store TimedKeyStore, maxAge time.Duration, timeProvider TimeProvider) {
```
jhowarth-pal: Made this a function so it can be reused between the 2, I don't feel strongly though

bmoylan:
Hey @splucs would you be willing to do a first review of this? I think you have thought more about these healthcheck semantics than most

status/health/window/key_error_source.go
splucs: Personally I feel this should be a method of the `TimedKeyStore`.

status/health/window/key_error_source_test.go
```diff
@@ -544,3 +544,125 @@ func TestAnchoredMultiKeyHealthyIfNotAllErrorsSource_GapThenRepairingThenError(t
...
+			},
+			timeProvider: NewOrdinaryTimeProvider(),
+		},
+		{
+			name: "healthy when outside window",
+			keyErrorPairs: []keyErrorPair{
+				{key: "1", err: werror.Error("Error #1 for key 1")},
+			},
+			expectedCheck: whealth.HealthyHealthCheckResult(testCheckType),
+			timeProvider: &sliceTimeProvider{[]time.Time{
```
splucs: This is very fragile and hard to understand. We need to know where in the code `Now` is being called and it will break if another `Now` call is added or removed.
This seems to be the only test that is time sensitive, it can be done by adding a single `RestlessSleep` between the submission of the errors and the health check evaluation.

status/health/window/key_error_source_test.go
```diff
@@ -544,3 +544,125 @@ func TestAnchoredMultiKeyHealthyIfNotAllErrorsSource_GapThenRepairingThenError(t
...
+		name          string
+		keyErrorPairs []keyErrorPair
+		expectedCheck health.HealthCheckResult
+		timeProvider  TimeProvider
+	}{
+		{
+			name:          "healthy when there are no items",
+			keyErrorPairs: nil,
+			expectedCheck: whealth.HealthyHealthCheckResult(testCheckType),
+			timeProvider:  NewOrdinaryTimeProvider(),
```
splucs: Don't use the real time provider for tests. Use the offset one.

status/health/window/key_error_source.go
```diff
@@ -82,6 +82,96 @@ func (m *multiKeyUnhealthyIfAtLeastOneErrorSource) HealthStatus(ctx context.Cont
...
+			m.checkType: healthCheckResult,
+		},
+	}
+}
+
+var _ status.HealthCheckSource = &multiKeyHealthyIfNoRecentErrorsSource{}
+
+// MustNewMultiKeyHealthyIfNoRecentErrorsSource returns the result of calling NewMultiKeyHealthyIfNoRecentErrorsSource, but panics if it returns an error.
+// Should only be used in instances where the inputs are statically defined and known to be valid.
+func MustNewMultiKeyHealthyIfNoRecentErrorsSource(checkType health.CheckType, messageInCaseOfError string, windowSize time.Duration) KeyedErrorHealthCheckSource {
```
splucs: The constructors come between the struct definition and the methods. Let's keep consistent in this package.

splucs (changes requested):

status/health/window/key_error_source.go
```diff
@@ -82,6 +82,96 @@ func (m *multiKeyUnhealthyIfAtLeastOneErrorSource) HealthStatus(ctx context.Cont
...
+// windowSize must be a positive value, otherwise returns error.
+// Once a non-nil error has been submitted, this will be unhealthy until a nil error is submitted or `windowSize` time
+// has passed without a non-nil error. Submitting a non-nil error resets the timer and stays unhealthy
+func NewMultiKeyHealthyIfNoRecentErrorsSource(checkType health.CheckType, messageInCaseOfError string, windowSize time.Duration) (KeyedErrorHealthCheckSource, error) {
+	return newMultiKeyHealthyIfNoRecentErrorsSource(checkType, messageInCaseOfError, windowSize, NewOrdinaryTimeProvider())
+}
+
+func newMultiKeyHealthyIfNoRecentErrorsSource(checkType health.CheckType, messageInCaseOfError string, windowSize time.Duration, timeProvider TimeProvider) (KeyedErrorHealthCheckSource, error) {
+	return &multiKeyHealthyIfNoRecentErrorsSource{
+		windowSize:           windowSize,
```
splucs, line 127: Return an error if windowSize <= 0.

status/health/window/key_error_source_test.go
jhowarth-pal: Agreed, thanks for the suggestion. Added a configurable duration between submissions and evaluation, which is nice since it makes use of the default zero value and can be omitted except when needed

splucs (changes requested):

status/health/window/key_error_source.go
jhowarth-pal: Gotcha, done

splucs (approved):

splucs:
@bmoylan PR LGTM, but my approval doesn't count.

status/health/window/key_error_source.go
jhowarth-pal: Good call, done

jhowarth-pal:
Thanks for the help @splucs!

bmoylan (approved):
thanks both!

PR #216, opened 2020-10-26
Add WitchcraftDebugService routes

bmoylan:
---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/216)


witchcraft/wdebug/resource.go
```diff
@@ -0,0 +1,139 @@
...
+		return werror.WrapWithContextParams(ctx, errors.NewInvalidArgument(), "path param not present", werror.SafeParam("pathParamName", "diagnosticType"))
+	}
+	diagnosticType := DiagnosticType(diagnosticTypeStr)
+
+	return r.writeDiagnostic(ctx, diagnosticType, rw)
+}
+
+func (r *debugResource) writeDiagnostic(ctx context.Context, diagnosticType DiagnosticType, rw http.ResponseWriter) error {
+	switch diagnosticType {
+	case "threaddump.v1":
```
bmoylan: TBD: How to advertise this list of supported diagnostics

changelog/@unreleased/pr-216.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: feature
```
bmoylan, line 1: // note to self, add some language in the readme about this feature

witchcraft/server_routes.go
```diff
@@ -48,6 +49,11 @@ func (s *Server) addRoutes(mgmtRouterWithContextPath wrouter.Router, runtimeCfg
 	if err := addDebuggingRoutes(mgmtRouterWithContextPath); err != nil {
```
tabboud: nit: we may want to change this to `addPprofRoutes()` or similar to disambiguate the debug routes below.

Also need to update the comment to make it plural `// add debugging endpoints ...`

witchcraft/wdebug/resource.go
tabboud: nit: how about adding these as const args below the type definition. This would help for test cases
```go
const (
	threadDumpV1   DiagnosticType = "threaddump.v1"
	goProfileCPUV1 DiagnosticType = "go.profile.cpu.v1"
	// ...
)
```

witchcraft/wdebug/resource.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type DiagnosticType string
+
+type debugResource struct {
+	SharedSecret refreshable.String
+}
+
+func RegisterRoute(router wrouter.Router, sharedSecret refreshable.String) error {
+	debugResource := debugResource{SharedSecret: sharedSecret}
```
tabboud: does this need to be a pointer given all methods take pointer receivers?

witchcraft/wdebug/resource.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (r *debugResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) error {
+	ctx := req.Context()
+	authHeader, err := httpserver.ParseBearerTokenHeader(req)
+	if err != nil {
+		return errors.WrapWithNewError(err, errors.DefaultPermissionDenied)
+	}
+	if authHeader != r.SharedSecret.CurrentString() {
+		return errors.NewPermissionDenied()
```
tabboud: I'm surprised returning 403 was not defined in the spec as well, although I guess it can be assumed.

witchcraft/wdebug/resource.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (r *debugResource) getCPUProfileV1(ctx context.Context, rw http.ResponseWriter) error {
+	// If diagnostics ever support parameters, this can be configurable
+	const defaultDuration = 30 * time.Second
+
+	rw.Header().Set(headerKeyContentType, codecs.Binary.ContentType())
+	rw.Header().Set(headerKeySafeLoggable, "true")
+
+	if err := pprof.StartCPUProfile(rw); err != nil {
```
tabboud: Do you think it's worth storing any state to detect when we're in the middle of a profile already, so that we can return a more informed error. I noticed this will return an error if one is already in flight.

witchcraft/wdebug/resource.go
tabboud: By advertise, are you implying how an external system can identify what is supported by the current WGS version?
One option could be to add another route to the debug endpoint that returns the set of `DiagnosticType` that are supported.

bmoylan: So according to the spec, these types need to make it into the SLS manifest for the service so the CD infra knows what is available without having to poll/synchronize with the running service. I think this means we'll need to extend the current manfiest-building godel plugin which currently only supports hardcoded values. I was thinking about some kind of import graph discovery, like a magic decl name or comment that we can search for when building the dist.

witchcraft/wdebug/resource.go
bmoylan: I'd kind of prefer to let the std lib do its thing rather than try to reimplement it with something like a mutex. Right now you'll get a 500 response with a logged error like `failed to start CPU profile: cpu profiling already in use`. 

Updated to make sure that message makes it into the conjure error and called it a Conflict (409)

witchcraft/wdebug/handlers.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/palantir/conjure-go-runtime/v2/conjure-go-contract/codecs"
+	"github.com/palantir/conjure-go-runtime/v2/conjure-go-contract/errors"
+	"github.com/palantir/pkg/metrics"
+	werror "github.com/palantir/witchcraft-go-error"
+	wparams "github.com/palantir/witchcraft-go-params"
+)
+
+const (
+	DiagnosticTypeCPUProfile1MinuteV1 DiagnosticType = "go.profile.cpu.1minute.v1"
```
tabboud: Whats the reasoning behind moving this to 1m instead of 30s?
Also we will need to update the spec with this new notation / time value.

witchcraft/wdebug/handlers.go
```diff
@@ -0,0 +1,199 @@
...
+	DiagnosticTypeHeapProfileV1       DiagnosticType = "go.profile.heap.v1"
+	DiagnosticTypeAllocsProfileV1     DiagnosticType = "go.profile.allocs.v1"
+	DiagnosticTypeGoroutinesV1        DiagnosticType = "go.goroutines.v1"
```
tabboud: similar comment about the spec changes, given this went from threaddump to goroutines and the format of the content type is different

README.md
```diff
@@ -60,9 +60,20 @@ configuration, then `witchcraft-server` starts a second management server on the
 endpoints on that port. This can be useful in scenarios where all of the traffic to the main endpoints require client
 certificates for TLS but the status endpoints need to be served without requiring client TLS certificates.
 
-### Debug Routes
+### Debug & Diagnostic Routes
+
+Witchcraft servers register a route on the management server at `/debug/diagnostic/{diagnosticType}`, where
+diagnosticType represents a payload used for debugging a running server node. The response's `Content-Type` header
+specifies the encoding and format of the response. The following types are currently supported:
+* `go.goroutines.v1`: JSON-formatted representation of all running goroutines and their stacktraces
```
tabboud: Is the output here still JSON-formatted or plain text?

witchcraft/wdebug/handlers.go
```diff
@@ -0,0 +1,199 @@
...
+
+var diagnosticHandlers = map[DiagnosticType]DiagnosticHandler{
+	DiagnosticTypeCPUProfile1MinuteV1: handlerCPUProfile1MinuteV1{},
+	DiagnosticTypeHeapProfileV1:       handlerHeapProfileV1{},
+	DiagnosticTypeAllocsProfileV1:     handlerAllocsProfileV1{},
+	DiagnosticTypeGoroutinesV1:        handlerGoroutinesV1{},
+	DiagnosticTypeMetricNamesV1:       handlerMetricNamesV1{},
+}
+
+type DiagnosticHandler interface {
```
tabboud: Thoughts on keeping this private until we decide how to advertise everything in case we need to change anything and avoid breaks?
Alternatively we could move the entirety of wdebug to internal until the details are hashed out. I am not too concerned with the existing public api, so an internal move might not be worth it.

witchcraft/wdebug/resource_test.go
```diff
@@ -0,0 +1,66 @@
...
+	r := wrouter.New(whttprouter.New())
+	secret := refreshable.NewDefaultRefreshable("secret1")
+	err := RegisterRoute(r, refreshable.NewString(secret))
+	require.NoError(t, err)
+
+	server := httptest.NewServer(r)
+	defer server.Close()
+
+	for _, test := range []struct {
+		DiagnosticType DiagnosticType
```
tabboud: It would be nice to have coverage on the other types as well, but I won't block on it.

witchcraft/wdebug/handlers.go
bmoylan: This is to match what they're doing with JFR in Java, but I don't feel that strongly about it. We can also add multiple durations if we think people will want the flexibility.

README.md
bmoylan: Oops, fixed

witchcraft/wdebug/handlers.go
bmoylan: Yeah I'm going to be submitting a spec update today after some developments from both the Java and Go impls

witchcraft/wdebug/handlers.go
bmoylan: Sure, moving to internal where we can always pull it back out

witchcraft/wdebug/resource_test.go
bmoylan: added one for a pprof type, more will need some rework to set up a more real server (e.g. for the metrics one)

witchcraft/wdebug/handlers.go
tabboud: ok great. My main concern was due to request timeouts given the length of duration, but we can make sure these are set up appropriately for our clients.

tabboud (approved):
👍 
I am happy to move forward with this, especially now that you've moved it to internal in case we need to iterate further.

PR #222, opened 2020-11-06
Bump godel-conjure-plugins to v6

asanderson15:
## After this PR

==COMMIT_MSG==
Bump godel-conjure-plugins to v6
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)

### DIFF
--- a/changelog/@unreleased/pr-222.v2.yml
+++ b/changelog/@unreleased/pr-222.v2.yml
@@ -1,5 +1,5 @@
-type: improvement
-improvement:
-  description: Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention.
+type: break
+break:
+  description: Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention. This will be a break for downstream things vendoring WGS health libraries.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/222
--- a/changelog/@unreleased/pr-222.v2.yml
+++ b/changelog/@unreleased/pr-222.v2.yml
@@ -3,7 +3,8 @@ break:
   description: |
     Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention.
     This is a breaking change for code that references the `HealthComponent` interface and the `HealthState`
-    enum value constants.
+    enum value constants. Consumers should update references to use the new `HealthState` enum type and
+    `HealthState_Value` constants.
 
     In addition, this change also makes `HealthStateStatusCodes` unexported; consumers
     should switch to using the new `HealthStateStatusCode(state health.HealthState_Value) int` function instead.
--- a/changelog/@unreleased/pr-222.v2.yml
+++ b/changelog/@unreleased/pr-222.v2.yml
@@ -3,7 +3,12 @@ break:
   description: |
     Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention.
     This is a breaking change for code that references the `HealthComponent` interface and the `HealthState`
-    enum value constants. In addition, this change also makes `HealthStateStatusCodes` unexported; consumers
+    enum value constants.
+
+    In addition, this change also makes `HealthStateStatusCodes` unexported; consumers
     should switch to using the new `HealthStateStatusCode(state health.HealthState_Value) int` function instead.
+
+    Finally, the `{Starting|Healthy|Warning|Error}State` constants have been removed. Consumers should instead
+    create a new `HealthState` enum value directly.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/222
--- a/changelog/@unreleased/pr-222.v2.yml
+++ b/changelog/@unreleased/pr-222.v2.yml
@@ -1,5 +1,9 @@
 type: break
 break:
-  description: Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention. This will be a break for downstream things vendoring WGS health libraries.
+  description: |
+    Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention.
+    This is a breaking change for code that references the `HealthComponent` interface and the `HealthState`
+    enum value constants. In addition, this change also makes `HealthStateStatusCodes` unexported; consumers
+    should switch to using the new `HealthStateStatusCode(state health.HealthState_Value) int` function instead.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/222
--- a/changelog/@unreleased/pr-222.v2.yml
+++ b/changelog/@unreleased/pr-222.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention.
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/222
--- a/conjure/witchcraft/api/health/enums.conjure.go
+++ b/conjure/witchcraft/api/health/enums.conjure.go
@@ -13,37 +13,61 @@ import (
 
 var enumValuePattern = regexp.MustCompile("^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$")
 
-type HealthState string
+type HealthState struct {
+	val HealthState_Value
+}
+
+type HealthState_Value string
 
 const (
 	// The service node is fully operational with no issues.
-	HealthStateHealthy HealthState = "HEALTHY"
+	HealthState_HEALTHY HealthState_Value = "HEALTHY"
 	// The service node is fully operational with no issues; however, it is requesting to defer shutdown or restart. A deferring node should not accept "new" jobs but should allow polling of existing jobs.
-	HealthStateDeferring HealthState = "DEFERRING"
+	HealthState_DEFERRING HealthState_Value = "DEFERRING"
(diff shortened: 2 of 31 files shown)


asanderson15:
This fixes the enum breaks introduced by conjure-go v6.

nmiyake (changes requested):
> Reviewed 2 of 2 files at r1, 28 of 28 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: all files reviewed, 2 unresolved discussions (waiting on @asanderson15)

___
*[status/status.go, line 27 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLTeACQ6RmgG8V2-y4F:-MLTeACQ6RmgG8V2-y4G:b-q0llj7) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c45bb0d0f6852dcac29c32615cd93ac5f404a020/status/status.go#L27)):*
> ```Go
> )
> 
> var HealthStateStatusCodes = map[health.HealthState_Value]int{
> ```

I know this isn't directly related, but do you know why this variable is exported? This is a dangerous pattern because any library can do something like `status.HealthStateStatusCodes = nil` or modify the map contents and break internal code.

If clients need access to this map for its contents, then we should create a `HealthStateStatusCods() map[...]` function that returns a newly allocated map that has the same content but keep this variable unexported.

This would be the time to do it -- the current PR is making a breaking change for this field anyway (because it's changing the type), so might as well bundle it.

I think the above is reasonable, but if someone feels that it's not, then we could keep this field and comment on its danger but then just switch internal implementations to use a separate parallel private map with the same content.
___
*[status/reporter/healthcomponent.go, line 35 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLTflt58aQ6UboagNgq:-MLTflt58aQ6UboagNgr:bpursne) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c45bb0d0f6852dcac29c32615cd93ac5f404a020/status/reporter/healthcomponent.go#L35)):*
> ```Go
> }
> 
> var (
> ```

This isn't safe to do -- this allows any package to update/reassign the global values. Likely need to change to functions that return the desired value.

asanderson15 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: all files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[status/status.go, line 27 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLTeACQ6RmgG8V2-y4F:-MLTgsie-svvVEDDU5B4:b-b9ij47) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c45bb0d0f6852dcac29c32615cd93ac5f404a020/status/status.go#L27)):*

I know this isn't directly related, but do you know why this variable is exported? This is a dangerous pattern because any library can do something like `status.HealthStateStatusCodes = nil` or modify the map contents and break internal code.

If clients need access to this map for its contents, then we should create a `HealthStateStatusCods() map[...]` function that returns a newly allocated map that has the same content but keep this variable unexported.

This would be the time to do it -- the current PR is making a breaking change for this field anyway (because it's changing the type), so might as well bundle it.

I think the above is reasonable, but if someone feels that it's not, then we could keep this field and comment on its danger but then just switch internal implementations to use a separate parallel private map with the same content.
</blockquote></details>

Agree on this as well. Will make this break.
___
*[status/reporter/healthcomponent.go, line 35 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLTflt58aQ6UboagNgq:-MLTggLg1C-qxHFnfm83:brssrko) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/c45bb0d0f6852dcac29c32615cd93ac5f404a020/status/reporter/healthcomponent.go#L35)):*

This isn't safe to do -- this allows any package to update/reassign the global values. Likely need to change to functions that return the desired value.
</blockquote></details>

Yeah, that's fair. I almost removed this entirely because it seems completely redundant. I think since we're breaking things I'll just remove.

nmiyake (changes requested):
> Reviewed 9 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: all files reviewed, 2 unresolved discussions (waiting on @asanderson15)

___
*[changelog/@unreleased/pr-222.v2.yml, line 3 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiCLz-EuFNzb5zSh1w:-MLiCLz-EuFNzb5zSh1x:b-udg6oj) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/changelog/@unreleased/pr-222.v2.yml#L3)):*
> ```YAML
> type: break
> break:
>   description: Bump conjure-go to v6, which updates health status enum value constants to use the new naming convention. This will be a break for downstream things vendoring WGS health libraries.
> ```

nit: can we use more precise language here? Would like more precision around what and where the break is and the consumers it impacts, and remove the "vendoring" language (a project can import/use a package without vendoring).

`This is a breaking change for code that references the health constants declared in the witchcraft-go-server health packages {list of packages or specific package}`.

If the new code location is known would be helpful for the changelog to include instructions on how consumers can fix the break.
___
*[status/status.go, line 121 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiE4sECxlm_neGrUxV:-MLiE4sECxlm_neGrUxW:b-ng90wh) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/status/status.go#L121)):*
> ```Go
> }
> 
> func HealthStateStatusCode(state health.HealthState_Value) int {
> ```

Can you add documentation to this function?

asanderson15 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: 29 of 31 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-222.v2.yml, line 3 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiCLz-EuFNzb5zSh1w:-MLiQyYWCsfALZdqcHla:bn495nr) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/changelog/@unreleased/pr-222.v2.yml#L3)):*

nit: can we use more precise language here? Would like more precision around what and where the break is and the consumers it impacts, and remove the "vendoring" language (a project can import/use a package without vendoring).

`This is a breaking change for code that references the health constants declared in the witchcraft-go-server health packages {list of packages or specific package}`.

If the new code location is known would be helpful for the changelog to include instructions on how consumers can fix the break.
</blockquote></details>

Sorry, yeah, should have been clearer. This is updated now.
___
*[status/status.go, line 121 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiE4sECxlm_neGrUxV:-MLiQxBB0KedAQnA8IeO:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/status/status.go#L121)):*

Can you add documentation to this function?
</blockquote></details>

Done.

nmiyake (changes requested):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: all files reviewed, 2 unresolved discussions (waiting on @asanderson15 and @nmiyake)

___
*[status/status.go, line 121 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiE4sECxlm_neGrUxV:-MLiS6FR15HGvyX97rAq:bwo0n3n) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/status/status.go#L121)):*
<details><summary><i>Previously, asanderson15 (Adam Anderson) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I think it's worth calling out that `http.StatusInternalServerError` is returned for any unrecognized health state value

asanderson15 (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: 30 of 31 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[status/status.go, line 121 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/222#-MLiE4sECxlm_neGrUxV:-MLiUnfpBFagItmBZlN9:bombmv2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/47f9bbba00359e3e9a175bd4b8f22d4c7de8ed22/status/status.go#L121)):*

I think it's worth calling out that `http.StatusInternalServerError` is returned for any unrecognized health state value
</blockquote></details>

Good call. Done.

nmiyake (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/222)* status: all files reviewed, 1 unresolved discussion (waiting on @nmiyake)

PR #233, opened 2020-12-04
Add godel-sls-diagnostics-generator

bmoylan:
This PR adds a `tools/` module which includes a godel plugin for generated a `sls-diagnostics.json` manifest file in every package with types that implement `wdebug.DiagnosticHandler`.

It's written as a full godel plugin that can be published, but I didn't want to jump through the repo creation hoops until we get consensus on what we want to do. In the meantime, it is invoked as a go:generate command.

Eventually this logic can be used by downstream projects that declare their own diagnostic handlers, though this is blocked as long as we leave the package in `internal/` (an explicit decision made in #216 to not yet expose the types).

The build doesn't like my module setup, but I didn't want to introduce a bunch of deps for the main w-g-s module. Open to ideas for how we want to ship this.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233)


nmiyake:
Thanks for making this PR! Can you clarify the aspects on which you'd like to get feedback as part of this review? Should we be evaluating the high-level approach or reviewing the code as if it would merge? Is hooking this up via go generate in this repo done just so that the logic can be run, or are you proposing that it be merged into this repo in this form?

nmiyake (changes requested):
Some initial feedback.

At a high level, it would help to have some more function descriptions and an articulation of the contract for how things are parsed/interpreted to create the JSON output. It would also be helpful to have tests.

However, because this generator is specific to this project right now, it's effectively just a black box that is generating JSON output, and for now you know that the output that is being generated is sufficient for your purposes. From that perspective, it's not critical to articulate the full contract at this point -- because we don't have to worry about anyone else conforming to the contract, "the implementation is the contract" is acceptable at the current point.

Given your stated timeline for this, we could take the approach of just making sure there isn't anything egregiously wrong and just pushing through for now, and basically kick the can down the road and do the full level of due diligence when/if this is broken out as its own tool that other projects can consume.

Let me know how you'd like to proceed and will review further accordingly.

> Reviewed 15 of 675 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-:-MO5FnK3AvRdLD1xpdPg:b-63v9q2)* status: 16 of 675 files reviewed, 8 unresolved discussions (waiting on @bmoylan, @nmiyake, and @tabboud)

___
*[tools/godelw, line 1 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO55DS57jX3_qvxyIBv:-MO55DS57jX3_qvxyIBw:b-x2xz4x) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godelw#L1)):*
> ```
> #!/bin/bash
> ```

If we're not going to invoke any of these commands as part of CI, then I think it's probably not worth adding the separate `godelw` and `godel` config for this inner module.
___
*[tools/godel-sls-diagnostics-generator/README.md, line 2 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO562qj1snPpLectW9W:-MO562qj1snPpLectW9X:b1zzlzb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/README.md#L2)):*
> ```Markdown
> godel-sls-diagnostics-generator
> ============
> ```

Please fill out content as part of PR
___
*[tools/godel-sls-diagnostics-generator/cmd/generate.go, line 54 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO56VNT8E91HTZPH6rB:-MO56VNT8E91HTZPH6rC:b-5ttub3) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/cmd/generate.go#L54)):*
> ```Go
> }
> 
> func GodelConfigPackageMatcher(godelConfigFile string) (matcher.Matcher, error) {
> ```

It's a bit awkward for the `cmd` package to export a function like this. I think this is mainly because the generator main also wants to use this logic right now -- in a world where that didn't exist, this woudl likely be private.

However, given that the generator does exist, it might make more sense to decomp more of the logic so that this generate command and the generator main call into a single "runner" function that handles the shared logic for doing things like reading these excludes.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 28 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AFc_-eslaKE3bXLo:-MO5AFc_-eslaKE3bXLp:b-40tsrv) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L28)):*
> ```Go
> )
> 
> func LoadDiagnosticHandlerImpls(ctx context.Context, pkgs []string, projectDir string) (map[*packages.Package][]DiagnosticHandlerMetadata, error) {
> ```

Can you provide a comment that describes what this function is doing at a high level? This will also help me in my review of the logic.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AX4u8kIHmb1ukPIE:-MO5AX4u8kIHmb1ukPIF:b-vmb9ut) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L30)):*
> ```Go
> func LoadDiagnosticHandlerImpls(ctx context.Context, pkgs []string, projectDir string) (map[*packages.Package][]DiagnosticHandlerMetadata, error) {
> 	findResult, err := findimpls.Find(&findimpls.Query{
> 		LoadConfig: &packages.Config{
> ```

What is the expected contract/behavior around build constraints? It's fine to be restrictive/opinionated about them, but we should probably establish some form of contract.

For example, if we have "foo_darwin.go" and "foo_linux.go" and each of them declares different diagnostics, what is the expected behavior? We can forbid this case as part of our stated contract (and without enforcement in code), but I think it should be articulated somewhere.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/packages.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5CLjk4PjTKw7OVjOw:-MO5CLjk4PjTKw7OVjOx:bjzsu06) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/packages.go#L30)):*
> ```Go
> // PkgsInProject returns all the packages in the projectDir except those that match exclude.
> // Taken from https://github.com/palantir/okgo/blob/d5f6b9f4/cmd/check.go#L61-L96
> func PkgsInProject(projectDir string, exclude matcher.Matcher) ([]string, error) {
> ```

As stated here, I think this implementation should be fine given that this is only expected to run within the constraints of existing godel projects, which effectively means that the packages we care about will be within the project directory and we don't support submodules (we assume that it and any vendor directory will be part of the ignore directive).

Now that modules exist I think I would write this logic differently, but for now I think this should work (and if it's good enough for the rest of the godel layers, should be fine here as well :).
___
*[tools/godel-sls-diagnostics-generator/internal/findimpls/find.go, line 48 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5DCcO9ZQ91gTaHUdC:-MO5DCcO9ZQ91gTaHUdD:b-6wbo5b) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/internal/findimpls/find.go#L48)):*
> ```Go
> 
> // Find searches through packages specified in the Query for the InterfacePackage/InterfaceName provided.
> // Once the interface is found, all packages are searched for struct types which implement the interface.
> ```

Is there a reason we only care about struct types that implement the interface? Does this mean that if the interface is implemented by a non-struct type, then that type is not returned?
___
*[tools/godel-sls-diagnostics-generator/internal/findimpls/find.go, line 110 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5EZq3B4VK6A5d5YaO:-MO5EZq44HKX8Ae4WIaY:b-1qfsb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/internal/findimpls/find.go#L110)):*
> ```Go
> }
> 
> // For every struct type definition in the package, check if it implements the interface and collect all matches.
> ```

Can we add any tests? For example, I'd be curious to check/test whether the current logic identifies interface implementations against both structs and struct pointers.

I know that individual unit-level testing can be intractable for AST code, but I've found that it's often possible to set up tests that parse hard-coded code as a package/AST, and once that's done it becomes easier to write high-level tests like "given this source code, do I extract the expected values". Integration tests that run through the full workflow are also helpful, but at the very least I think it would help to have some of these code-level tests to be able to quickly test different edge cases.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233)* status: 7 of 678 files reviewed, 8 unresolved discussions (waiting on @nmiyake and @tabboud)

___
*[tools/godelw, line 1 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO55DS57jX3_qvxyIBv:-MO7Utny9IglgCnnvxLf:b-shcael) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godelw#L1)):*

If we're not going to invoke any of these commands as part of CI, then I think it's probably not worth adding the separate `godelw` and `godel` config for this inner module.
</blockquote></details>

Leaving them in here makes it easier to run manually/locally so I'm inclined to leave them unless you strongly disagree
___
*[tools/godel-sls-diagnostics-generator/README.md, line 2 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO562qj1snPpLectW9W:-MO8_Mmz0JgDvG7K0z3U:b-epqnyo) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/README.md#L2)):*

Please fill out content as part of PR
</blockquote></details>

Done! Let me know if you think there's more info that might be helpful
___
*[tools/godel-sls-diagnostics-generator/cmd/generate.go, line 54 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO56VNT8E91HTZPH6rB:-MO7Wvix8ULYGPGEhF9W:bod2ecc) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/cmd/generate.go#L54)):*

It's a bit awkward for the `cmd` package to export a function like this. I think this is mainly because the generator main also wants to use this logic right now -- in a world where that didn't exist, this woudl likely be private.

However, given that the generator does exist, it might make more sense to decomp more of the logic so that this generate command and the generator main call into a single "runner" function that handles the shared logic for doing things like reading these excludes.
</blockquote></details>

yeah agreed, moved to internal/godelproject and collapsed them into one function
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 28 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AFc_-eslaKE3bXLo:-MO7Y5Z22qOQ1XLh8NWl:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L28)):*

Can you provide a comment that describes what this function is doing at a high level? This will also help me in my review of the logic.
</blockquote></details>

Done.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AX4u8kIHmb1ukPIE:-MO7Yb644oti-wQkry_w:b-ir412q) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L30)):*

What is the expected contract/behavior around build constraints? It's fine to be restrictive/opinionated about them, but we should probably establish some form of contract.

For example, if we have "foo_darwin.go" and "foo_linux.go" and each of them declares different diagnostics, what is the expected behavior? We can forbid this case as part of our stated contract (and without enforcement in code), but I think it should be articulated somewhere.
</blockquote></details>

Added to the function comment, but if you don't set anything it uses the local constraints
___
*[tools/godel-sls-diagnostics-generator/internal/findimpls/find.go, line 48 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5DCcO9ZQ91gTaHUdC:-MO8RXOL4NjUTO4osgTo:b-mkadf2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/internal/findimpls/find.go#L48)):*

Is there a reason we only care about struct types that implement the interface? Does this mean that if the interface is implemented by a non-struct type, then that type is not returned?
</blockquote></details>

Interface types were causing problems, but it turns out those are the only ones that need to be excluded. I've broadened this and added test cases for other types (a builtin and a function type)
___
*[tools/godel-sls-diagnostics-generator/internal/findimpls/find.go, line 110 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5EZq3B4VK6A5d5YaO:-MO8RSyw031pxc_itUWO:b-v0di48) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/internal/findimpls/find.go#L110)):*

Can we add any tests? For example, I'd be curious to check/test whether the current logic identifies interface implementations against both structs and struct pointers.

I know that individual unit-level testing can be intractable for AST code, but I've found that it's often possible to set up tests that parse hard-coded code as a package/AST, and once that's done it becomes easier to write high-level tests like "given this source code, do I extract the expected values". Integration tests that run through the full workflow are also helpful, but at the very least I think it would help to have some of these code-level tests to be able to quickly test different edge cases.
</blockquote></details>

Yeah I added some test code and an integration test, let me know what you think!

I did encode line numbers which is pretty brittle, but given that they are lines of test-specific code I think that's acceptable.

nmiyake (changes requested):
Not fully finished, but submitting another round of feedback (sending email now on process)

> Reviewed 5 of 77 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-:-MOD6YbU3Pc05167GByi:bngui0d)* status: 12 of 679 files reviewed, 9 unresolved discussions (waiting on @bmoylan, @nmiyake, and @tabboud)

___
*[tools/godelw, line 1 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO55DS57jX3_qvxyIBv:-MOAKm2BCRDIm4O9_ppL:bzelp8x) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godelw#L1)):*

Leaving them in here makes it easier to run manually/locally so I'm inclined to leave them unless you strongly disagree
</blockquote></details>

Sure, that's fine, especially given that the intention is for this to be short-lived. Main concern/drawback over the long term is that this isn't updated by things like excavator and not run as part of CI so very susceptible to decay.
___
*[tools/godel-sls-diagnostics-generator/README.md, line 8 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOAJINn9jAWyvwCAdWv:-MOAJINn9jAWyvwCAdWw:b-hy9ifz) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/823816691ef4b6977d67efcfc88e4316616665e7/tools/godel-sls-diagnostics-generator/README.md#L8)):*
> ```Markdown
> 
> This operation happens in three steps:
> 1. Load all packages managed by the godel project and find the wdebug.DiagnosticHandler interface type.
> ```

If it's known that the interface type we want is `wdebug.DiagnosticHandler`, why does it need to be searched for in the project?

Is this just an implementation issue where you want a reference to the interface "object" and thus search for it within the current project packages?

Also, even for the above, `wdebug` is an external package, so it's not found within the project packages, right? I'm assuming that you're loading all project packages including dependencies and then basically assuming/requiring that `wdebug` must be in the package dependency graph and basically locating that specific package within the graph?
___
*[tools/godel-sls-diagnostics-generator/README.md, line 10 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOAJ0jR9BEFDh1oc91A:-MOAJ0jR9BEFDh1oc91B:b-1tnxow) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/823816691ef4b6977d67efcfc88e4316616665e7/tools/godel-sls-diagnostics-generator/README.md#L10)):*
> ```
> implemenatations
> ```

typo: "implementations"
___
*[tools/godel-sls-diagnostics-generator/README.md, line 43 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOAJEuL3L0rUEpJseAa:-MOAJEuL3L0rUEpJseAb:b-wr7fi) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/823816691ef4b6977d67efcfc88e4316616665e7/tools/godel-sls-diagnostics-generator/README.md#L43)):*
> ```Markdown
>   ]
> }
> ```
> ```

nit: file should end with newline
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 28 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AFc_-eslaKE3bXLo:-MOCbkED-qY1cc1kjeQh:b-9xrplm) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L28)):*

Done.
</blockquote></details>

Thanks, this is very helpful.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AX4u8kIHmb1ukPIE:-MOD5oM77xJilhj8LEyY:b8fmacl) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L30)):*

Added to the function comment, but if you don't set anything it uses the local constraints
</blockquote></details>

OK. Although it's an edge case, this does pose a possible issue since local dev environments are often different environments than the CI environment. This means that any interfaces/implementations that are behind an OS or arch-based build constraint could create different output locally versus in CI.

The most thorough/correct way to handle this would probably be to parse code independent of build constraints and either generate the output for all build constraints or raise an error if the output is build-constraint dependent. However, don't think this is necessary for the first pass.

bmoylan (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233)* status: 11 of 679 files reviewed, 8 unresolved discussions (waiting on @nmiyake and @tabboud)

___
*[tools/godel-sls-diagnostics-generator/README.md, line 8 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOAJINn9jAWyvwCAdWv:-MODA_bj9wQHP9TWx5cp:b-dycvrm) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/823816691ef4b6977d67efcfc88e4316616665e7/tools/godel-sls-diagnostics-generator/README.md#L8)):*

If it's known that the interface type we want is `wdebug.DiagnosticHandler`, why does it need to be searched for in the project?

Is this just an implementation issue where you want a reference to the interface "object" and thus search for it within the current project packages?

Also, even for the above, `wdebug` is an external package, so it's not found within the project packages, right? I'm assuming that you're loading all project packages including dependencies and then basically assuming/requiring that `wdebug` must be in the package dependency graph and basically locating that specific package within the graph?
</blockquote></details>

Yup exactly -- we need the pointer to the go object so we can get its method signature to compare against using `types.Implements()`.

If that package is not in the project packages or dependencies, we return an error. I assumed that this plugin would only be applied to WGS projects, but we could make it exit gracefully if you think that would be better.
___
*[tools/godel-sls-diagnostics-generator/README.md, line 10 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOAJ0jR9BEFDh1oc91A:-MODAREeAIGX05C98Ide:b-896fix) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/823816691ef4b6977d67efcfc88e4316616665e7/tools/godel-sls-diagnostics-generator/README.md#L10)):*

> ```
> implemenatations
> ```

typo: "implementations"
</blockquote></details>

Done.
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 30 at r2](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MO5AX4u8kIHmb1ukPIE:-MOD8qDwBNo4l-S8wSaF:b1on8bw) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/9dd44902eb58363f355a705f38cd63043b417041/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L30)):*

OK. Although it's an edge case, this does pose a possible issue since local dev environments are often different environments than the CI environment. This means that any interfaces/implementations that are behind an OS or arch-based build constraint could create different output locally versus in CI.

The most thorough/correct way to handle this would probably be to parse code independent of build constraints and either generate the output for all build constraints or raise an error if the output is build-constraint dependent. However, don't think this is necessary for the first pass.
</blockquote></details>

Yeah especially since this happens in the library we don't really know how consumers will compile it. I think the best we can do in the particular case of diagnostics where false entries just return errors, the best we could do is ignore all build constraints and search all files no matter the tags. I agree that we can punt this down the road, though.

nmiyake (changes requested):
Thanks for updating! Some feedback. Core of the change looks good pending the small items of feedback, but would like to discuss the point I flagged about how we want to write the output (JSON versus Go file), so publishing comments with a block rather than an approve until we sort that out.

> Reviewed 625 of 675 files at r1, 41 of 77 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-:-MOFjGV68hEQoJleaex9:b-gyb83r)* status: all files reviewed, 12 unresolved discussions (waiting on @bmoylan and @tabboud)

___
*[changelog/@unreleased/pr-233.v2.yml, line 6 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFaKNA1Ml-sZMiuP67:-MOFaKNA1Ml-sZMiuP68:bn6lvep) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/changelog/@unreleased/pr-233.v2.yml#L6)):*
> ```YAML
>     Add godel-sls-diagnostics-generator
> 
>     This PR adds a `tools/` module which includes a godel plugin for generated a `sls-diagnostics.json` manifest file in every package with types that implement `wdebug.DiagnosticHandler`. Consuming products can read these files to know what diagnostics are implemented in each pacakge.
> ```

nit: since this is changelog content that will become part of the release notes, can you update this content to gear more towards that?  I think the main focus should be that the PR adds the diagnostics JSON and a mechanism that keeps it up-to-date for the repository.
___
*[tools/godel-sls-diagnostics-generator/generate.go, line 45 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFbLH54z9nCSH2OClo:-MOFbLH54z9nCSH2OClp:be1yezd) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/generate.go#L45)):*
> ```Go
> 		return err
> 	}
> 	if err := os.Chdir(projectDir); err != nil {
> ```

It's fine to keep this just to be safe, but is it actually necessary given that `projectDir` is an input to the subsequent functions? In general, depending on/requiring a specific working directory can be pretty fragile, so just thought I'd mention it (although `go generate` is a pretty special case where the guarantees around this are much stronger)
___
*[tools/godel-sls-diagnostics-generator/README.md, line 5 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFc0Os4NUBWPQV22Ce:-MOFc0Os4NUBWPQV22Cf:be7zbji) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/README.md#L5)):*
> ```Markdown
> 
> godel-sls-diagnostics-generator is a [godel](https://github.com/palantir/godel) plugin which generates a file named `sls-diagnostics.json` in each package containing types that implement `wdebug.DiagnosticHandler`.
> This allows downstream projects to identify at build-time what diagnostics are supported by a dependency.
> ```

Can you add a section (either before or after this current introductory one) that describes why this is internal/contained within this WGS project rather than being external? This section would obviously be removed if/when this is broken out into its own repository, but in the meantime I think it is helpful context to have here.
___
*[tools/godel-sls-diagnostics-generator/README.md, line 8 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFcXP-4hy58l3R2S83:-MOFcXP-4hy58l3R2S84:b-ecol8b) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/README.md#L8)):*
> ```Markdown
> 
> This operation happens in three steps:
> 1. Load all packages managed by the godel project and find the wdebug.DiagnosticHandler interface type.
> ```

Per the discussion here and the one that we had, can you add a bit of clarification here to specify the full package name of the `wdebug` package that's searched for and that it's known to be in the package graph because implementing the interface requires referencing a type in the package?
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/api.go, line 21 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFhXZc074fvrFabvK3:-MOFhXZc074fvrFabvK4:bkwawl0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/api.go#L21)):*
> ```Go
> 	wdebugInterfaceName = "DiagnosticHandler"
> 
> 	diagnosticsJSONPath = "sls-diagnostics.json"
> ```

I'd like to pick your brains on an alternate approach to this I tried in another project: rather than writing out a `.json` file directly, what do you think about writing out a `.go` file with a custom build constraint (something like `// +build json`) with content consisting just of the package declaration and then a comment of the form `// json:` followed by the JSON content?

The main advantage of this approach is on the consumption side -- you can load all packages using the custom build constraint to identify the files of interest. Also, by making the file an actual `.go` file, it ensures that it is loaded natively in all loader implementations as part of the API, and also provides some future-proofing in case there are vendoring or module loading implementations that decide to omit non-Go files.

Would be intersted in discussing this further since I think it's worth standardizing how we handle this for the class of things we might do this for (and sorry for not flagging earlier -- had forgot about this myself until I went back and looked at the other project I worked on that did something similar to this).
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go, line 30 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFdkab3uqmuTKAosbw:-MOFdkab3uqmuTKAosbx:b-3mgr9w) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go#L30)):*
> ```Go
> )
> 
> func Generate(ctx context.Context, pkgs []string, verify bool, projectDir string, stdout io.Writer) error {
> ```

Can you add a comment to this exported function?
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go, line 48 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFfK5S4riHHiRMLBGL:-MOFfK5S4riHHiRMLBGM:b-eqzso0) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go#L48)):*
> ```Go
> 		if err != nil {
> 			if len(impls) > 0 {
> 				return nil, fmt.Errorf("package contained interface implementations but no detectable files: %v", err)
> ```

Any reason not to use the `errors` package to include stack information?
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go, line 50 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFe0t13DsfXZDedwhR:-MOFe0t13DsfXZDedwhS:b-dgj7so) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go#L50)):*
> ```Go
> 				return nil, fmt.Errorf("package contained interface implementations but no detectable files: %v", err)
> 			}
> 			continue
> ```

Is it intentional that the error from `getPackageDiagnosticsOutputPath` is ignored if `len(impls) == 0`? If so, would be helpful to have a comment on the continue explaining why
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go, line 69 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFequv4zxz3nnspss3:-MOFequv4zxz3nnspss4:bgkkw9s) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go#L69)):*
> ```Go
> 	for outputPath, outputContent := range files {
> 		exists := false
> 		if _, err := os.Stat(outputPath); err == nil {
> ```

Wouldn't it be more accurate to check that non-nil error is `os.IsNotExist`? Not sure if this is actually possible, but logically there could be a case where a `stat` call returns an error but the file does exist (and could be overwritten)
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go, line 124 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFfoFqB7w2v_7ct3ZM:-MOFfoFqB7w2v_7ct3ZN:b-qmqllz) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/generate.go#L124)):*
> ```Go
> 	})
> 
> 	metadataJSON, err := json.MarshalIndent(SLSDiagnosticsWrapper{impls}, "", "  ")
> ```

Should this use `safejson` package instead?
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 29 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFg6eN136a-msiSMRx:-MOFg6eN136a-msiSMRy:bz7z47v) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L29)):*
> ```Go
> 
> // LoadDiagnosticHandlerImpls returns a mapping from each package to metadata describing its implementations
> // of wdebug.DiagnosticHandler. It loads all go packages specified in pkgs, finds the interface type and its methods,
> ```

It also loads all their dependent package, right?
___
*[tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go, line 69 at r4](https://reviewable.io/reviews/palantir/witchcraft-go-server/233#-MOFg_eu0Fc1t8ZqSTG-:-MOFg_eu0Fc1t8ZqSTG0:b-8yxpeb) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/e3df7c54a53387ed7a799c9762686fb8005de131/tools/godel-sls-diagnostics-generator/diagnosticsplugin/load.go#L69)):*
> ```Go
> 				return nil, fmt.Errorf("failed to extract return value from SafeLoggable() method for impl %s: %v", typ.String(), err)
> 			}
> 			result[pkg] = append(result[pkg], DiagnosticEntry{
> ```

Is it valid if there are multiple DiagnosticEntry values where `(Type, Docs, Safe)` are identical (within the same package or across different packages)?

bmoylan:
As discussed offline, we're going to move this to a godel plugin used by downstream product repos rather than generate intermediate files within libraries.

PR #234, opened 2020-12-07
Rename debug.shared-secret to diagnostics.debug-shared-secret

bmoylan:
This is a change to the config introduced in #216 to match the internal java implementation's config location.

This is obviously a breaking change, but we are tracking the 1 deployed product depending on this config (at pal) and I will just handle the upgrade. I think eating that one-time cost is worth it to get consistency with witchcraft-java wherever we can.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/234)

### DIFF
--- a/config/runtime.go
+++ b/config/runtime.go
@@ -23,13 +23,13 @@ import (
 // Runtime specifies the base runtime configuration fields that should be included in all witchcraft-server-go
 // server runtime configurations.
 type Runtime struct {
-	DebugConfig  DebugConfig        `yaml:"debug,omitempty"`
-	HealthChecks HealthChecksConfig `yaml:"health-checks,omitempty"`
-	LoggerConfig *LoggerConfig      `yaml:"logging,omitempty"`
+	DiagnosticsConfig DiagnosticsConfig  `yaml:"debug,omitempty"`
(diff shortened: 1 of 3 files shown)


config/runtime.go
```diff
@@ -23,13 +23,13 @@ import (
 // Runtime specifies the base runtime configuration fields that should be included in all witchcraft-server-go
 // server runtime configurations.
 type Runtime struct {
-	DebugConfig  DebugConfig        `yaml:"debug,omitempty"`
-	HealthChecks HealthChecksConfig `yaml:"health-checks,omitempty"`
-	LoggerConfig *LoggerConfig      `yaml:"logging,omitempty"`
+	DiagnosticsConfig DiagnosticsConfig  `yaml:"debug,omitempty"`
```
sponcebus: Need to rename yaml field to `diagnostics`.

bmoylan: oops, thanks!

nmiyake (approved):
We discussed this, and we're comfortable with the break being done in this manner in this case given that this field was introduced so recently and we have determined that the impact will be minimal. We also plan to look at implementing better process/controls for creating or changing fields that may be serialized so that we can avoid this situation going forward.

PR #247, opened 2021-01-18
Replace negroni recovery middleware

### LINKED ISSUE #246: Improved panic logging
## What happened?
I was investigating an error in a service which ended up being a failed type coercion that caused a panic. I was surprised that the panic itself was not logged - I only discovered it because the server response included a panic message.

Panics should be logged in recovery, ideally with a relevant traceId. This would have made debugging the case above trivial - the panics would have been discoverable in the logs generally, and drilling down into a specific traceId would have given the full picture of what happened during during the request.

bmoylan:
## Before this PR
Panic responses from negroni are HTML formatted and not logged correctly.

## After this PR
==COMMIT_MSG==
Replace negroni recovery middleware to return conjure error responses and improve logging
==COMMIT_MSG==

I removed most of the forked copy of negroni, leaving only the response_writer and the license file -- let me know if there's a good reason to keep them around (or if we just want to copy the ResponseWriter into the middleware package and be done with Negroni)

Fixes #246 (cc @sandorw)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/247)


witchcraft/internal/middleware/route.go
```diff
@@ -107,3 +115,37 @@ func NewRouteLogTraceSpan() wrouter.RouteHandlerMiddleware {
 		next(rw, req, reqVals)
 	}
 }
+
+// NewRoutePanicRecovery returns a middleware which recovers panics within the inner route handler.
+// This is distinct from NewRequestPanicRecovery in that it runs when all logging/telemetry are configured on the request.
+func NewRoutePanicRecovery() wrouter.RouteHandlerMiddleware {
```
asanderson15, line 121: Is there any way to add a test for this?

bmoylan: This is tested in the integration test by the routes that panic inside the handler functions

asanderson15: Ah, I see - got it.

asanderson15 (approved):

tabboud (approved):

PR #249, opened 2021-01-28
Deprecate refreshable types in favor of palantir/pkg/refreshable

bmoylan:
We are creating palantir/pkg/refreshable in https://github.com/palantir/pkg/pull/196 to faciliate a smaller dependency footprint. This PR deprecates the old types in a backwards-compatible way, delegating to pkg.

I've left the `FileRefreshable` types here, given they are more closely tied to Witchcraft and have more specific logging requirements we don't want in pkg.

DRAFT until https://github.com/palantir/pkg/pull/196 is merged and there is a first release

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/249)


tabboud (approved):
👍

PR #251, opened 2021-02-25
Add TCP logging output

tabboud:
## Before this PR
No support to publish logs to a network log target.

## After this PR

==COMMIT_MSG==
Add support to write JSON encoded log lines to a TCP socket and expose service-discovery runtime config.

Services can now declare an "sls-log-tcp-json-receiver" service-discovery block, where logs will be multiplexed between the TCP receiver and the default log output destination. The log lines will be wrapped in an envelope that contains metadata about the source origin, including hostname, product, product version, etc. The TCP receiver is expected to use TLS authentication so valid security material must also be configured under the service-discovery block.
==COMMIT_MSG==

## Notes about the architecture
- All but the `service-discovery` config is internal for the TCP writer, so we will be free to make any changes without worrying about breaking consumers.
- Rather than go for a full blown TCP connection pool, I decided to implement a simple TCP connection provider that will round robin the set of URIs, establish a connection and return the connection to the writer. It is up to the writer to handle the lifecycle of the `net.Conn`. `net.Conn` can be used concurrently, so it is currently cached on the writer and a new one is retrieved if a connection is closed.
- All loggers share a single `TCPWriter`, which is in line with the Java implementation. We could make it so each logger has it's own `TCPWriter` especially if there is a lot of contention. This would require significantly more changes to how we instantiate our loggers today and would increase the overall number of outgoing connections.
- The envelope metadata is static, but we currently `json.Marshal` this along with the payload on every write, which is not very efficient. However without doing some manual JSON manipulation (i.e. serialize the envelope then remove the last "}" and jam in the `,"payload":payload}`), there is not an easy way to pre-serialize the metadata.
- The `TCPWriter` implements the `io.Writer` interface which lets it be used for the wlog output. Since witchcraft-go-server uses wlog-zap internally we are safe to assume all log output is JSON. However if we ever change that then this implied API will be broken. Also based on the zap docs, each call to `Write()` should contain an entire log payload, which deviates slightly from the `io.Writer` interface. See `Entry` here: https://pkg.go.dev/go.uber.org/zap@v1.16.0/zapcore#Core.

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/251)


config/runtime.go
```diff
@@ -17,15 +17,17 @@ package config
...
 // Runtime specifies the base runtime configuration fields that should be included in all witchcraft-server-go
 // server runtime configurations.
 type Runtime struct {
-	DiagnosticsConfig DiagnosticsConfig  `yaml:"diagnostics,omitempty"`
-	HealthChecks      HealthChecksConfig `yaml:"health-checks,omitempty"`
-	LoggerConfig      *LoggerConfig      `yaml:"logging,omitempty"`
+	DiagnosticsConfig DiagnosticsConfig         `yaml:"diagnostics,omitempty"`
+	HealthChecks      HealthChecksConfig        `yaml:"health-checks,omitempty"`
+	LoggerConfig      *LoggerConfig             `yaml:"logging,omitempty"`
+	ServiceDiscovery  httpclient.ServicesConfig `yaml:"service-discovery,omitempty"`
```
tabboud: Given this is a public facing addition, I am happy to make this a pre-commit to differentiate between the internal TCP writer things. Although this shows how it would be used.

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/url"
+	"sync/atomic"
+
+	werror "github.com/palantir/witchcraft-go-error"
+)
+
+// ConnProvider defines the behavior to retrieve an established net.Conn.
+type ConnProvider interface {
+	// GetConn returns a net.Conn or an error if there is no connection established.
+	// It is the users responsibility to close the returned net.Conn.
```
nmiyake: nit: "users" -> "caller's"

nmiyake: Is it safe for any given caller to close the connection, or will doing so possibly close the connection used by all callers?

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	ErrNoURIs           = "no URIs provided to connect to"
+	ErrFailedParsingURI = "failed to parse provided URI"
+	ErrFailedDial       = "failed to dial the host:port"
+)
+
+var _ ConnProvider = (*TCPConnProvider)(nil)
+
+// TCPConnProvider implements a ConnProvider that will round-robin TCP connections to a specific set of hosts.
+type TCPConnProvider struct {
```
nmiyake: Since `ConnProvider` is the interface that callers interact with, this struct can probably be private

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+	hostIdx   int32
+	hosts     []string
+	tlsConfig *tls.Config
+}
+
+func NewTCPConnProvider(uris []string, tlsCfg *tls.Config) (*TCPConnProvider, error) {
```
nmiyake: Return value can be `ConnProvider`

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+	if len(uris) < 1 {
+		return nil, werror.Error(ErrNoURIs)
+	}
+
+	// Extract the host:port from each provided URI to simplify establishing a Dial connection
+	var hosts []string
+	for _, uri := range uris {
+		u, err := url.Parse(uri)
+		if err != nil {
+			return nil, werror.Error(ErrFailedParsingURI)
```
nmiyake: Include the string of the URI that failed to parse in the output so that it's easier to debug/determine what went wrong

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+	ErrNoURIs           = "no URIs provided to connect to"
+	ErrFailedParsingURI = "failed to parse provided URI"
+	ErrFailedDial       = "failed to dial the host:port"
+)
+
+var _ ConnProvider = (*TCPConnProvider)(nil)
+
+// TCPConnProvider implements a ConnProvider that will round-robin TCP connections to a specific set of hosts.
+type TCPConnProvider struct {
+	hostIdx   int32
```
nmiyake: Comment would be helpful -- note that this is the index of the next host that will be handed out

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return &TCPConnProvider{
+		hostIdx:   0,
+		hosts:     hosts,
+		tlsConfig: tlsCfg,
+	}, nil
+}
+
+func (s *TCPConnProvider) GetConn() (net.Conn, error) {
+	oldIdx := atomic.LoadInt32(&s.hostIdx)
```
nmiyake: Current logic works, but is a bit confusing to follow -- specifically, it's a bit strange to me that the idx being stored is the "next" one, but before the mod operation is applied, meaning that the value in the struct as stored may actually be out of bounds until the mod operation is applied.

Think it would be cleaner to have the `s.hostIdx` be the index of the next host that will be handed out, and then apply the mod operation before updating.

If the logic is written this way to support the case where the length of `s.hosts` might change between calls then maybe the current approach makes more sense, but doubt that is the case given that there isn't any code to guarantee the safety around that slice between accesses (and if this is the case, then I think it would be helpful to have a comment to explain so).

witchcraft/internal/tcpjson/conn_provider_test.go
```diff
@@ -0,0 +1,88 @@
...
+				require.EqualError(t, err, tc.expectedErr)
+			} else {
+				require.NoError(t, err)
+				require.NotNil(t, connProvider)
+			}
+		})
+	}
+}
+
+func TestGetConn(t *testing.T) {
```
nmiyake, line 76: Can you add a brief comment to explain what this is testing?

witchcraft/internal/tcpjson/envelope.go
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+package tcpjson
+
+import (
+	"encoding/json"
+	"os"
+)
+
+type SlsEnvelopeV1 struct {
```
nmiyake: nit: should be `SLSEnvelopeV1`.

Is this specification taken from somewhere? If so, might be helpful to include link.

witchcraft/witchcraft.go
```diff
@@ -601,17 +603,42 @@ func (s *Server) Start() (rErr error) {
...
+	// extract the TCP JSON receiver from the runtime config
+	servicesCfg := baseRefreshableRuntimeCfg.CurrentBaseRuntimeConfig().ServiceDiscovery
+	receiverCfg, err := servicesCfg.MustClientConfig("sls-log-tcp-json-receiver")
+	if err == nil && len(receiverCfg.URIs) > 0 {
+		tlsConfig, err := tlsconfig.NewClientConfig(
+			tlsconfig.ClientRootCAFiles(receiverCfg.Security.CAFiles...),
+			tlsconfig.ClientKeyPairFiles(receiverCfg.Security.CertFile, receiverCfg.Security.KeyFile),
+		)
+		if err != nil {
+			return err
```
nmiyake: Can we add message/context to error that says that security material could not be created for the TCP JSON receiver?

witchcraft/witchcraft.go
```diff
@@ -601,17 +603,42 @@ func (s *Server) Start() (rErr error) {
...
+		}
+		envelopeMetadata := tcpjson.GetEnvelopeMetadata()
+		tcpWriter := tcpjson.NewTCPWriter(envelopeMetadata, connProvider)
+		defer func() {
+			_ = tcpWriter.Close()
+		}()
+		// re-initialize the loggers with the TCP writer
+		s.initLoggers(baseInstallCfg.UseConsoleLog, wlog.InfoLevel, metricsRegistry, tcpWriter)
```
nmiyake, line 638: Do we know that running this again is safe? Would it make more sense to do the TCP writer logic first and then pass `tcpWriter` to the place where `initLoggers` is currently called?

witchcraft/witchcraft.go
```diff
@@ -601,17 +603,42 @@ func (s *Server) Start() (rErr error) {
 	ctx = metrics.WithRegistry(ctx, metricsRegistry)
 
 	// initialize loggers
-	s.initLoggers(baseInstallCfg.UseConsoleLog, wlog.InfoLevel, metricsRegistry)
-
-	// add loggers to context
-	ctx = s.withLoggers(ctx)
```
nmiyake, line 607: Is there any logic that logs on the context loggers in `s.initRuntimeConfig`? That would be impacted by the removal here.

witchcraft/witchcraft.go
nmiyake: Ah OK I see, this requires the runtime config to be loaded first. Then would just want to double-check the contract for the init loggers call.

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -0,0 +1,139 @@
...
+		NodeID:         d.metadata.NodeID,
+		Service:        d.metadata.Service,
+		ServiceID:      d.metadata.ServiceID,
+		Stack:          d.metadata.Stack,
+		StackID:        d.metadata.StackID,
+		Product:        d.metadata.Product,
+		ProductVersion: d.metadata.ProductVersion,
+		Payload:        logPayload,
+	}
+	b, err := json.Marshal(envelopeToWrite)
```
nmiyake: Should we use `safejson.Marshal` here? I seem to recall that the default adds a newline and doesn't handle certain encodings etc. (which was our motivation for having `safejson`)

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -0,0 +1,141 @@
...
+func NewTCPWriter(metadata LogEnvelopeMetadata, provider ConnProvider) *TCPWriter {
+	return &TCPWriter{
+		metadata:   metadata,
+		provider:   provider,
+		closedChan: make(chan struct{}),
+		conn:       nil,
+	}
+}
+
+func (d *TCPWriter) Write(logPayload []byte) (int, error) {
```
tabboud: I am starting to realize that this breaks the contract for an `io.Writer` since we end up writing more than `len(logPayload)` bytes to the underlying connection.

nmiyake (changes requested):
Thanks for submitting this -- was very clear to follow, especially with the overall contextual comment. I think the overall approach looks good. High-level feedback:

* Yes, would appreciate separate PR that adds the configuration first/separately
* Wanted to clarify a bit about the last bullet point in your overview. Your point is basically that since the TCP writer wraps the loggers, if the wrapped loggers don't emit JSON then the TCP writer will wrap a payload that's not actually JSON -- is that correct? If so, I think that's acceptable -- the contract can just be stipulated that the TCP writer's payload will be the output of the underlying loggers, whether that be JSON, text, or anything else. Since the envelope itself is fully marshalled as JSON, I think the output will still be valid JSON even if the payload isn't, which I think is the only guarantee that's important.

I am actually a bit concerned about the use of `json.Marshal` -- I think that's known to be a pretty inefficient implementation in terms of both space and time, and since logging is such a hot path I think it could be significant.

Could we write benchmarks to test this? The main other approaches that come to mind are pre-creating the JSON wrapper and printing the provided bytes into it (but then downside is that if provided bytes are not valid JSON then output will not be either), or even using a zerolog instance here that constructs the output that way.

witchcraft/internal/tcpjson/tcp_logger.go
nmiyake: I think this should be OK from a contract standpoint -- the API reports the number of input bytes written/processed, so the amount that we write to the underlying connection doesn't really matter.

However, it is true that the return value should reflect that (so in this example, if we write all the input, then it needs to return `len(logPayload)` rather than the number of byte that were written to the connection)

nmiyake: For reference, https://stackoverflow.com/questions/32571504/what-number-of-bytes-written-should-my-io-writer-return-when-it-intentionally#comment52998337_32571504

tabboud: Thanks for clarifying this. I was getting tripped up thinking all io.Writer's had to return the total bytes actually written rather than the bytes processed of the original input.

witchcraft/internal/tcpjson/conn_provider.go
tabboud: Doing so will close the connection for all callers. If the connection is closed between the time of retrieving and then writing then a "use of closed connection" error will be returned. Added this to the doc comment as well.

In our case the TCPWriter will catch this error and will drop the log line, which may not be ideal. In order to get around this, we would have to not share a single `net.Conn` across all writers, but instead have each writer take a lock on each `net.Conn`, so we would lose 1 log line at the max.

witchcraft/internal/tcpjson/conn_provider.go
tabboud: I usually like to return concrete types in situations where an implementation could satisfy multiple interfaces rather than defining a single large interface that groups everything. In this case there is only 1 method, so I am fine to do make it private and return the interface.

witchcraft/internal/tcpjson/conn_provider.go
tabboud: I see how that can be confusing and think your suggestion makes a lot of sense. Given we're currently expecting `s.hosts` to be static, I am going to update with your suggestion and always store the next index to use and perform the mod before storing. This way the index we use is just whatever the current value is.

witchcraft/internal/tcpjson/envelope.go
tabboud: This is mirrored from the internal sls-spec, which has not been migrated to witchcraft-api yet.
I updated this to `LogEnvelopeV1` instead of SLS since we're re-defining it internally.

witchcraft/witchcraft.go
tabboud: Yep that is the correct observation. The `initLoggers` call only operates over the logger objects, so it's safe to re-run this again given we shouldn't have to worry about concurrent access or anything getting clobbered. However, you did point out a bug that I expanded on below.

witchcraft/witchcraft.go
tabboud: Good catch! The `s.initRuntimeConfig` does need the loggers on the context, so I moved it back to where it was. I also looked into what `s.withLoggers` is doing and it appears safe to re-run after we re-initialize the loggers for the TCP writer, so I added an additional call below which will overwrite the existing logger on the context.

witchcraft/internal/tcpjson/tcp_logger.go
tabboud: unfortunately not because `safejson.Marshal` uses `json.Encoder` under the hood and then strips off the `\n` char that we need to keep. I have since replaced this entirely as well.

tabboud:
@nmiyake I believe I have addressed all of the comments and this should be ready for another round of review.

To address your 2 bullet points
- I went ahead and opened up #252 which adds in the `service-discovery` config as a pre-commit. I will remove it from here once we merge that later today.
- Yes this is correct. We expect that the payload is JSON, but if it is not we will still wrap it in the envelope.

I added in a benchmark for the serializers to see what the result would be between json.Marshal, zerolog, or a manual formatting. Here are the results:

```
pkg: github.com/palantir/witchcraft-go-server/v2/witchcraft/internal/tcpjson
BenchmarkEnvelopeSerializer
BenchmarkEnvelopeSerializer/zerolog
BenchmarkEnvelopeSerializer/zerolog-16         	 2279970	       530 ns/op	    1505 B/op	       5 allocs/op
BenchmarkEnvelopeSerializer/JSON-Encoder
BenchmarkEnvelopeSerializer/JSON-Encoder-16    	  778161	      1609 ns/op	     688 B/op	       3 allocs/op
BenchmarkEnvelopeSerializer/JSON-Marshaler
BenchmarkEnvelopeSerializer/JSON-Marshaler-16  	  698350	      1523 ns/op	     640 B/op	       2 allocs/op
BenchmarkEnvelopeSerializer/manual
BenchmarkEnvelopeSerializer/manual-16          	 6416482	       175 ns/op	     768 B/op	       1 allocs/op
```

Overall the manual injection is approximately memory and CPU efficient than all other approaches, with the side affect of how hacky it is. The zerolog serializer (as it stands right now) is much more CPU efficient, but does allocate a bit more than regular json marshaling, which is interesting. In the end I opted for zerolog, since it seems like a nice middle ground between code complexity and performance. I currently use a bytes buffer for each serialization, where we could use a buffer pool to save some cost, but deferred that for now.

Also, another interesting thing we could do is write directly to the `net.Conn`, since it supports concurrent writes by default, rather than use a byte buffer. For each zerolog invocation we update the output, which can be any io.Writer, to be the local buffer but we could pass it the `net.Conn` and let it handle the writing. The challenge here becomes how to handle the transient errors and close the `net.Conn` appropriately.

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,83 @@
...
+	"net/url"
+	"sync/atomic"
+
+	werror "github.com/palantir/witchcraft-go-error"
+)
+
+// ConnProvider defines the behavior to retrieve an established net.Conn.
+type ConnProvider interface {
+	// GetConn returns a net.Conn or an error if there is no connection established.
+	// It is the callers responsibility to close the returned net.Conn and
```
nmiyake: typo: should be "caller's"

witchcraft/internal/tcpjson/conn_provider.go
```diff
@@ -0,0 +1,83 @@
...
+// tcpConnProvider implements a ConnProvider that will round-robin TCP connections to a specific set of hosts.
+type tcpConnProvider struct {
+	// nextHostIdx contains the index of the next host to connect to from the hosts slice below.
+	// The index will be reset to 0 when len(hosts) is reached to facilitate round-robin connections.
+	nextHostIdx int32
+	hosts       []string
+	tlsConfig   *tls.Config
+}
+
+func NewTCPConnProvider(uris []string, tlsCfg *tls.Config) (ConnProvider, error) {
```
nmiyake, line 55: Add documentation for function that describes that:
* Returns a new `ConnProvider`
* Provided set of uris must have at least one element and must be able to be parsed as URIs
* If config is `nil`, then default is used

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -0,0 +1,163 @@
...
+	envelopeSerializer envelopeSerializerFunc
+
+	// closedChan is used to signal that the writer is shutting down
+	closedChan chan struct{}
+
+	mu   sync.RWMutex // guards conn below
+	conn net.Conn
+}
+
+func NewTCPWriter(metadata LogEnvelopeMetadata, provider ConnProvider) *TCPWriter {
```
nmiyake, line 49: Comment please

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -0,0 +1,163 @@
...
+func newTCPWriterInternal(provider ConnProvider, serializerFunc envelopeSerializerFunc) *TCPWriter {
+	return &TCPWriter{
+		envelopeSerializer: serializerFunc,
+		provider:           provider,
+		closedChan:         make(chan struct{}),
+		conn:               nil,
+	}
+}
+
+func (d *TCPWriter) Write(p []byte) (int, error) {
```
nmiyake, line 66: Think comment would be helpful for this as well: describe that input is wrapped, written to connection, behavior if there is a connection error etc.

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Close will close any existing client connections and shuts down the writer from any future writes.
+func (d *TCPWriter) Close() error {
+	close(d.closedChan)
+	return d.closeConn()
+}
+
+func zerologSerializer(metadata LogEnvelopeMetadata) envelopeSerializerFunc {
+	// create a new top level logger with no a scratch output since each
```
nmiyake: nit: grammar

```
// create a new top-level logger that writes to ioutil.Discard since each
// serialization will write to its own local buffer instead of a single writer
```

witchcraft/loggers.go
```diff
@@ -41,7 +41,7 @@ const (
 // initLoggers initializes the Server loggers with instrumented loggers that record metrics in the given registry.
 // If useConsoleLog is true, then all loggers log to stdout.
 // The provided logLevel is used when initializing the service logs only.
-func (s *Server) initLoggers(useConsoleLog bool, logLevel wlog.LogLevel, registry metrics.Registry) {
+func (s *Server) initLoggers(useConsoleLog bool, logLevel wlog.LogLevel, registry metrics.Registry, tcpWriter io.Writer) {
```
nmiyake, line 45: Update function documentation to describe `tcpWriter` param

nmiyake (changes requested):
Thanks -- just some minor feedback in terms of content, and as discussed will wait for config PR to merge before this one.

witchcraft/internal/tcpjson/tcp_logger.go
tabboud: I had a comment on the exported type, so moved that one down here instead.

witchcraft/internal/tcpjson/tcp_logger.go
tabboud: 🤦‍♂️ fixed

witchcraft/internal/tcpjson/envelope.go
```diff
@@ -20,8 +20,8 @@ import (
 )
 
 type LogEnvelopeV1 struct {
-	Metadata LogEnvelopeMetadata `json:",inline"`
-	Payload  json.RawMessage     `json:"payload"`
+	LogEnvelopeMetadata `json:",inline"`
```
nmiyake: I think the `json":,inline"` doesn't actually do anything, so might as well remove

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -88,6 +95,14 @@ func (d *TCPWriter) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
+// trimNewLine will remove a trailing new line character and return a new byte slice.
+func trimNewLine(b []byte) []byte {
```
nmiyake: The https://golang.org/pkg/bytes/#TrimSuffix function basically does this -- can do `bytes.TrimSuffix(b, []byte("\n"))` in place of calling this function.

nmiyake (changes requested):
Minor feedback

witchcraft/internal/tcpjson/envelope.go
tabboud: I saw the issue about that actually which is what made me embed the type, just forgot to remove the tag. thanks!

witchcraft/internal/tcpjson/tcp_logger.go
tabboud: I intentionally wrote this to avoid allocating yet another `[]byte`, but in the grand scheme of things feels like an over optimization. Went ahead and changed it to the stdlib function.

nmiyake: By allocation, do you mean the `[]byte("\n")` line (since I don't believe the implementation of the function performs any allocations)? If we think that's an issue, then fine with going with this function, although would explain that in a comment. If the allocation is the only concern we could also have a package-level `var newlineBytes = []byte("\n")` that is allocated once and used in the trim operation

tabboud: Yep I was referring to the `[]byte("\n")`. I ran a quick benchmark to check my hypothesis and it turns out inlining the `[]byte("\n")` does not actually allocate, so no need for the local var.

bmoylan:
___
*[witchcraft/internal/tcpjson/conn_provider.go, line 81 at r9](https://reviewable.io/reviews/palantir/witchcraft-go-server/251#-MUVB4Zl-RqBeEMmIWof:-MUVB4Zl-RqBeEMmIWog:b-bziv8p) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5eb72d299b777e5f42e481f018b65527580fe0d8/witchcraft/internal/tcpjson/conn_provider.go#L81)):*
> ```Go
> 	atomic.CompareAndSwapInt32(&s.nextHostIdx, hostIdx, int32(nextHostIdx))
> 
> 	tlsConn, err := tls.Dial("tcp", s.hosts[hostIdx], s.tlsConfig)
> ```

we probably want to set some timeout on the Dial so it can't just hang forever. This requires constructing a dialer struct like
```go
tlsDialer := &tls.Dialer{NetDialer: &net.Dialer{Timeout: 10*time.Second}, Config: tlsConfig}
```

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/251)* status: 0 of 49 files reviewed, 18 unresolved discussions (waiting on @bmoylan, @nmiyake, and @tabboud)

___
*[witchcraft/internal/tcpjson/conn_provider.go, line 81 at r9](https://reviewable.io/reviews/palantir/witchcraft-go-server/251#-MUVB4Zl-RqBeEMmIWof:-MUoIkEK8RL7VoNcB7h-:b6anvs8) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/5eb72d299b777e5f42e481f018b65527580fe0d8/witchcraft/internal/tcpjson/conn_provider.go#L81)):*

we probably want to set some timeout on the Dial so it can't just hang forever. This requires constructing a dialer struct like
```go
tlsDialer := &tls.Dialer{NetDialer: &net.Dialer{Timeout: 10*time.Second}, Config: tlsConfig}
```
</blockquote></details>

Updated to use `tls.DialWithDialer` since `tls.Dialer` was not added until go1.15 (https://golang.org/pkg/crypto/tls/#Dialer). Once we update and vet the circle images to bump from go1.14 as the min, then we can update this as well

nmiyake (approved):

tabboud:
@bmoylan Did you want to take another pass before merging? 
I am going to make a follow up PR with a health check that monitors the active connection for the TCP writer (i.e. checking if tcpWriter.conn != nil).

bmoylan (approved):
> Reviewed 2 of 10 files at r1, 41 of 43 files at r3, 2 of 6 files at r4, 1 of 2 files at r6, 1 of 2 files at r7, 3 of 3 files at r8, 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/251)* status: all files reviewed, 17 unresolved discussions (waiting on @nmiyake and @tabboud)

bmoylan:
Let's do it!

PR #253, opened 2021-03-02
Collect all internal and user defined health check sources

tabboud:
## Before this PR
There was special-case handling for the config HealthCheckSource, which made it difficult to add any additional health check sources.

## After this PR
==COMMIT_MSG==
Collect all internal and user-defined health check sources before adding the status routes
==COMMIT_MSG==

Now all internal health check sources (except for `s.stateManager` itself) will be added to the user-supplied health check sources. The motivation for this is because we are looking to add a new health check source for the TCPWriter, and it's not trivial to add currently. Once this is refactored we can just append to the `internalHealthCheckSources` and everything will "just work".

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/253)

### DIFF
--- a/witchcraft/server_routes.go
+++ b/witchcraft/server_routes.go
@@ -44,7 +44,7 @@ func (s *Server) initRouters(installCfg config.Install) (rRouter wrouter.Router,
 	return routerWithContextPath, mgmtRouterWithContextPath
 }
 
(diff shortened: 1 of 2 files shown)


nmiyake (approved):
Thanks -- refactor makes sense and explanation was very understandable.

PR #254, opened 2021-03-03
Add TCPWriter health check source

tabboud:
## Before this PR
No external health checks to monitor the status of TCP logging

## After this PR

==COMMIT_MSG==
Add TCPWriter health check source to monitor the TCP connection status
==COMMIT_MSG==

## Notes
This health check is a bit strange given it's entirely dependent on the amount of logging done by consumers. For example, if no logging is performed, then this will always be healthy even though there is no connection. Maybe it's better to use `Deferring` or `Repairing`? I am proxying the connection status by checking the cached `net.Conn` which should only ever be `nil` when no writes have taken place or a write failed. I am open to other ideas on how to monitor this and also the levels at which we report.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/254)

### DIFF
--- a/witchcraft/internal/tcpjson/tcp_logger.go
+++ b/witchcraft/internal/tcpjson/tcp_logger.go
@@ -16,18 +16,29 @@ package tcpjson
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"io/ioutil"
 	"net"
 	"sync"
 
 	werror "github.com/palantir/witchcraft-go-error"
+	"github.com/palantir/witchcraft-go-health/conjure/witchcraft/api/health"
+	"github.com/palantir/witchcraft-go-health/status"
 	"github.com/rs/zerolog"
 )
 
-var _ io.WriteCloser = (*TCPWriter)(nil)
+var (
+	_ io.WriteCloser           = (*TCPWriter)(nil)
+	_ status.HealthCheckSource = (*TCPWriter)(nil)
+)
 
-const errWriterClosed = "writer is closed"
+const (
+	errWriterClosed = "writer is closed"
+
+	// tcpWriterHealthCheckName is the name used for the external health check
+	tcpWriterHealthCheckName = "TCP_LOGGER_CONNECTION_STATUS"
+)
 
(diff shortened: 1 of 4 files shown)


witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -39,8 +50,11 @@ type TCPWriter struct {
 	// closedChan is used to signal that the writer is shutting down
 	closedChan chan struct{}
 
-	mu   sync.RWMutex // guards conn below
+	mu   sync.RWMutex // guards all fields below
 	conn net.Conn
+	// started will be set to true when the first connection has been established
```
nmiyake: nit: grammar

Suggest:

```
// started is set to true when the first connection has been established
```

witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -141,6 +157,34 @@ func (d *TCPWriter) Close() error {
 	return d.closeConn()
 }
 
+// HealthStatus implements the status.HealthCheckSource interface for the TCPWriter.
+// An ERROR health status will be returned if attempts have been made to establish a connection but none
```
nmiyake: nit: use "is" rather than "will be" in tense here and below

nmiyake (changes requested):
Very minor nits for comments, but content looks good!

tabboud:
Seems like I have a habit of writing "will be" 🤦‍♂️.

bmoylan:
I wonder if there's a more responsive way to indicate when there are problems writing to the connection. What would you think about something like af81fe8 which uses the windowed source to return when the most recent write failed?

tabboud:
@bmoylan yeah I like that much better since it's monitoring the success of each write instead of proxying the connection status and also handles the edge cases I pointed out.

Want to open a separate PR or I can adapt that model here.

bmoylan:
sure, I opened #255

tabboud:
Closing in favor of #255

PR #258, opened 2021-03-05
Only enable TCP logging if both log metadata and receiver are configured

tabboud:
## Before this PR
If the URIs for the log receiver were set then we would enable the TCP logging output. Some internal infrastructure does not allow service-discovery in configuration so the fields are hardcoded per environment, which makes it impossible to determine whether we enable or disable the TCP logger solely on the presence of the URIs.

## After this PR

==COMMIT_MSG==
Only enable TCP logging if both log metadata and receiver are configured
==COMMIT_MSG==

In order to enable the TCP logger both the log envelope metadata AND log receiver must be configured. 
There are a few cases that need to be handled based on the two fields:
1. metadata is not configured (or partially) and receiver is **NOT** configured.
	a. Noop and logging will continue as normal
2. metadata is not configured (or partially) and the receiver **IS** configured.
	a. We emit a WARN log line to be able to trace the fact that TCP logging is disabled
	b. This is the case of hardcoding receiver config
3. metadata is configured and receiver is **NOT** configured
	a. This is an unexpected state, since we want to guarantee that if env vars are set, then we should be TCP logging.
	b. We set the `TCP_LOGGER_CONNECTION_STATUS` to a permanent `WARN` state with info that it's not configured correctly.
4. both metadata and receiver are configured
	a. enable TCP logging as usual

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/258)

### DIFF
--- a/changelog/@unreleased/pr-258.v2.yml
+++ b/changelog/@unreleased/pr-258.v2.yml
@@ -1,5 +1,8 @@
 type: improvement
 improvement:
-  description: Only enable TCP logging if both log metadata and receiver are configured
+  description: |-
+    Only enable TCP logging if both log metadata and receiver are configured
+
+    TCP logging will now only be enabled if both envelope metadata and the TCP reciever are properly configured. If the TCP receiver is configured without the envelope metadata then a warning log line will be produced to indicate a possible mis-configuration, but will otherwise proceed as normal. Some environments use hard-coded config, where it may be expected that the TCP receiver is configured, but the envelope metadata is not. Conversely, if the envelope metadata is configured and the TCP receiver is not, then the `TCP_LOGGER_CONNECTION_STATUS` health check will be set to a permanent warning state to indicate that TCP logging is disabled, since the reciever must be mis-configured.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/258
--- a/changelog/@unreleased/pr-258.v2.yml
+++ b/changelog/@unreleased/pr-258.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Only enable TCP logging if both log metadata and receiver are configured
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/258
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -621,9 +621,17 @@ func (s *Server) Start() (rErr error) {
 	envelopeMetadata, err := tcpjson.GetEnvelopeMetadata()
 	if err != nil {
 		if len(receiverCfg.URIs) > 0 {
+			// Presence of TCP receiver config without envelope metadata may indicate a mis-configuration,
+			// but can also be expected in environments where the config is always hard-coded.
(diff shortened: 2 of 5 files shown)


changelog/@unreleased/pr-258.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Only enable TCP logging if both log metadata and receiver are configured
```
nmiyake: Can you expand this a bit to explain the different behaviors for the different combinations? The explanation in the PR was very helpful to understand. The changelog doesn't need to be quite as extensive, but still think it would help to have further details about the expected behavior/usage here.

witchcraft/witchcraft.go
```diff
@@ -615,10 +616,20 @@ func (s *Server) Start() (rErr error) {
...
 
-	// extract the TCP JSON receiver from the runtime config
-	servicesCfg := baseRefreshableRuntimeCfg.CurrentBaseRuntimeConfig().ServiceDiscovery
-	receiverCfg, err := servicesCfg.MustClientConfig("sls-log-tcp-json-receiver")
-	if err == nil && len(receiverCfg.URIs) > 0 {
+	// enable TCP logging if the envelope metadata and the TCP receiver are both configured
+	receiverCfg := baseRefreshableRuntimeCfg.CurrentBaseRuntimeConfig().ServiceDiscovery.ClientConfig("sls-log-tcp-json-receiver")
+	envelopeMetadata, err := tcpjson.GetEnvelopeMetadata()
+	if err != nil {
+		if len(receiverCfg.URIs) > 0 {
```
nmiyake, line 623: Can you add a comment after this block to explain why it's okay to proceed without warning/logging?

witchcraft/witchcraft.go
```diff
@@ -615,10 +616,20 @@ func (s *Server) Start() (rErr error) {
...
+			s.svcLogger.Warn("TCP logging will not be enabled since all environment variables are not set.", svc1log.Stacktrace(err))
+		}
+	} else if len(receiverCfg.URIs) <= 0 {
```
nmiyake: nit: can just be `== 0` since `len` cannot return negative

witchcraft/witchcraft.go
```diff
@@ -615,10 +616,20 @@ func (s *Server) Start() (rErr error) {
...
+		internalHealthCheckSources = append(internalHealthCheckSources,
```
nmiyake, line 635: Can you add comment that explains what behavior this enables?

nmiyake (changes requested):
Thanks for the PR and the detailed explanation.

Implementation looks good overall -- just some feedback on documentation for clarify.

tabboud:
@nmiyake Thanks for the documentation feedback. I was trying to strike a balance of verbosity and traceability if any issues were encountered, but I think this looks better to include the reasoning in line rather than have to find the PR for context.

nmiyake (approved):
Looks good -- thanks!

PR #259, opened 2021-03-09
Refactor RouteParam interface to use builder pattern

dtrejod:
## Before this PR
Whenever adding a new field as a `RouteParam`, it required that all types satisfy an interface which returned a no-op except for the one method that was applicable for said field. 

## After this PR
We use a builder type to construct a RouteParam where all options are added to the routeParamBuilder by using an `apply()` method. 
==COMMIT_MSG==
Refactor RouteParam interface to use builder pattern. Now all RouteParam options are applied to an underlying routeParamBuilder at instantiation.
==COMMIT_MSG==

## Possible downsides?
None. All exported method signatures were kept consistent so that this will be a non-breaking change.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/259)

### DIFF
--- a/wrouter/route_params.go
+++ b/wrouter/route_params.go
@@ -18,120 +18,98 @@ import (
 	"github.com/palantir/pkg/metrics"
 )
 
+type routeParamBuilder struct {
+	paramPerms RouteParamPerms
+	metricTags metrics.Tags
+}
+
 type RouteParam interface {
-	perms() RouteParamPerms
-	metricTags() metrics.Tags
+	apply(*routeParamBuilder) error
 }
 
-type requestParamPermsFunc func() RouteParamPerms
+type routeParamFunc func(*routeParamBuilder) error
 
-func (f requestParamPermsFunc) perms() RouteParamPerms {
-	return f()
-}
-
-func (f requestParamPermsFunc) metricTags() metrics.Tags {
-	return nil
+func (f routeParamFunc) apply(b *routeParamBuilder) error {
+	return f(b)
 }
 
 func RouteParamPermsParam(perms RouteParamPerms) RouteParam {
-	return requestParamPermsFunc(func() RouteParamPerms {
-		return perms
+	return routeParamFunc(func(b *routeParamBuilder) error {
+		var pathParamPerms []ParamPerms
+		var queryParamPerms []ParamPerms
+		var headerParamPerms []ParamPerms
+		if b.paramPerms != nil {
+			pathParamPerms = append(pathParamPerms, b.paramPerms.PathParamPerms())
+			queryParamPerms = append(queryParamPerms, b.paramPerms.QueryParamPerms())
+			headerParamPerms = append(headerParamPerms, b.paramPerms.HeaderParamPerms())
+		}
+
+		pathParamPerms = append(pathParamPerms, perms.PathParamPerms())
+		queryParamPerms = append(queryParamPerms, perms.QueryParamPerms())
+		headerParamPerms = append(headerParamPerms, perms.HeaderParamPerms())
+
+		b.paramPerms = &requestParamPermsImpl{
+			pathParamPerms:   NewCombinedParamPerms(pathParamPerms...),
+			queryParamPerms:  NewCombinedParamPerms(queryParamPerms...),
+			headerParamPerms: NewCombinedParamPerms(headerParamPerms...),
+		}
+		return nil
 	})
 }
 
 func SafePathParams(safeParams ...string) RouteParam {
-	return requestParamPermsFunc(func() RouteParamPerms {
-		return &requestParamPermsImpl{
-			pathParamPerms: newSafeParamPerms(safeParams...),
-		}
+	return RouteParamPermsParam(&requestParamPermsImpl{
+		pathParamPerms: newSafeParamPerms(safeParams...),
 	})
 }
 
 func ForbiddenPathParams(forbiddenParams ...string) RouteParam {
-	return requestParamPermsFunc(func() RouteParamPerms {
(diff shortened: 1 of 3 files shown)


andybradshaw (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/259)* status: all files reviewed, 1 unresolved discussion (waiting on @bmoylan, @dtrejod, and @nmiyake)

___
*[wrouter/route_params.go, line 27 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/259#-MVN1p-66Nwej4d50oaX:-MVN1p-66Nwej4d50oaY:b-dxmd95) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/592d9e3dd2b3070b0e5ee96ca127b872429a7619/wrouter/route_params.go#L27)):*
> ```Go
> 
> type RouteParam interface {
> 	apply(*routeParamBuilder) error
> ```

Are there implementations of `RouteParam` that we expect to not interact with the `RequestVals` type? Would imagine something like the following interface better describes the intent:

```golang
type RouteParam interface {
    apply(*RequestVals) error
}
```

wrouter/route_params.go
```diff
@@ -18,120 +18,98 @@ import (
...
+		b.metricTags = tags
+		return nil
 	})
 }
 
-func toMetricTags(params []RouteParam) metrics.Tags {
-	var tags metrics.Tags
-	for _, param := range params {
-		tags = append(tags, param.metricTags()...)
+func toRequestParamPerms(b *routeParamBuilder) RouteParamPerms {
```
nmiyake: Since `toRequestParamPerms` and `toMetricTags` now both operate on `*routeParamBuilder`, let's have these functions be functions on the `*routeParamBuilder` type and move their definition next to the definition for the builder type.

wrouter/route_params.go
```diff
@@ -18,120 +18,98 @@ import (
...
+	}
+	return &requestParamPermsImpl{}
+}
+
+func toMetricTags(b *routeParamBuilder) metrics.Tags {
+	if b.metricTags != nil {
+		return b.metricTags
 	}
-	return tags
+	return metrics.Tags{}
```
nmiyake: Can we keep the previous structure but just replace the `for` loop with the `if` check? That will ensure that if `b.metricTags == nil`, this function still returns a `nil` `metrics.Tag` as before

wrouter/route_params.go
```diff
@@ -18,120 +18,98 @@ import (
...
+		var pathParamPerms []ParamPerms
+		var queryParamPerms []ParamPerms
+		var headerParamPerms []ParamPerms
+		if b.paramPerms != nil {
+			pathParamPerms = append(pathParamPerms, b.paramPerms.PathParamPerms())
+			queryParamPerms = append(queryParamPerms, b.paramPerms.QueryParamPerms())
+			headerParamPerms = append(headerParamPerms, b.paramPerms.HeaderParamPerms())
+		}
+
+		pathParamPerms = append(pathParamPerms, perms.PathParamPerms())
```
nmiyake, line 62: Why are the new params prepended rather than appended?

Think it would be more natural to initialize the perms to the current values and then append:

```go
pathParamPerms := perms.PathParamPerms()
queryParamPerms := perms.QueryParamPerms()
headerParamPerms := perms.HeaderParamPerms()
if b.paramPerms != nil {
...
}
```

nmiyake (changes requested):
Thanks! Some feedback.

wrouter/route_params.go
dtrejod: > Think it would be more natural to initialize the perms to the current values and then append:

Think the ordering I have is correct for this behavior. `b.paramPerms` are the existing params and the passed in `perms` argument are the new params we are appending.

wrouter/route_params.go
dtrejod: Done

wrouter/route_params.go
dtrejod: done

wrouter/route_params.go
nmiyake: Ah yes, you're right -- thanks!

nmiyake (approved):
Thanks, this looks good to me!

dtrejod (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/259)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @andybradshaw, @bmoylan, @dtrejod, and @nmiyake)

___
*[wrouter/route_params.go, line 27 at r1](https://reviewable.io/reviews/palantir/witchcraft-go-server/259#-MVN1p-66Nwej4d50oaX:-MVNCR7mDPQGl4w1xCJo:b-5sjhx2) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/592d9e3dd2b3070b0e5ee96ca127b872429a7619/wrouter/route_params.go#L27)):*
<details><summary><i>Previously, andybradshaw (Andy Bradshaw) wrote…</i></summary><blockquote>

Are there implementations of `RouteParam` that we expect to not interact with the `RequestVals` type? Would imagine something like the following interface better describe the intent?

```golang
type RouteParam interface {
    apply(*RequestVals) error
}
```
</blockquote></details>

I'd rather we not do this since we don't want to imply configurability of all RequestVals fields. The builder isolates the fields that actually influence RequestVals used by the handler here:

https://github.com/palantir/witchcraft-go-server/blob/develop/wrouter/router_root.go#L139-L144

andybradshaw (approved):

bmoylan (approved):

PR #262, opened 2021-03-11
feature: TCP writer uses a buffered channel to queue messages

bmoylan:
In the case that the TCP logger connection is slow or failing, we do not want to block the primary process from continuing. We may consider using this for the file based logger(s) as well, though disk IO speed has never been a problem.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/262)

### DIFF
--- a/changelog/@unreleased/pr-262.v2.yml
+++ b/changelog/@unreleased/pr-262.v2.yml
@@ -2,7 +2,5 @@ type: feature
 feature:
   description: |-
     TCP writer uses a buffered channel to queue messages
-
-    In the case that the TCP logger connection is slow or failing, we do not want to block the primary process from continuing. We may consider using this for the file based logger(s) as well, though disk IO speed has never been a problem.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/262
--- a/changelog/@unreleased/pr-262.v2.yml
+++ b/changelog/@unreleased/pr-262.v2.yml
@@ -1,6 +1,6 @@
 type: feature
 feature:
   description: |-
-    TCP writer uses a buffered channel to queue messages
+    TCP writer uses a buffered channel to queue messages. In the case that the TCP logger connection is slow or failing, we do not want to block the primary process from continuing.
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/262
--- a/changelog/@unreleased/pr-262.v2.yml
+++ b/changelog/@unreleased/pr-262.v2.yml
@@ -1,0 +1,8 @@
+type: feature
+feature:
+  description: |-
+    TCP writer uses a buffered channel to queue messages
+
+    In the case that the TCP logger connection is slow or failing, we do not want to block the primary process from continuing. We may consider using this for the file based logger(s) as well, though disk IO speed has never been a problem.
+  links:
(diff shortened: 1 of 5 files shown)


changelog/@unreleased/pr-262.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: feature
+feature:
+  description: |-
+    TCP writer uses a buffered channel to queue messages
```
nmiyake: If this was motivated by a particular issue (performance/configuration or otherwise), can you add a blurb on that here so it's clear why this would matter when reviewing the release notes?

witchcraft/internal/tcpjson/async_writer.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"io"
+	"log"
+
+	"github.com/palantir/pkg/metrics"
+	werror "github.com/palantir/witchcraft-go-error"
+)
+
+const (
+	asyncWriterBufferCap      = 1000
```
nmiyake: Is there any context around this value or is it arbitrary? Also is this bytes or number of messages or something else?

witchcraft/internal/tcpjson/async_writer.go
```diff
@@ -0,0 +1,64 @@
...
+	asyncWriterBufferLenGauge = "sls.logging.queued"
+)
+
+type asyncWriter struct {
+	buffer chan []byte
+	output io.Writer
+	stop   chan struct{}
+}
+
+func StartAsyncWriter(output io.Writer, registry metrics.Registry) io.WriteCloser {
```
nmiyake, line 40: Can you add a blurb about what this does and contract around closing etc.?

nmiyake (changes requested):
Minor feedback/clarification around documentation

bmoylan:
Thanks @nmiyake, updated!

witchcraft/loggers.go
```diff
@@ -63,7 +64,8 @@ func (s *Server) initLoggers(useConsoleLog bool, logLevel wlog.LogLevel, registr
 	logWriterFn := func(slsFilename string) io.Writer {
 		internalWriter := newDefaultLogOutputWriter(slsFilename, useConsoleLog, loggerStdoutWriter)
 		if tcpWriter != nil {
-			internalWriter = io.MultiWriter(internalWriter, tcpWriter)
+			asyncTCPWriter := tcpjson.StartAsyncWriter(tcpWriter, registry)
```
nmiyake: It doesn't look like this writer is ever closed -- what's the thinking behind how the lifecycle should be managed here?

I think this happens to work out right now because, although `initLoggers` is called multiple times on initialization, `tcpWriter` is only non-nil on one of these invocations. However, if it's the case that calling `initLoggers` with a non-nil `tcpWriter` will always start a new goroutine with no way to close/stop it, that seems like a bit of a concern.

nmiyake (changes requested):
Thanks! All of the minor things have been addressed, but in looking closer at the implementation I have a question/concern about how closing the TCP writer is handled.

nmiyake (approved):

PR #264, opened 2021-03-15
Fix async writer to no longer retain the provided byte slice

tabboud:
## Before this PR
The asyncWriter was storing the provided byte slice in the buffer channel which conflicts with the `io.Writer` contract (https://pkg.go.dev/io#Writer), which states to never retain the bytes.

## After this PR

==COMMIT_MSG==
Fix the async writer to copy the provided byte slice rather than retain the original in the buffer channel, which caused the the output to be mangled and non-deterministic
==COMMIT_MSG==

## Possible downsides?
- There will be more memory overhead and GC pressure since we're copying the byte slice on every write.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/264)

### DIFF
--- a/witchcraft/internal/tcpjson/async_writer_test.go
+++ b/witchcraft/internal/tcpjson/async_writer_test.go
@@ -16,11 +16,17 @@ package tcpjson
 
 import (
 	"bytes"
+	"context"
+	"encoding/json"
 	"strconv"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/palantir/pkg/metrics"
+	"github.com/palantir/witchcraft-go-logging/conjure/witchcraft/api/logging"
+	"github.com/palantir/witchcraft-go-logging/wlog"
+	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


witchcraft/internal/tcpjson/tcp_logger_test.go
```diff
@@ -188,15 +189,17 @@ func getEnvelopeBytes(t *testing.T, payload []byte) []byte {
 // bytes buffer instead of to the net.Conn.
 type bufferedConnProvider struct {
 	net.Conn
-	err    error
-	buffer bytes.Buffer
+	err        error
+	buffer     bytes.Buffer
+	writeCount int32
```
bmoylan, line 196: Can you add a comment that this is only used in testing?

tabboud: This conn provider can only be used from `_test.go` files, but I added the comment anyways.

witchcraft/internal/tcpjson/async_writer_test.go
```diff
@@ -49,3 +55,68 @@ func TestAsyncWriter(t *testing.T) {
 		require.EqualError(t, err, "write to closed asyncWriter")
 	})
 }
+
+// TestAsyncWriteWithSvc1log verifies the that svc1log lines are properly written to the output
+// when using the async writer. This also ensures the original input bytes are added to the buffered channel
+// correctly and the underlying byte slice is not stored directly which would cause the output to be malformed.
+func TestAsyncWriteWithSvc1log(t *testing.T) {
```
tabboud, line 62: This test fails on develop as well

bmoylan (approved):

witchcraft/internal/tcpjson/tcp_logger_test.go
bmoylan: oh whoops! I didn't realize this was in a test file. My bad

PR #267, opened 2021-03-16
TCPWriter: Close connections on network timeout errors

tabboud:
## Before this PR
Connections could get stuck in a "connection timed out" state and the underlying net.Conn would never be reset causing logs to be dropped.

According to the docs for `tls.Conn.SetDeadline()` (https://pkg.go.dev/crypto/tls#Conn.SetDeadline), after a write has timed out once, the TLS state is corrupt and thus the same error is returned. My hypothesis is that once we hit the timeout, we got stuck in this state since we never closed the connection.

## After this PR

==COMMIT_MSG==
Close TCP connections if there are network timeouts
==COMMIT_MSG==

## Notes
I wasn't able to repro the exact issues we were seeing below, since the `connection timed out` error comes from the [ETIMEDOUT syscall](https://github.com/golang/sys/blob/77fc1eacc6aaf85d786cc0e8343c191ce26ef821/unix/zerrors_linux_amd64.go#L738) on Linux.
```
2021/03/16 18:38:25 write failed: write tcp <src-ip>:<src-port> -> <dst-ip>:<dst-port>: write: connection timed out
```
However when digging into `syscall.ETIMEDOUT`, I realized it happens to implement the `net.Error` interface and returns true for [`Timeout()`](https://github.com/golang/go/blob/7a1e963058460d4136603b86386f2bae6fe0d5f2/src/syscall/syscall_unix.go#L145) and ALSO returns true for [`Temporary()`](https://github.com/golang/go/blob/7a1e963058460d4136603b86386f2bae6fe0d5f2/src/syscall/syscall_unix.go#L141) if there was a timeout!

Relevant issue: https://github.com/golang/go/issues/31449

## Possible downsides?
- Not directly caused by this change, but we upon any timeout, we will still drop the log line.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/267)

### DIFF
--- a/witchcraft/internal/tcpjson/tcp_logger.go
+++ b/witchcraft/internal/tcpjson/tcp_logger.go
@@ -106,8 +106,8 @@ func (d *TCPWriter) Write(p []byte) (n int, err error) {
 		n, err := conn.Write(envelope[total:])
 		total += n
 		if err != nil {
-			if nerr, ok := err.(net.Error); !ok || !nerr.Temporary() {
-				// permanent error so close the connection
+			if nerr, ok := err.(net.Error); !ok || nerr.Timeout() || !nerr.Temporary() {
+				// permanent error or timeout so close the connection
 				_ = d.closeConn()
 				return total, err
 			}
--- a/witchcraft/internal/tcpjson/tcp_logger.go
+++ b/witchcraft/internal/tcpjson/tcp_logger.go
@@ -106,7 +106,7 @@ func (d *TCPWriter) Write(p []byte) (n int, err error) {
 		n, err := conn.Write(envelope[total:])
 		total += n
 		if err != nil {
-			if nerr, ok := err.(net.Error); !(ok && (nerr.Temporary() || nerr.Timeout())) {
+			if nerr, ok := err.(net.Error); !ok || !nerr.Temporary() {
 				// permanent error so close the connection
 				_ = d.closeConn()
 				return total, err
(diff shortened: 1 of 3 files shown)


witchcraft/internal/tcpjson/tcp_logger.go
```diff
@@ -106,7 +106,7 @@ func (d *TCPWriter) Write(p []byte) (n int, err error) {
 		n, err := conn.Write(envelope[total:])
 		total += n
 		if err != nil {
-			if nerr, ok := err.(net.Error); !(ok && (nerr.Temporary() || nerr.Timeout())) {
+			if nerr, ok := err.(net.Error); !ok || !nerr.Temporary() {
```
tabboud: I am slightly nervous that a timeout error could also be considered temporary, but looking at the [tls.Conn](https://go.googlesource.com/go/+/go1.16/src/crypto/tls/conn.go#178) code it doesn't seem like that is the case. Mind double checking this logic?

tabboud: I just went ahead and added the explicit check for a timeout as well to be safe.

bmoylan (approved):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/267)* status: all files reviewed, 1 unresolved discussion (waiting on @tabboud)

PR #274, opened 2021-03-24
Log metrics that go from 0 to 1

nmiyake:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/274)


witchcraft/server_metrics.go
```diff
@@ -78,6 +80,9 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
 		metricTypeValuesBlacklist = defaultMetricTypeValuesBlacklist()
 	}
 
+	// records metrics that have been logged. Key consists of "{metricID}-{metricType}-{tags}".
+	seenMetrics := make(map[string]struct{})
```
tabboud: Do we need to guard against concurrent writes/reads?

witchcraft/server_metrics.go
```diff
@@ -101,7 +101,32 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
 		// most up-to-date metric logger is used (s.metricLogger may be updated during initialization).
 		// s.metricLogger is not guaranteed to be non-nil at this point.
 		if s.metricLogger != nil {
-			s.metricLogger.Metric(metricID, metricType, metric1log.Values(valuesToUse), metric1log.Tags(tags.ToMap()))
+			metricTagsParam := metric1log.Tags(tags.ToMap())
+
+			if metricType == "meter" || metricType == "timer" || metricType == "histogram" {
+				mapKey := fmt.Sprintf("%s:%s:%v", metricID, metricType, tags)
```
tabboud: Is it possible that 2 metrics could have equivalent `metricID`, `metricType`, and `tags`, but the tags be in different orders?

tabboud (commented):
I have not looked at the test yet, but the main logic looks pretty sane to me. Left a couple questions below.

nmiyake:
Updated PR to address feedback -- this is now no longer a draft and is ready for review.

changelog/@unreleased/pr-274.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: "Log zero values for certain metric logger entry types\n    \nWhen logging \"meter\", \"timer\", or \"histogram\" metric logger entries,\nif an entry has not already been logged for a metric with a specific\nset of (id, type, tags), then emit a metric log entry for that metric\nwith zero values first."
```
tabboud, line 3: I assume this will render correctly with all the `\n` chars?

witchcraft/server_metrics.go
```diff
@@ -57,6 +62,14 @@ var (
 	initTime = time.Now()
 )
 
+// seenMetricsSet pairs a set with a lock that can be used to control access to the set.
+type seenMetricsSet struct {
+	// records metrics that have been logged
+	seenSet map[string]struct{}
+	// protects access to seenSet
+	lock sync.RWMutex
```
tabboud: nit: I generally like to have lock declarations sit on top of what they guard (otherwise referred to as a "mutex hat")

witchcraft/server_metrics.go
```diff
@@ -118,6 +147,106 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
...
+	}
+	seenMetrics.lock.Unlock()
+	return zeroValuesLogged
+}
+
+// tagMapKey returns a string that provides a unique key for the provided metricID, metricType and tagsMap. The content
+// of the map is put into a slice where each element is a string of the form "{key}:{value}" and the slice is then
+// sorted. The tags map (rather than slice) is used to construct the key because the map is what the metric logger
+// ultimately emits.
+func tagMapKey(metricID, metricType string, tagsMap map[string]string) string {
```
tabboud, line 223: note: I am slightly nervous about the performance hit of this function, but the code is structured in a way where we can easily profile and replace if needed.

witchcraft/server_metrics.go
```diff
@@ -118,6 +147,106 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
...
+	// that it will return true vast majority of the time. However, want to ensure that only one zero-value log line is
+	// emitted, so check condition again while holding write lock.
+	if _, ok := seenMetrics.seenSet[mapKey]; !ok {
+		// metric not seen before: emit zero-value and record
+		metricLogger.Metric(metricID, metricType, metric1log.Values(zeroValuesToUse), metricTagsParam)
+		seenMetrics.seenSet[mapKey] = struct{}{}
+		zeroValuesLogged = zeroValuesToUse
+	}
+	seenMetrics.lock.Unlock()
+	return zeroValuesLogged
```
tabboud: nit: While the logic is correct, we could re-structure this to reduce indentation and keep the locking together.
For example
```go
seenMetrics.lock.Lock()
defer seenMetrics.lock.Unlock()

if _, metricSeen := seenMetrics.seenSet[mapKey]; metricSeen {
	return nil
}

// metric not seen before: emit zero-value and record
metricLogger.Metric(metricID, metricType, metric1log.Values(zeroValuesToUse), metricTagsParam)
seenMetrics.seenSet[mapKey] = struct{}{}
return zeroValuesToUse

```

Defer to you since both ways seem fine with me, just noting.

tabboud (commented):
This looks great now, thanks for adding this in!

All comments are non-blocking, but will wait to approve for your response first.

witchcraft/server_metrics.go
nmiyake: Sure, will do.

witchcraft/server_metrics.go
nmiyake: Yes, understood. However, not totally sure how to improve on this unless we assume that tag order will be stable for a given metric/we will treat different tag ordering differently (I don't think we can just print the map directly since map iteration order is random).

One thing to note is that the penalty/overhead for metric logging should be lower than for service logging -- this call is basically made for the number of matching metric types each "emit frequency" unit of time (unlike service logs, which are emitted on every call to the logger)

tabboud: Yeah, I am not sure how to optimize this without further changes and I believe it will be fine, but just wanted to flag in case we do see any problems. Good point on the `emit-frequency` which certainly amortizes the cost.

witchcraft/server_metrics.go
nmiyake: Thanks -- was mainly matching the style of the read lock, which was done on its own since we didn't want to defer in that case.

However, upon thinking about this further, I think that I'm going to update the implementation to just acquire a write lock. I believe that 99%+ of the time, this function will only be called in a single goroutine (the "emit" routine) -- the only instance where this isn't the case is when "emitFn" is called on shutdown. As such, I think it's cheaper to just acquire and release the lock once even if it's over a bigger section of logic -- the current optimization only helps when there may be more reads than writes from different goroutines, but since that isn't the case here I think the simplification will be worth it.

changelog/@unreleased/pr-274.v2.yml
nmiyake: I believe so (this was generated by the bot and I think it generally works)

nmiyake:
Thanks for the review -- addressed PR feedback

witchcraft/server_metrics.go
```diff
@@ -207,17 +210,10 @@ func logZeroValueMetric(
...
-		// metric not seen before: emit zero-value and record
-		metricLogger.Metric(metricID, metricType, metric1log.Values(zeroValuesToUse), metricTagsParam)
-		seenMetrics.seenSet[mapKey] = struct{}{}
-		zeroValuesLogged = zeroValuesToUse
-	}
-	seenMetrics.lock.Unlock()
+	// metric not seen before: emit zero-value and record
+	metricLogger.Metric(metricID, metricType, metric1log.Values(zeroValuesToUse), metricTagsParam)
+	seenMetrics.seenSet[mapKey] = struct{}{}
+	zeroValuesLogged = zeroValuesToUse
```
tabboud: This would probably get optimized away, but we can just return directly here.

witchcraft/server_metrics.go
tabboud: re style: I figured that was the case, but usually still fall back to the `defer` use case if possible later since the explicit locking usually only applies in one-off scenarios.

That was the reason I posed the question in the draft initially. It sounds like there will be very little lock contention (except for on shutdown), which as you eluded to makes more sense to just lock over all of the logic. I am in sync on this and it cleans up the code path nicely.

tabboud (approved):

PR #284, opened 2021-04-10
fix: Handle server shutdown before succesful start

bmoylan:
## Before this PR
If `ShutdownServer` was called before the server is actually started (for example, in the init func or a goroutine it spawns), we get a "server is not running" error and the startup continues as if it never happened.

## After this PR

==COMMIT_MSG==
Handle server shutdown before succesful start
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/284)

### DIFF
--- a/changelog/@unreleased/pr-284.v2.yml
+++ b/changelog/@unreleased/pr-284.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Handle server shutdown before succesful start
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/284
--- a/changelog/@unreleased/pr-284.v2.yml
+++ b/changelog/@unreleased/pr-284.v2.yml
@@ -1,5 +1,8 @@
 type: fix
 fix:
-  description: Handle server shutdown before succesful start
+  description: >
+    Handle race condition when `ShutdownServer` is called before successful start.
+    Previously, if the server is told to shut down before it has started, the shutdown request errors and is ignored,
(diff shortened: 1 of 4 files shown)


changelog/@unreleased/pr-284.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Handle server shutdown before succesful start
```
tabboud: Mind updating this to be a bit more descriptive of the bug and resolution, so that it's clear what this fixes by just reading release notes. Your "Before this PR" is a good starting point since it touches on the fact that there was a race condition and the server would remain started.

witchcraft/witchcraft.go
```diff
@@ -779,7 +780,9 @@ func (s *Server) Start() (rErr error) {
 		s.httpServer.SetKeepAlivesEnabled(false)
 	}
 
-	s.stateManager.setState(ServerRunning)
+	if !s.stateManager.compareAndSwapState(ServerInitializing, ServerRunning) {
+		return werror.ErrorWithContextParams(ctx, "Server was shut down before it could start")
```
tabboud: nit: lowercase `Server`

tabboud (changes requested):

changelog/@unreleased/pr-284.v2.yml
bmoylan: Agreed, updated!

tabboud (approved):
👍

PR #290, opened 2021-04-22
improvement: Drain async logger before exiting

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/290)


tabboud (approved):
👍

PR #295, opened 2021-05-12
Improvement: report unauthorized error as a health check

AdamDeHovitz:
## Before this PR
We believe we hit this case last week but it's difficult to confirm due to a lack of logging and the way an empty health status get's reported in autopilot. 

## After this PR

==COMMIT_MSG==
If user is unauthorized to view health checks return an unauthorized error health check rather than no health check.
==COMMIT_MSG==

## Possible downsides?
Possibly confusing to users/might interact poorly if existing monitoring/error catching depends on an empty health status object.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/295)

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -15,16 +15,15 @@
 package status
 
 import (
-	"github.com/palantir/witchcraft-go-health/sources"
 	"net/http"
 	"sync/atomic"
 
 	"github.com/palantir/conjure-go-runtime/v2/conjure-go-server/httpserver"
 	"github.com/palantir/pkg/refreshable"
 	"github.com/palantir/witchcraft-go-health/conjure/witchcraft/api/health"
+	"github.com/palantir/witchcraft-go-health/sources"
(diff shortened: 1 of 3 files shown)


status/status.go
```diff
@@ -66,7 +74,7 @@ func (h *healthHandlerImpl) computeNewHealthStatus(req *http.Request) (health.He
 	if sharedSecret := h.healthCheckSharedSecret.CurrentString(); sharedSecret != "" {
 		token, err := httpserver.ParseBearerTokenHeader(req)
 		if err != nil || sharedSecret != token {
-			return health.HealthStatus{}, http.StatusUnauthorized
+			return healthStatusUnauthorized, http.StatusUnauthorized
```
AdamDeHovitz, line 76: My preference was to log when we hit this case but there's no accessible ctx and the `loggingHealthStatusChangeHandler` used above derives the status code from the health status rather than from the status returned by `computeNewHealthStatus`

PR #320, opened 2021-08-30
Decrypt ECV configuration values in YAML nodes

### LINKED ISSUE #318: Encrypted values in configuration should be decoded while maintaining YAML structure
## What happened?
WGS crashed when attempting to start up using install configuration that contained the following, where the content of the encrypted value is a certificate with line breaks as part of the data:

```
client-ca-certs:
      test-ca: '${enc:encrypted_bytes}'
```

```
Failed to unmarshal install base configuration YAML
yaml: line 13: could not find expected ':'
github.com/palantir/witchcraft-go-server/v2/witchcraft.(*Server).initInstallConfig
	/go/src/.../vendor/github.com/palantir/witchcraft-go-server/v2/witchcraft/witchcraft.go:834
...
```

WGS should start and load the proper configuration

## Analysis
The issue here is that, currently, if the configuration contains any ECV values, all of those values are decrypted within the configuration bytes and then the resulting bytes are unmarshalled. This means that, if the decrypted values render in a manner that semantically change the meaning of the YAML (for example, by specifying key-value pairs), this impacts the manner in which the decryption occurs.

Instead, the ideal would be to read in the YAML nodes and then, for any text nodes that are encrypted, use the decrypted text as the value.

nmiyake:
Fixes #318

## After this PR

==COMMIT_MSG==
Updates the way in which encrypted configuraiton values are handled
in configuration (both install and runtime configuration). Previously,
if configuration contained encrypted configuration values, the values
were decrypted in the raw bytes and the result was then unmarshaled as
YAML. This meant that it was possible for the content of the encrypted
configuration value to influence the structure of the YAML document.

This behavior has been modified such that, if configuration contains
an encrypted configuration value, the configuration is interpreted as
YAML, the encrypted configuration values are decrypted and set as the
value on the original YAML node, and the result is then unmarshaled.
This means that it is no longer possible for the content of an encrypted
configuration value to modify the structure of the configuration YAML,
with the result being that there should be better support for values such
as multiline strings. This also makes the behavior more consistent with
other existing implementations.

Although using the previous behavior to modify the structure of configuration
YAML using encrypted values was not recommended, this is a behavioral change:
any products that depended on the previous behavior of encrypted configuration
values modifying YAML structure on decryption will no longer work. Decrypting
encrypted configuration values in YAML is now done by doing an unmarshal/marshal
round-trip using yaml.v3, so this may add some extra overhead and there is a
small possibility that it could cause behavioral issues.
==COMMIT_MSG==

## Possible downsides?
* Any products that depended on the previous behavior of ECV unmarshal providing literal values (or modified configuration to work around it) will no longer work (or will need to remove the workaround)
* Adds new YAML round-trip overhead for configuration with ECV values (previously, a simple find-replace was performed in raw bytes; now, raw bytes are read into yaml.v3 node, all ECV nodes are updated recursively, then node is rewritten as YAML bytes)
* If yaml.v3 marshal/unmarshal behavior is different in edge cases, this could cause problems. However, think this is mitigated somewhat by having yaml.v3 used only for unmarshal/marshal to raw yamlv3.Node and []byte (actual unmarshal into struct is still handled by previous logic using yaml.v2 library)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/320)


bmoylan (commented):
Overall this change makes sense, though I'm not sure how to handle/message the behavior breaks. Do we think an email would be enough?

> Reviewed 12 of 14 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/320#-:-MiNKtdO1XwKCaaiQK1v:bk3u2m0)* status: all files reviewed, 1 unresolved discussion (waiting on @Neverblade, @nmiyake, and @tabboud)

___
*[integration/config_test.go, line 82 at r3](https://reviewable.io/reviews/palantir/witchcraft-go-server/320#-MiNJsjK-0jCF1-heECF:-MiNJsjLBKJT5kdafYn3:b4ktmri) ([raw file](https://github.com/palantir/witchcraft-go-server/blob/320a7411fde777c3d3bc062c507b31781c6c2a10/integration/config_test.go#L82)):*
> ```Go
> 		},
> 		{
> 			Name:          "encrypted multi-line value is decrypted properly",
> ```

Some additional cases that would be good to test: 
* the scalar value has content outside the secret: `message: prefix ${enc:...} suffix` (including multilines)
* secret value is not a string -- previously, something like `message: ${enc:...}` replaced with `5` would result in `message: 5` (an integer) but I worry that this will produce `message: "5"` (a string)

nmiyake:
@bmoylan thanks for the input, very much appreciated.

1. With regards to the first case you outlined (`message: prefix ${enc:...} suffix`), this will fail with the current implementation, and is expected -- the contract for the updated implementation is that, if there's an ECV value in the value, it should be the only thing in the value. Perhaps this should be made clearer in the documentation/notes. Although this isn't strictly necessary (we could easily handle the case where we expand it instead), I think this makes the contract cleaner, and for what it's worth I believe this is what the Java implementation does (https://github.com/palantir/encrypted-config-value/blob/eb62721106e6996c49af42af69ce87109ccb9d62/encrypted-config-value-module/src/main/java/com/palantir/config/crypto/jackson/EncryptedConfigMapperUtils.java#L36, https://github.com/palantir/encrypted-config-value/blob/eb62721106e6996c49af42af69ce87109ccb9d62/encrypted-config-value-module/src/main/java/com/palantir/config/crypto/jackson/JsonNodeStringReplacer.java#L34).

2. Agree with the second point you outlined as well -- it is true that, previously, if the rendered value was interpreted as something other than a string in YAML (`5`, `["this", "is", "a", "list"]`), then it was possible for the render + decryption to handle this, but now this will no longer be the case and the value will be a string. Per the first point this is by design, but is a breaking change.

Basically, the contract is being changed from:

"ECV values render directly and the resulting configuration is decrypted as YAML"

to

"ECV values may only by used as the content of YAML string nodes and when they occur they must be the only value in that node. The configuration bytes are replaced with config bytes where encrypted ECV values nodes are decrypted and then the result is unmarshaled".

changelog/@unreleased/pr-320.v2.yml
```diff
@@ -0,0 +1,29 @@
+type: improvement
+improvement:
+  description: |-
+    Updates the way in which encrypted configuraiton values are handled
```
tabboud: typo on `configuration`

tabboud (commented):
Overall this looks great to me. Thanks for the thorough explanation of the problem/solution as it made it very simple to review.

I feel like email would be sufficient, but if we did want to be extremely cautious we could major rev as well but I don't think it's necessary in this instance.

nmiyake:
@bmoylan just wanted to verify, to the responses above make sense?

@Neverblade does this seem okay on your end as well? Is there a process that you want to go through to try to pre-vet existing usages of configuration to verify that this is okay, or are you comfortable with us making this upgrade and keeping an eye out on any issues that may come up after release and then responding to those as needed?

Neverblade:
Just so I understand, the only way this is considered a behavior change is if a product was using ECV configuration values to change the YAML structure, correct? From my perspective I think this is unsupported behavior and I don't think we need to take steps to try and catch products that are exploiting this behavior. Additionally, I don't think it'd be feasible to vet usages of config because we cannot see the values of these ECV configs without being on the environment itself.

As you mentioned there's also a potential risk of side effects from switching to yaml V3, but given we've successfully verified an instance of a product using this change live I don't think there's much else we can do on that front

nmiyake:
@Neverblade yes, that's correct. Agree that I think the scope is limited enough and the behavioral change is only in an area that we didn't officially support, so should be okay.

@bmoylan please let us know if you have any NACKs given the updated context -- otherwise, we'll move forward.

@tabboud this is ready for full review -- if you approve, then we can merge and release (I've tagged with autorelease and merge when ready).

changelog/@unreleased/pr-320.v2.yml
tabboud: Manually fixed in the changelog bot comment

tabboud (approved):
👍

PR #325, opened 2021-09-30
improvement: cache sample for logging.sls.length histogram

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/325)


tabboud (approved):
👍

PR #252, opened 2021-02-25
Add service discovery configuration

tabboud:
## Before this PR
No standardized way to add service discovery information to runtime config.

## After this PR
==COMMIT_MSG==
Add service discovery configuration for generating clients

Services can now declare all dependent client configurations in a single location and share common settings across all clients. Although the new service-discovery block is in the runtime configuration, it is not currently live-reloadable. The service-discovery block is not yet consumed from within witchcraft-go-server, but instead provides a means to standardize how and where clients are configured.
==COMMIT_MSG==

## Possible downsides?
- The CGR services config block is currently **not** live-reloadable (albeit within runtime config)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/252)

PR #327, opened 2021-10-04
feature: Add ServiceDiscovery to InitInfo for reloading HTTP clients

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/327)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.0.0
-	github.com/palantir/conjure-go-runtime/v2 v2.19.0
+	github.com/palantir/conjure-go-runtime/v2 v2.19.1-0.20211004160052-888f130ee444
```
bmoylan: Depends on https://github.com/palantir/conjure-go-runtime/pull/171

witchcraft/service_discovery.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/palantir/witchcraft-go-server/v2/config"
+)
+
+type ServiceDiscovery interface {
+	// New is an alias for httpclient.NewClientFromRefreshableConfig based on the 'service-discovery' block in runtime configuration.
+	New(ctx context.Context, serviceName string, additionalParams ...httpclient.ClientParam) (httpclient.Client, error)
+	NewHTTP(ctx context.Context, serviceName string, additionalParams ...httpclient.HTTPClientParam) (httpclient.RefreshableHTTPClient, error)
+}
+
+type ConfigurableServiceDiscovery interface {
```
bmoylan, line 36: I added all these options so that if products have client configuration in other places and can't just move to the default field because there are config overrides, we can merge the old overrides into the default config

integration/runtime_test.go
```diff
@@ -503,6 +507,174 @@ exclamations: 4
 	assert.Equal(t, validCfg2, currCfg)
 }
 
+func TestRuntimeReloadServiceDiscovery(t *testing.T) {
+	upstreamServer := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		assert.Equal(t, "example/1.2.3", req.Header.Get("User-Agent"), "expected user-agent based on config")
```
tabboud: nit: might as well add a `userAgent` const to the common_test.go as well and reference it here.

witchcraft/service_discovery_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/palantir/pkg/refreshable"
+	"github.com/palantir/witchcraft-go-server/v2/config"
+	"github.com/stretchr/testify/require"
+)
+
+func TestServiceDiscovery_RefreshableClientConfig(t *testing.T) {
+	startingConfig := httpclient.ServicesConfig{}
+	defaultRefreshable := refreshable.NewDefaultRefreshable(startingConfig)
+	refreshingConfig := config.NewRefreshingServicesConfig(defaultRefreshable)
+	discovery := newServiceDiscovery(config.Install{ProductName: "my-product", ProductVersion: "1.2.3"}, refreshingConfig)
```
tabboud: nit: use defined consts

witchcraft/witchcraft.go
```diff
@@ -633,7 +637,7 @@ func (s *Server) Start() (rErr error) {
 	internalHealthCheckSources := []healthstatus.HealthCheckSource{configReloadHealthCheckSource}
 
 	// enable TCP logging if the envelope metadata and the TCP receiver are both configured
-	receiverCfg := baseRefreshableRuntimeCfg.CurrentBaseRuntimeConfig().ServiceDiscovery.ClientConfig("sls-log-tcp-json-receiver")
+	receiverCfg := baseRefreshableRuntimeCfg.CurrentRuntime().ServiceDiscovery.ClientConfig("sls-log-tcp-json-receiver")
```
tabboud: nit: should we instead leverage the generated refreshables and grab this via
```go
baseRefreshableRuntimeCfg.ServiceDiscovery().CurrentServicesConfig().ClientConfig("sls-log-tcp-json-receiver")
```
I am personally indifferent, so no need to change but thought I would mention

witchcraft/witchcraft.go
```diff
@@ -705,10 +709,8 @@ func (s *Server) Start() (rErr error) {
 	}
 
 	// handle built-in runtime config changes
-	unsubscribe := baseRefreshableRuntimeCfg.Map(func(in interface{}) interface{} {
-		return in.(config.Runtime).LoggerConfig
-	}).Subscribe(func(in interface{}) {
-		if loggerCfg := in.(*config.LoggerConfig); loggerCfg != nil {
+	unsubscribe := baseRefreshableRuntimeCfg.LoggerConfig().SubscribeToLoggerConfigPtr(func(loggerCfg *config.LoggerConfig) {
```
tabboud, line 717: note: This is so much cleaner with the generated refreshables!

witchcraft/service_discovery.go
```diff
@@ -0,0 +1,147 @@
...
+	"sync"
+
+	"github.com/palantir/conjure-go-runtime/v2/conjure-go-client/httpclient"
+	"github.com/palantir/witchcraft-go-server/v2/config"
+)
+
+type ServiceDiscovery interface {
+	// New is an alias for httpclient.NewClientFromRefreshableConfig based on the 'service-discovery' block in runtime configuration.
+	New(ctx context.Context, serviceName string, additionalParams ...httpclient.ClientParam) (httpclient.Client, error)
+	NewHTTP(ctx context.Context, serviceName string, additionalParams ...httpclient.HTTPClientParam) (httpclient.RefreshableHTTPClient, error)
```
tabboud: What do you think about adding a `Client` suffix to these? When I first read `New` without the context, I assumed it was to construct a new service-discovery block, albeit that it returns a `httpclient.Client`.

witchcraft/service_discovery.go
```diff
@@ -0,0 +1,152 @@
...
+	NewHTTPClient(ctx context.Context, serviceName string, additionalParams ...httpclient.HTTPClientParam) (httpclient.RefreshableHTTPClient, error)
+}
+
+// ConfigurableServiceDiscovery is an extension to the ServiceDiscovery interface which allows for injecting external
+// configuration to be used when constructing clients. New usages should prefer the runtime.service-discovery block
+// but in some cases reading from deprecated config locations is necessary to avoid breaking changes when config files
+// are loosely coupled to code versions.
+type ConfigurableServiceDiscovery interface {
+	ServiceDiscovery
+	// SetRefreshableServicesConfig provides for a base RefreshableServicesConfig different from the default
```
tabboud: nit: maybe explicitly mention that this will override the default witchraft service-discovery location, so live reloaded clients will be constructed from here instead.

Also do you think it's worth exposing this option given we also expose the `WithServiceConfig`? Removing the knob to change the default location of service-discovery config may be impetus enough to get teams to switch over. Certainly happy to leave it in too, but I am not aware of any current use case for this option. I suppose the main downside would be where you have default config in the proper location, but another service config declared elsewhere that would be constructed with potentially different defaults. That may be reason enough to leave this in, but curious for your thoughts.

witchcraft/service_discovery.go
```diff
@@ -0,0 +1,152 @@
...
+	// runtime.service-discovery block.
+	SetRefreshableServicesConfig(cfg config.RefreshableServicesConfig)
+	// WithDefaultConfig adds the provided ClientConfig to a list of extra defaults to be merged together for the final
+	// client configurations returned by New and NewHTTP.
+	WithDefaultConfig(defaults httpclient.ClientConfig)
+	// WithServiceConfig allows for additional configuration from outside the default runtime.service-discovery block.
```
tabboud, line 41: Should we mention that the clients will still be constructed using defaults from the main service-discovery location regardless of where this client is declared.

witchcraft/service_discovery.go
```diff
@@ -0,0 +1,152 @@
...
+// configuration to be used when constructing clients. New usages should prefer the runtime.service-discovery block
+// but in some cases reading from deprecated config locations is necessary to avoid breaking changes when config files
+// are loosely coupled to code versions.
+type ConfigurableServiceDiscovery interface {
+	ServiceDiscovery
+	// SetRefreshableServicesConfig provides for a base RefreshableServicesConfig different from the default
+	// runtime.service-discovery block.
+	SetRefreshableServicesConfig(cfg config.RefreshableServicesConfig)
+	// WithDefaultConfig adds the provided ClientConfig to a list of extra defaults to be merged together for the final
+	// client configurations returned by New and NewHTTP.
```
tabboud: nit: `NewClient` and `NewHTTPClient`

tabboud (commented):
This looks great! Left a few comments on the godoc's but otherwise I feel comfortable moving forward with this.

witchcraft/service_discovery.go
bmoylan: So the idea here is that I have seen products add something like `runtime.clients` from before we introduced `runtime.service-discovery` in #252. The reason this one overwrites and throws away the default config is that we do not have a way of merging multiple refreshable configs together (the mapping and subscription semantics don't make it easy).

An alternative to this which might be better in the long run is exposing a `NewServiceDiscovery(RefreshableServicesConfig) ServiceDiscovery` constructor so they can still use our "factory" but we don't have to worry about coordinating the two refreshables.

witchcraft/service_discovery.go
bmoylan: Done.

witchcraft/service_discovery.go
bmoylan: Done.

witchcraft/service_discovery.go
tabboud: Okay got it. In that case, there are valid use cases for something like this, so it seems fine to move forward with this. The alternative sounds intriguing as it delegates constructing the `RefreshableServicesConfig` to the user which I don't think is any better or worse than having this method. The main advantage to removing this method would be that we take a stronger stance on where things should live.

bmoylan: Yeah after looking at what happens in Java and thinking about the migration path, this first approach does not provide optionality for consumers to migrate to the preferred default location.

I've pushed a commit to remove this method and expose the NewServiceDiscovery constructor so callers with nonstandard config can offroad however they want while still using `info.Clients` for the config at the correct place

tabboud (approved):
This looks great. Thanks for making the additional changes 👍

PR #332, opened 2021-10-11
improvement: Check $NETWORK_LOGGING_URL for sls-log-tcp-json-receiver URLs

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/332)

PR #337, opened 2021-10-20
improvement: Use $NETWORK_LOGGING_URL for sls-log-tcp-json-receiver URIs

bmoylan:
Reapplies #332 reverted in #333. This time, we do not crash the server in any case related to network logging.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/337)

### DIFF
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -638,9 +638,14 @@ func (s *Server) Start() (rErr error) {
 
 	// enable TCP logging if the envelope metadata and the TCP receiver are both configured
 	receiverCfg := baseRefreshableRuntimeCfg.ServiceDiscovery().CurrentServicesConfig().ClientConfig("sls-log-tcp-json-receiver")
+	// If we've been provided a URL in the environment, prefer that to whatever is in config.
+	receiverURIs := receiverCfg.URIs
+	if envURL := os.Getenv("NETWORK_LOGGING_URL"); envURL != "" {
+		receiverURIs = []string{envURL}
+	}
 	envelopeMetadata, err := tcpjson.GetEnvelopeMetadata()
 	if err != nil {
-		if len(receiverCfg.URIs) > 0 {
+		if len(receiverURIs) > 0 {
 			// Presence of TCP receiver config without envelope metadata may indicate a mis-configuration,
 			// but can also be expected in environments where the config is always hard-coded.
 			// In this case we emit a warning log to help debug potential issues, but otherwise proceed as normal.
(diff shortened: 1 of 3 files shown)


witchcraft/witchcraft.go
```diff
@@ -636,56 +636,9 @@ func (s *Server) Start() (rErr error) {
...
-		s.asyncLogWriter = tcpjson.StartAsyncWriter(tcpWriter, metricsRegistry)
-		internalHealthCheckSources = append(internalHealthCheckSources, tcpWriter)
-
-		// re-initialize the loggers with the TCP writer and overwrite the context
-		s.initDefaultLoggers(baseInstallCfg.UseConsoleLog, wlog.InfoLevel, metricsRegistry)
-		ctx = s.withLoggers(ctx)
-	}
+	// Initialize network logging client if configured
+	ctx, netLoggerHealth := s.initNetworkLogging(ctx, baseInstallCfg, baseRefreshableRuntimeCfg)
+	internalHealthCheckSources = append(internalHealthCheckSources, netLoggerHealth)
```
tabboud: how about wrapping this in a `if netLoggerHealth != nil` just so we avoid appending the nil health check source in case that causes downstream issues.

witchcraft/witchcraft.go
```diff
@@ -939,6 +892,65 @@ func (s *Server) initStackTraceHandler(ctx context.Context) {
...
+	if err != nil {
+		// The existence of receiverURIs means TCP logging was expected to be enabled, but the TCP receiver must be mis-configured.
+		// Since the server may otherwise be functional, an error log is emitted to indicate logging issues, rather
+		// than setting the TCP writer health to a state that can cause pages.
+		s.svcLogger.Warn("TCP logging will not be enabled since TLS config is unset or invalid.", svc1log.Stacktrace(err))
+		return ctx, nil
+	}
+	connProvider, err := tcpjson.NewTCPConnProvider(receiverURIs, tcpjson.WithTLSConfig(tlsConfig))
+	if err != nil {
+		s.svcLogger.Error("TCP logging will not be enabled since connection provider is invalid.", svc1log.Stacktrace(err))
```
tabboud, line 931: note: If we get to this point, then this should never actually happen, but it seems reasonable that we just disable TCP logging and log loudly rather than error out

tabboud (commented):
The change looks good to me, pending an ack from relevant parties.

witchcraft/witchcraft.go
bmoylan: 👍  We added graceful nil handling in https://github.com/palantir/witchcraft-go-health/pull/45 but I've still added this check just to remove the amount of magic

tabboud (approved):

PR #388, opened 2022-02-25
feature: Emit `server.metric_cardinality.value` with total number of timeseries in the registry

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/388)


tabboud (approved):
👍

PR #390, opened 2022-03-02
improvement: Expose server.WithRuntimeConfigProviderFunc for dynamically reading runtime config bytes

bmoylan:
Currently, the witchcraft builder does not allow the caller to customize how the runtime configuration bytes are provided. `WithRuntimeConfigProvider` accepts a constructed refreshable, but we'd like to leverage witchcraft's ctx and error handling by deferring construction with the custom function.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/390)

### DIFF
--- a/changelog/@unreleased/pr-390.v2.yml
+++ b/changelog/@unreleased/pr-390.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Expose server.WithRuntimeConfigProviderFunc for dynamically reading
+    runtime config bytes
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/390
--- a/witchcraft/witchcraft.go
+++ b/witchcraft/witchcraft.go
@@ -336,6 +336,13 @@ func (s *Server) WithRuntimeConfigProvider(r refreshable.Refreshable) *Server {
 	return s
 }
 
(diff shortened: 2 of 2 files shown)


tabboud (approved):
The main benefit I see from this approach is that it moves the construction of the refreshable and associated error handling into this library rather than being handled in the application code before hand. This seems like a reasonable trade off since most applications that instantiate a new server typically wire through these config points at the cmd layer where the loggers will not be setup yet.

Currently:
```go
func RunServer(ctx context.Context) error {
	// ctx is usually not setup with loggers yet
	fileRefreshable, err := refreshable.NewFileRefreshable(ctx, "runtime.yml")
	if err != nil {
		return err
	}
	return server.New().
		WithRuntimeConfigType(config.RuntimeConfig{}).
		WithRuntimeConfigProvider(fileRefreshable).
		Start()
}
```

After this change:
```go
func RunServer(ctx context.Context) error {
	return server.New().
		WithRuntimeConfigType(config.RuntimeConfig{}).
		WithRuntimeConfigProviderFunc(func(ctx context.Context) (refreshable.Refreshable, error) {
			return refreshable.NewFileRefreshable(ctx, "runtime.yml")
		}).
		Start()
}
```

PR #416, opened 2022-04-04
fix: Fix signal log param in shutdown handler

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/416)


witchcraft/witchcraft.go
```diff
@@ -963,7 +962,7 @@ func (s *Server) initShutdownSignalHandler(ctx context.Context) {
 
 	go wapp.RunWithRecoveryLogging(ctx, func(ctx context.Context) {
 		sig := <-shutdownSignal
-		ctx = wparams.ContextWithSafeParam(ctx, "signal", sig.String())
```
bmoylan, line 966: This context is used to control the lifecycle of `Shutdown` but is not passed to any logging. This made debugging shutdown behavior difficult because the server seemingly dies without an obvious trigger in the log stream.

tabboud (approved):

PR #417, opened 2022-04-04
Improvement: Add support for application server only middleware

asanderson15:
==COMMIT_MSG==
Add support for application server only middleware
==COMMIT_MSG==

* Currently, any configured middleware is applied to both the main application and management routers. This breaks down if you need any special handling for the main application than from the status ports.
* This change adds the ability to separately configure middleware that applies to the main application router only

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/417)


asanderson15:
Closing in favor of https://github.com/palantir/witchcraft-go-server/pull/418

PR #432, opened 2022-05-04
fix: Use consistent time string comparison when checking auth tokens.

dtrejod:
## Before this PR
We were directly comparing auth tokens strings.

## After this PR
Utilize the new secret string comparison function from CGR to check auth tokens. Fixes timing attacks on the diagnostic and health status endpoints. 
https://github.com/palantir/conjure-go-runtime/pull/294

==COMMIT_MSG==
Use consistent time string comparison when checking auth tokens.
==COMMIT_MSG==

## Possible downsides?
None

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/432)

### DIFF
--- a/witchcraft/internal/wdebug/resource.go
+++ b/witchcraft/internal/wdebug/resource.go
@@ -50,12 +50,14 @@ func RegisterRoute(router wrouter.Router, sharedSecret refreshable.String) error
 
 func (r *debugResource) ServeHTTP(rw http.ResponseWriter, req *http.Request) error {
 	ctx := req.Context()
-	authHeader, err := httpserver.ParseBearerTokenHeader(req)
-	if err != nil {
-		return errors.WrapWithNewError(err, errors.DefaultPermissionDenied)
-	}
-	if authHeader != r.SharedSecret.CurrentString() {
-		return errors.NewPermissionDenied()
+	if sharedSecret := r.SharedSecret.CurrentString(); sharedSecret != "" {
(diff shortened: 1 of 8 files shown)


witchcraft/internal/wdebug/resource.go
```diff
@@ -50,12 +50,14 @@ func RegisterRoute(router wrouter.Router, sharedSecret refreshable.String) error
...
-	authHeader, err := httpserver.ParseBearerTokenHeader(req)
-	if err != nil {
-		return errors.WrapWithNewError(err, errors.DefaultPermissionDenied)
-	}
-	if authHeader != r.SharedSecret.CurrentString() {
-		return errors.NewPermissionDenied()
+	if sharedSecret := r.SharedSecret.CurrentString(); sharedSecret != "" {
+		token, err := httpserver.ParseBearerTokenHeader(req)
+		if err != nil {
+			return errors.WrapWithUnauthorized(err)
```
dtrejod, line 56: Changed this to return a 401 rather than a 403 which is more compliant with the spec. https://www.rfc-editor.org/rfc/rfc7235#section-3.1

bmoylan (approved):
> Reviewed 7 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/432)* status: all files reviewed, 1 unresolved discussion (waiting on @dtrejod)

PR #469, opened 2022-07-01
feature: Add SERVICE_DEPENDENCY health check

bmoylan:
_See internal infra-projects#69 for background_

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/469)


changelog/@unreleased/pr-469.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: feature
+feature:
+  description: Add SERVICE_DEPENDENCY health check
```
tabboud: Can you include some of the details about how this works as well. I thought the description you wrote on the internal issue (or even the `NewServiceDependencyHealthCheck` godoc) was great to get a quick gist of when this health check would trip and what it's actually measuring.

integration/health_test.go
```diff
@@ -95,6 +97,158 @@ func TestAddHealthCheckSources(t *testing.T) {
...
+	var clients witchcraft.ConfigurableServiceDiscovery
+	server, serverErr, cleanup := createAndRunCustomTestServer(t, port, port,
+		func(ctx context.Context, info witchcraft.InitInfo) (func(), error) {
+			clients = info.Clients
+			return nil, nil
+		},
+		ioutil.Discard,
+		func(t *testing.T, initFn witchcraft.InitFunc, installCfg config.Install, logOutputBuffer io.Writer) *witchcraft.Server {
+			return createTestServer(t, initFn, installCfg, logOutputBuffer)
+		},
```
tabboud: nit: This can be simplified to just `createTestServer`

integration/health_test.go
```diff
@@ -95,6 +97,158 @@ func TestAddHealthCheckSources(t *testing.T) {
...
+
+	port, err := httpserver.AvailablePort()
+	require.NoError(t, err)
+	var clients witchcraft.ConfigurableServiceDiscovery
+	server, serverErr, cleanup := createAndRunCustomTestServer(t, port, port,
+		func(ctx context.Context, info witchcraft.InitInfo) (func(), error) {
+			clients = info.Clients
+			return nil, nil
+		},
+		ioutil.Discard,
```
tabboud, line 112: note to self: Replace all deprecated `ioutil` things with their equivalents. We can leave this as is for now for consistency, but I can follow up with this task.

witchcraft/service_discovery.go
```diff
@@ -54,11 +55,20 @@ type serviceDiscovery struct {
...
+	Health    *dependencyhealth.ServiceDependencyHealthCheck
 	UserAgent string
 }
 
-func NewServiceDiscovery(install config.Install, services config.RefreshableServicesConfig) ConfigurableServiceDiscovery {
-	return &serviceDiscovery{Services: services, UserAgent: userAgent(install)}
+func NewServiceDiscovery(
+	install config.Install,
+	services config.RefreshableServicesConfig,
+	health *dependencyhealth.ServiceDependencyHealthCheck,
```
tabboud: This is technically a break and one of the internal logging daemons currently uses this in tests. What do you think about adding a variadic options arg to this to avoid the break?

In addition, the `dependencyhealth` pkg is internal so consumers will not be able to provide this given this pkg is public facing. We should probably use some sort of interface as a bridge here.

witchcraft/internal/dependencyhealth/host_metrics.go
```diff
@@ -0,0 +1,136 @@
...
+
+	Mark2xx()
+	Mark3xx()
+	Mark4xx()
+	Mark5xx()
+	MarkError()
+	MarkOther()
+}
+
+// NewHostRegistry returns a HostMetricsRegistry. It should generally be a singleton per application.
```
tabboud:
```suggestion
// NewHostRegistry returns a new HostMetricsRegistry that is safe for concurrent use and should generally be used as a singleton per application.
```

witchcraft/internal/dependencyhealth/host_metrics.go
```diff
@@ -0,0 +1,136 @@
...
+	return &hostMetricsRegistry{
+		registry:   map[ServiceName]map[HostPort]*hostStatus{},
+		windowSize: windowSize,
+	}
+}
+
+type hostMetricsRegistry struct {
+	registry   map[ServiceName]map[HostPort]*hostStatus
+	windowSize time.Duration
+	mux        sync.Mutex
```
tabboud: nit: Move the mux to the top as a "hat" since it's guarding both fields.

witchcraft/internal/dependencyhealth/host_metrics.go
```diff
@@ -0,0 +1,136 @@
...
+package dependencyhealth
+
+import (
+	"net"
+	"sync"
+	"time"
+)
+
+type ServiceName string
+type HostPort string
```
tabboud, line 26: nit: Let's add a godoc to this mentioning that it takes the form `"host:port"`

witchcraft/internal/dependencyhealth/counter.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// timeBucket measure the number of times Mark is called within the bucket interval.
+// This is necessary over using gometrics.Meter because it only updates every 5 seconds.
+// We will call Count relatively infrequently, so optimize for correctness.
+type timeBucket struct {
+	bucket time.Duration
+	now    func() time.Time
+	slice  []time.Time
+	mux    sync.Mutex
```
tabboud: nit: put the mux on top since it's guarding all fields

witchcraft/internal/dependencyhealth/counter.go
```diff
@@ -0,0 +1,76 @@
...
+package dependencyhealth
+
+import (
+	"sync"
+	"time"
+)
+
+// timeBucket measure the number of times Mark is called within the bucket interval.
+// This is necessary over using gometrics.Meter because it only updates every 5 seconds.
+// We will call Count relatively infrequently, so optimize for correctness.
```
tabboud:
```suggestion
// Count is optimized for correctness since it is called relatively infrequently.
```

Can you confirm what you mean by "optimized for correctness"? I assume this is pointing to the fact that we prune the list before returning the count to have the most accurate count for all call-sites at a given time. If that is the case, I don't know if this adds much for the reader given we also prune in the `Mark` call too.

integration/health_test.go
```diff
@@ -95,6 +97,158 @@ func TestAddHealthCheckSources(t *testing.T) {
...
+				Type:    health.CheckType("SERVICE_DEPENDENCY"),
+				Message: stringPtr("All remote services are healthy"),
+				State:   health.New_HealthState(health.HealthState_HEALTHY),
+				Params:  make(map[string]interface{}),
+			},
+		},
+	}, getHealth())
+
+	_, _ = clientA.CurrentHTTPClient().Get(errServer.URL)
+	_, _ = clientA.CurrentHTTPClient().Get(errServer.URL)
```
tabboud, line 191: Should we just set the `httpclient.WithMaxRetries(0)` rather than request twice? I don't have strong opinions either way, but this could be confusing if you don't know that CGR retries 2 times by default.

witchcraft/internal/dependencyhealth/healthcheck.go
```diff
@@ -0,0 +1,144 @@
...
+		case resp.StatusCode < 600:
+			metric.Mark5xx()
+		}
+	}
+	return resp, respErr
+}
+
+// HealthStatus iterates through all tracked hosts for each service.
+// If there are both healthy and unhealthy hosts, the check is HEALTHY
+// but details are included in message and params.
```
tabboud:
```suggestion
// but details are included in the message and params.
```

witchcraft/internal/dependencyhealth/healthcheck.go
```diff
@@ -0,0 +1,144 @@
...
+// If there are only unhealthy hosts for at least one service, the check is WARNING.
+func (h *ServiceDependencyHealthCheck) HealthStatus(context.Context) health.HealthStatus {
+	allServices := h.hosts.AllServices()
+	if len(allServices) == 0 {
+		return health.HealthStatus{}
```
tabboud, line 93: Can you remind me how returning an empty `HealthStatus{}` affects the overall service health? Does this just not get included in the cumulative health that's reported for the service?

tabboud (changes requested):
Apologies for the delayed review. Overall this is a pretty neat feature! Most of the changes look good, but left a few comments/nits/questions below.

integration/health_test.go
bmoylan: So this is using a standard library http.Client that doesn't use CGR's retries. I've added a comment which hopefully helps.

integration/health_test.go
bmoylan: done

witchcraft/internal/dependencyhealth/counter.go
bmoylan: I think I've reworded/renamed some things to make my intention clearer. I think "consistent" is a better word than "correct"

witchcraft/internal/dependencyhealth/healthcheck.go
bmoylan: HealthStatus is a struct with a `Checks map[string]Check` field that gets `range`d over in the CombinedHealthCheck: https://github.com/palantir/witchcraft-go-health/blob/v1.14.0/status/status.go#L65

If the map is nil, the range is a no-op and nothing happens.

witchcraft/service_discovery.go
bmoylan: Agreed, thanks for calling this out. I've changed the approach to expose builder methods for arbitrary client params. The constructor signature is now unchanged. Let me know what you think

changelog/@unreleased/pr-469.v2.yml
bmoylan: done!

witchcraft/service_discovery.go
```diff
@@ -46,15 +46,22 @@ type ConfigurableServiceDiscovery interface {
 	// to benefit from live-reloadability. All config provided by WithServiceConfig will be merged into the default
 	// refreshable ServicesConfig. In case of conflict, the config provided here overrides the defaults.
 	WithServiceConfig(serviceName string, cfg httpclient.ClientConfig)
+	// WithDefaultParams stores the provided params function. When constructing a new client, the function will
+	// be evaluated and the params added to all clients.
+	WithDefaultParams(params func(serviceName string) ([]httpclient.ClientParam, error))
```
bmoylan: these changes extracted to #481

tabboud (approved):

PR #481, opened 2022-08-01
improvement: ConfigurableServiceDiscovery adds WithDefaultParams and WithServiceParams

bmoylan:
Previously, applications could set additional config overrides for witchcraft-provided clients, but only for fields available in httpclient.ClientConfig. With this change, users can now provide any httpclient param and it will be applied when those clients are constructed.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/481)


witchcraft/service_discovery.go
```diff
@@ -138,6 +182,16 @@ func (s *serviceDiscovery) serviceConfig(serviceName string) httpclient.Refresha
 	}))
 }
 
+func filterHTTPParams(in []httpclient.ClientParam) []httpclient.HTTPClientParam {
+	var out []httpclient.HTTPClientParam
+	for _, param := range in {
+		if h, ok := param.(httpclient.HTTPClientParam); ok {
```
splucs, line 197: Clients will have to convert from the HTTPClientParam interface to the ClientParam interface in order to use this feature. This will require adding a mock implementation of ClientParam, which seems quite janky.

Could we instead accept a list of ClientOrHTTPClientParam instead? This way the client chooses how to apply to a vanilla client and how to apply to an HTTP client.

splucs (changes requested):

witchcraft/service_discovery.go
bmoylan: Why does anything need to be converted? We know the exhaustive set of params because the [interface](https://github.com/palantir/conjure-go-runtime/blob/db418fba36ddde2ed5d5201e5cbe69eafa7ecb1a/conjure-go-client/httpclient/client_params.go#L33-L46) methods are private. The only one (today) that does not implement ClientParam or ClientOrHTTPClientParam is WithConfigForHTTPClient, which doesn't make sense in this case because there are already plenty of better ways to provide config. I've added a test and comment which might help.

witchcraft/service_discovery.go
```diff
@@ -46,15 +46,22 @@ type ConfigurableServiceDiscovery interface {
 	// to benefit from live-reloadability. All config provided by WithServiceConfig will be merged into the default
 	// refreshable ServicesConfig. In case of conflict, the config provided here overrides the defaults.
 	WithServiceConfig(serviceName string, cfg httpclient.ClientConfig)
+	// WithDefaultParams stores the provided params function. When constructing a new client, the function will
+	// be evaluated and the params added to all clients.
+	WithDefaultParams(params func(serviceName string) ([]httpclient.ClientParam, error))
```
tabboud: Can you mention that this can be called multiple times or something similar to make it clear the params are additive rather than overwrite an existing param func. The ordering for how the params are applied might also be relevant since params added last can overwrite a previous param.

changelog/@unreleased/pr-481.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: ConfigurableServiceDiscovery adds WithDefaultParams and WithServiceParams
```
tabboud: Can you expand on this a little bit. For example, what do these do and when would I want to use them as a consumer.

tabboud (approved):

PR #486, opened 2022-08-19
improvement: Return a proper unauthorized response if the health check shared secret is incorrect

splucs:
## Before this PR
If the health check shared secret is set and a request on the health endpoint does not set it or set it incorrectly, we return an empty health check response. This is confusing.

This stems from the fact that the status code for the health request is the health status code for the server and not for the health request itself.

## After this PR
If the health check shared secret is set and a request on the health endpoint does not set it or set it incorrectly, we return an error message.

==COMMIT_MSG==
Return a proper unauthorized response if the health check shared secret is incorrect
==COMMIT_MSG==

## Possible downsides?
Any automation that does not check the status code before attempting to unmarshal the health response might see an unexpected error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/486)

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -18,12 +18,12 @@ import (
 	"net/http"
 	"sync/atomic"
 
+	"github.com/palantir/conjure-go-runtime/v2/conjure-go-contract/errors"
 	"github.com/palantir/conjure-go-runtime/v2/conjure-go-server/httpserver"
 	"github.com/palantir/pkg/refreshable"
 	"github.com/palantir/witchcraft-go-health/conjure/witchcraft/api/health"
 	"github.com/palantir/witchcraft-go-health/status"
-	"github.com/palantir/witchcraft-go-server/vendor/github.com/palantir/conjure-go-runtime/v2/conjure-go-contract/errors"
-	wparams "github.com/palantir/witchcraft-go-server/vendor/github.com/palantir/witchcraft-go-params"
+	wparams "github.com/palantir/witchcraft-go-params"
 )
 
 // HealthHandler is responsible for checking the health-check-shared-secret if it is provided and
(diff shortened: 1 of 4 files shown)


status/status.go
```diff
@@ -49,7 +51,22 @@ func NewHealthCheckHandler(checkSource status.HealthCheckSource, sharedSecret re
 }
 
 func (h *healthHandlerImpl) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	metadata, newHealthStatusCode := h.computeNewHealthStatus(req)
+	if sharedSecret := h.healthCheckSharedSecret.CurrentString(); sharedSecret != "" {
+		token, err := httpserver.ParseBearerTokenHeader(req)
+		if err != nil {
+			errors.WriteErrorResponse(w, errors.NewUnauthorized(wparams.NewSafeParam("message", err.Error())))
```
bmoylan: generally conjure errors should be strongly typed and not include arbitrary messages like this. If we want to log more detail, it should happen server-side. I'd just do an empty `Unauthorized` here and `PermissionDenied` below once we have a token to check.

status/status.go
```diff
@@ -49,7 +50,22 @@ func NewHealthCheckHandler(checkSource status.HealthCheckSource, sharedSecret re
...
-	metadata, newHealthStatusCode := h.computeNewHealthStatus(req)
+	if sharedSecret := h.healthCheckSharedSecret.CurrentString(); sharedSecret != "" {
+		token, err := httpserver.ParseBearerTokenHeader(req)
+		if err != nil {
+			errors.WriteErrorResponse(w, errors.NewUnauthorized())
+			return
+		}
+
+		if !httpserver.SecretStringEqual(sharedSecret, token) {
+			errors.WriteErrorResponse(w, errors.NewUnauthorized())
```
bmoylan:
```suggestion
			errors.WriteErrorResponse(w, errors.NewPermissionDenied())
```

status/status.go
splucs: Addressed, though I think this will make debugging harder.

bmoylan (approved):

bmoylan (approved):

PR #560, opened 2022-12-12
improvement: Automatically rotate log files per Palantir convention

splucs:
## Summary
Per Palantir convention, we should be rotating and compressing log files on a cadence. Audit logs rotate hourly while the rest rotate daily.

## Testing
Tested locally by running the server with a shorter rotation interval and can attest log files are rotated correctly.

## Changelog
==COMMIT_MSG==
Automatically rotate log files per Palantir convention
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/560)

### DIFF
--- a/changelog/@unreleased/pr-560.v2.yml
+++ b/changelog/@unreleased/pr-560.v2.yml
@@ -1,5 +1,5 @@
 type: improvement
 improvement:
-  description: Automatically rotate log files per Palantir convention
+  description: Automatically rotate log files per Palantir convention - audit logs are rotated hourly, other logs are rotated daily
   links:
   - https://github.com/palantir/witchcraft-go-server/pull/560
--- a/changelog/@unreleased/pr-560.v2.yml
+++ b/changelog/@unreleased/pr-560.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Automatically rotate log files per Palantir convention
+  links:
+  - https://github.com/palantir/witchcraft-go-server/pull/560
--- a/witchcraft/loggers.go
+++ b/witchcraft/loggers.go
@@ -188,14 +188,15 @@ func periodicallyRotateLogFiles(ctx context.Context, logger *lumberjack.Logger, 
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
 	for {
+		// Rotate once at start. If there is no file, it is no op.
+		if err := logger.Rotate(); err != nil {
+			// At this point the logger should have been initialized, so we can log the error as a best effort.
+			svc1log.FromContext(ctx).Error("Error rotating log file", svc1log.Stacktrace(err))
+		}
 		select {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			if err := logger.Rotate(); err != nil {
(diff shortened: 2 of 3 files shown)


changelog/@unreleased/pr-560.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Automatically rotate log files per Palantir convention
```
bmoylan: Let's instead specify what these conventions are: Audit logs are rotated hourly; non-audit logs every 24 hours.

witchcraft/loggers.go
```diff
@@ -141,17 +162,43 @@ func (s *Server) initWrappedLoggers(useConsoleLog bool, productName, productVers
...
 		MaxAge:     30,
 		Compress:   true,
 	}
+	flusher := func(ctx context.Context) {
+		periodicallyRotateLogFiles(ctx, logger, slsFilename)
+	}
+	return logger, flusher
+}
+
+func periodicallyRotateLogFiles(ctx context.Context, logger *lumberjack.Logger, slsFilename string) {
```
bmoylan: nit: `periodicallyRotateLogFile` since it is only doing one file

witchcraft/loggers.go
```diff
@@ -141,17 +162,43 @@ func (s *Server) initWrappedLoggers(useConsoleLog bool, productName, productVers
...
+	interval := defaultLogRotationInterval
+	if slsFilename == auditLogFilename {
+		interval = auditLogRotationInterval
+	}
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		// Rotate once at start. If there is no file, it is no op.
+		if err := logger.Rotate(); err != nil {
```
bmoylan, line 192: I guess there's no way to restart the timer when the file is rotated due to MaxSize? We're going to end up with smaller-than-desired files if we have two uncoordinated thresholds

witchcraft/witchcraft.go
```diff
@@ -679,6 +683,11 @@ func (s *Server) Start() (rErr error) {
 		}
 	}
 
+	// initialize log flushers
+	for _, logFlusher := range s.logFlushers {
```
bmoylan, line 687: Why bother accumulating these and launching them in *Server? Couldn't this just `go periodicallyRotateLogFiles(ctx, logger, slsFilename)` in newDefaultLogOutputWriter?

bmoylan (changes requested):

witchcraft/loggers.go
splucs: There is no way unfortunately. The lumberjack logger doesn't expose the current file size nor when was it last rotated.

witchcraft/witchcraft.go
splucs: The rotation function stops if the context is cancelled and we only have a context in the main server function after all loggers are initialized. Basically we want the go routine to exit when the server is stopped.

witchcraft/loggers.go
splucs: Fixed.

PR #661, opened 2023-07-18
fix: Do not register /debug/pprof routes without management port

bmoylan:
## Before this PR
/debug/pprof endpoints were registered on the main server

## After this PR
==COMMIT_MSG==
Do not register /debug/pprof routes without management port
==COMMIT_MSG==

## Possible downsides?
Possible workflow break for users using these endpoints

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/661)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ specifies the encoding and format of the response. The following types are curre
 * `metric.names.v1`: Records all metric names and tag sets in the process's metric registry.
 
 #### \[Deprecated] Pprof routes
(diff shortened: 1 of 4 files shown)


ashrayjain:
Noting that the reason this is worth doing is so that potentially sensitive unauthed endpoints are not accidentally exposed broadly.

ashrayjain:
👍

ungureanuvladvictor (approved):

tabboud (approved):

nmiyake (approved):

PR #673, opened 2023-08-14
add debug-shared-secret-file

teekenzie:
## Before this PR

There is no way to pass in a file for the shared-secret for the diagnostics route. This forces the users to inline the shared-secret.

## After this PR

==COMMIT_MSG==
Users are able to pass in the path to the file that stores the shared secret to the `debug-shared-secret-file` field. However, if the `debug-shared-secret` field is defined, it takes precedence.
==COMMIT_MSG==

### DIFF
--- a/witchcraft/server_routes.go
+++ b/witchcraft/server_routes.go
@@ -196,7 +196,10 @@ func getSecretRefreshable(ctx context.Context, diagnosticsConfig config.Refresha
 		return nil, err
 	}
 	secretStringFromFileRefreshable := refreshable.NewString(fileRefreshable.Map(func(i interface{}) interface{} {
-		secretBytes := i.([]byte)
+		secretBytes, ok := i.([]byte)
+		if !ok {
+			return ""
+		}
 		return string(secretBytes)
 	}))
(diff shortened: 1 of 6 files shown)


witchcraft/server_routes.go
```diff
@@ -172,3 +179,25 @@ func heap(w http.ResponseWriter, _ *http.Request) {
...
+
+	if secretFromConfig.CurrentString() != "" {
+		return secretFromConfig, nil
+	}
+	fileRefreshable, err := refreshablefile.NewFileRefreshable(ctx, secretFilePath.CurrentString())
+	if err != nil {
+		return nil, err
+	}
+	secretStringFromFileRefreshable := refreshable.NewString(fileRefreshable.Map(func(i interface{}) interface{} {
+		secretBytes := i.([]byte)
```
jamesross3: can we be extra safe here and do:
```
secretBytes, ok := i.([]byte)
if !ok { return "" }
```
just to avoid any panics? We typically do this in places even where we are confident in the type assertion always succeeding.

jamesross3:
👍

PR #694, opened 2023-10-10
improvement: Use os.Executable() for default user-agent

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/694)

### DIFF
--- a/witchcraft/service_discovery.go
+++ b/witchcraft/service_discovery.go
@@ -16,6 +16,8 @@ package witchcraft
 
 import (
 	"context"
+	"os"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


witchcraft/service_discovery.go
```diff
@@ -209,5 +211,8 @@ func userAgent(install config.Install) string {
 		}
 		return agent
 	}
+	if exe, err := os.Executable(); err == nil && exe != "" {
```
bmoylan, line 214: https://pkg.go.dev/os#Executable

PR #758, opened 2024-03-11
improvement: Bump tlsconfig pkg to pickup new TLS ciphers. Bump Go to 1.21.x.

dtrejod:
## Before this PR
We were using ciphers that are considered insecure in Go 1.22. Bump tlsconfig pkg to pickup new ciphers:
https://github.com/palantir/pkg/pull/340

## After this PR
Side effect of bumping tlsconfig is the need to also bump the Go version to at least 1.21.x. 

==COMMIT_MSG==
Bump tlsconfig pkg to pickup new TLS ciphers.
==COMMIT_MSG==


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/palantir/witchcraft-go-server/v2
 
-go 1.20
+go 1.21
+
+toolchain go1.21.5
```
dtrejod: 💭 Not sure why these were added after running `./godelw mod`?

dtrejod: Removed the toolchain directive at the very least.

integration/server_test.go
```diff
@@ -318,7 +318,7 @@ func TestClientTLS(t *testing.T) {
 	// Assert regular client receives error
 	_, err = testServerClient().Get(fmt.Sprintf("https://localhost:%d/example", port))
 	require.Error(t, err, "Client allowed to make request without cert")
-	assert.Contains(t, err.Error(), "tls: bad certificate")
+	assert.ErrorContains(t, err, "tls", "expected error to contain 'tls' in its string representation")
```
dtrejod, line 321: The error string here changed between 1.20 and 1.21. To accommodate I updated the test to check look for `tls` somewhere in the error string.

dtrejod: replaced by https://github.com/palantir/witchcraft-go-server/pull/761

PR #761, opened 2024-03-11
fix: Fix tls integration test to allow for go 1.21 upgrade.

dtrejod:
## Before this PR
The go 1.20->1.21 version bump is failing because the TLS error strings changes between versions:
https://github.com/palantir/witchcraft-go-server/pull/760

## After this PR
==COMMIT_MSG==
Fix tls integration test to allow for go 1.21 upgrade.
==COMMIT_MSG==

### DIFF
--- a/integration/server_test.go
+++ b/integration/server_test.go
@@ -318,7 +318,7 @@ func TestClientTLS(t *testing.T) {
 	// Assert regular client receives error
 	_, err = testServerClient().Get(fmt.Sprintf("https://localhost:%d/example", port))
 	require.Error(t, err, "Client allowed to make request without cert")
-	assert.Contains(t, err.Error(), "tls: bad certificate")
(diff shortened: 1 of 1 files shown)


integration/server_test.go
```diff
@@ -318,7 +318,7 @@ func TestClientTLS(t *testing.T) {
 	// Assert regular client receives error
 	_, err = testServerClient().Get(fmt.Sprintf("https://localhost:%d/example", port))
 	require.Error(t, err, "Client allowed to make request without cert")
-	assert.Contains(t, err.Error(), "tls: bad certificate")
+	assert.ErrorContains(t, err, "tls", "expected error to contain 'tls' in its string representation")
```
dtrejod, line 321: The error string here changed between 1.20 and 1.21. To accommodate I updated the test to check look for tls somewhere in the error string.

In go 1.20 the error is `tls: bad certificate`. In go 1.21 it's `tls: certificate required`.

bmoylan (approved):

PR #847, opened 2024-10-03
improvement: Tolerate missing client CA files at startup

adisb01:
## Before this PR
Currently, `serverconfig.ClientCAFiles` specifies the paths the server expects the CA certificates used to verify the certificates of clients to exist. However, if at startup certificates don't exist there, or there is a failure to parse them, the server fails to come up. 

## After this PR
A server can now be configured fall back to the default system CAs if the certificates at paths specified in `serverconfig.ClientCAFiles` don't exist at startup time. This is achieved by constructing the sever using `WithIgnoreMissingClientCAFiles`.

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/witchcraft/server_run.go
+++ b/witchcraft/server_run.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"math/big"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/palantir/pkg/tlsconfig"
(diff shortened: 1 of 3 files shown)


witchcraft/server_run.go
```diff
@@ -115,6 +120,19 @@ func newTLSConfig(serverConfig config.Server, useSelfSignedServerCertificate boo
 	return tlsConfig, nil
 }
 
+func serverClientCAFiles(tlsCertProvider tlsconfig.TLSCertProvider, ignoreMissingClientCAFiles bool, clientCAFiles ...string) tlsconfig.ServerParam {
+	clientCAsParam := tlsconfig.ServerClientCAFiles(clientCAFiles...)
+	if !ignoreMissingClientCAFiles {
+		return clientCAsParam
+	}
+	_, err := tlsconfig.NewServerConfig(tlsCertProvider, clientCAsParam)
```
adisb01: This feels a bit hacky, but I wasn't sure how else to determine if there's any errors with the current set of provided file without making contributions to other vendor libraries (like `ServerClientCAs` in tlsconfgi_server.go). Would appreciate any suggestions/improvements.

witchcraft/server_run.go
```diff
@@ -103,8 +106,10 @@ func newTLSConfig(serverConfig config.Server, useSelfSignedServerCertificate boo
 		return nil, werror.Error(msg + " for server not specified in configuration")
 	}
 
+	tlsCertProvider := newTLSCertProvider(useSelfSignedServerCertificate, serverConfig.CertFile, serverConfig.KeyFile)
 	tlsConfig, err := tlsconfig.NewServerConfig(
-		newTLSCertProvider(useSelfSignedServerCertificate, serverConfig.CertFile, serverConfig.KeyFile),
+		tlsCertProvider,
+		serverClientCAFiles(tlsCertProvider, ignoreMissingClientCAFiles, serverConfig.ClientCAFiles...),
```
AdamDeHovitz: You still call `tlsconfig.ServerClientCAFiles` on the next line

witchcraft/server_run.go
AdamDeHovitz: We can just check if the `clientCAFiles` exist

adisb01:
Pursuing a different fix

PR #848, opened 2024-10-10
Improvement: Avoid server shutdown race conditions by adding new ServerShutdownInProgress state

AdamEAnderson:
## Before this PR
 
Previously, there were several issues with server shutdown:
1. If shutdown was called multiple times, e.g. by different goroutines, all but one would return an error
2. If shutdown was called too frequently around when the first shutdown succeeded, the WaitGroup would panic
3. If the context was closed during a shutdown an error would be emitted
4. The server did not emit an error when restarting the server while a shutdown was in progress
5. Several other extremely unlikely race conditions (because Server has no locking)

Most clients will call `Shutdown` from multiple critical goroutines when they stop/fail unrecoverably. This introduces a race where the first caller begins the shutdown and the rest return errors. Clients often treat an error from Shutdown as a critical failure and `panic`. This can prevent a graceful shutdown from happening as expected, leading to the server being closed before connections are drained and `Start()` exiting with a non-nil error.

## After this PR

A new server state has been added called "ServerShutdownInProgress". This is used to synchronize shutdown attempts and ensure that exactly one shutdown for any reason will actually do the work of stopping the server. `Start()` on exit checks if the state is a shutdown in progress, and if so waits for the shutdown winner to report completion before exiting. Only the shutdown winner is allowed to complete the shutdown by closing the channel and flipping the state to idle.

I've updated some of the existing usages of `setState` to use the better-protected `compareAndSwapState` with proper protections against concurrent state change attempts.

As a client, the changes are:
- Calling shutdown multiple times or from multiple goroutines will result in the server being shutdown once and additional attempts not performing any actions
- Calling shutdown with a closed context or with a context that closes during shutdown results in a less graceful shutdown (e.g. `Close()`) but will still return a nil error
- Various race conditions around shutdown can no longer occur

I've added the following tests:
1. Starting, stopping, and then starting a server again works as expected
2. Stopping a server multiple times results in a single shutdown with other callers getting nil errors
3. Calling shutdown with a cancelled context succeeds

==COMMIT_MSG==
Avoid server shutdown race conditions by adding new ServerShutdownInProgress state
==COMMIT_MSG==

### DIFF
--- a/witchcraft/server_state.go
+++ b/witchcraft/server_state.go
@@ -30,6 +30,7 @@ const (
 	ServerIdle ServerState = iota
 	ServerInitializing
 	ServerRunning
+	ServerShutdownInProgress
 )
 
 func (s ServerState) String() string {
@@ -40,6 +41,8 @@ func (s ServerState) String() string {
 		return "initializing"
 	case ServerRunning:
 		return "running"
+	case ServerShutdownInProgress:
+		return "shutdown in progress"
 	default:
 		return "unknown state: " + strconv.Itoa(int(s))
 	}
@@ -61,6 +64,8 @@ func (s *serverStateManager) Start() error {
 		return werror.Error("server is already initializing and must be stopped before it can be started again")
 	case ServerRunning:
 		return werror.Error("server is already running and must be stopped before it can be started again")
+	case ServerShutdownInProgress:
+		return werror.Error("server is in the process of shutting down and must complete before it can be started again")
 	default:
 		return werror.Error("server is in an unknown state and must be stopped before it can be started again")
 	}
--- a/witchcraft/server_state.go
+++ b/witchcraft/server_state.go
@@ -30,7 +30,7 @@ const (
 	ServerIdle ServerState = iota
 	ServerInitializing
 	ServerRunning
-	ServerShutdownInProgress
+	ServerShuttingDown
 )
 
 func (s ServerState) String() string {
@@ -41,8 +41,8 @@ func (s ServerState) String() string {
 		return "initializing"
 	case ServerRunning:
 		return "running"
-	case ServerShutdownInProgress:
-		return "shutdown in progress"
+	case ServerShuttingDown:
+		return "shutting down"
 	default:
 		return "unknown state: " + strconv.Itoa(int(s))
 	}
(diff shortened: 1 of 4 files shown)


witchcraft/server_state.go
```diff
@@ -30,6 +30,7 @@ const (
 	ServerIdle ServerState = iota
 	ServerInitializing
 	ServerRunning
+	ServerShutdownInProgress
```
bmoylan: nit: `ServerShuttingDown` or `ServerClosing` might better fit the tense of the other states

witchcraft/witchcraft.go
```diff
@@ -944,19 +957,21 @@ func (s *Server) State() ServerState {
...
-		return svr.Shutdown(ctx)
+		err := svr.Shutdown(ctx)
+		if err != nil && ctx.Err() != nil && errors.Is(err, ctx.Err()) {
+			// Server shutdown was interrupted by context completion, but this still results in a successful shutdown
+			return nil
+		} else if err != nil {
+			return err
+		} else {
+			return nil
+		}
```
nmiyake: Suggestion to modify structure to reduce branches/make logic a bit easier to follow:

```suggestion
		if err := svr.Shutdown(ctx); err != nil && (ctx.Err() == nil || !errors.Is(err, ctx.Err()) {
		   // error is non-nil and not a context error: indicates that there was an error shutting down
		   return err
		}

		// errors was nil or server shutdown was interrupted by context completion: either one is considered a successful shutdown
		return nil
```

witchcraft/witchcraft.go
```diff
@@ -1035,10 +1050,27 @@ func decryptNodeValues(n *yamlv3.Node, kwt *encryptedconfigvalue.KeyWithType) er
...
+	for {
+		if curState == ServerIdle || curState == ServerShutdownInProgress {
+			// already shutting down or stopped
+			return nil
+		}
+		// state could be ServerRunning or ServerInitializing
+		if s.stateManager.compareAndSwapState(curState, ServerShutdownInProgress) {
+			break
+		}
+		curState = s.stateManager.State()
```
nmiyake, line 1061: What's the condition where this occurs (either of the above states are not met)? Trying to understand since in that scenario this will be a busy/active wait

nmiyake (changes requested):
Thanks! Looks good overall, but some questions and minor feedback.

witchcraft/witchcraft.go
AdamEAnderson: Two shutdown attempts happen at the same time. Both check the current state and see it as running (i.e. both read the state before either sets it). Then both set it to shutdown in progress and run the stopper. Now there is no single shutdown in progress, there are multiple.

Using atomics with compare and swap ensures that only one will win. This is a busy wait, but will almost never be run more than once, and its basically impossible to run more than twice. If you lost the first iteration, then you get the state again and see that a shutdown is already in progress so you let that thread handle it.

nmiyake: Got it, thanks for the explanation!

nmiyake (dismissed):
Thanks, looks good to me!

Removing "merge when ready" to give the opportunity for another review if needed, but also not opposed to merging with this one review.

bmoylan (dismissed):
Looks good to me! One comment on naming, but otherwise thanks for putting this together and the detailed explanations as to what's going on

bmoylan (approved):

PR #850, opened 2024-10-22
Add GoroutinesV2 diagnostic handler for emitting goroutine pprofs

jamesross3:


jamesross3:
@nmiyake and @bmoylan - I've hit a wall using the existing goroutine pprof diagnostic when debugging goroutine leaks. The main issue is that some runtime/pprof features such as goroutine labeling don't make it into the txt output which this existing diagnostic produces. I propose we add this pprof output for goroutines, which enables expedited goroutine leak detection. See https://rakyll.org/profiler-labels/ for a simple example of goroutine pprof visualizations with labels.

witchcraft/internal/wdebug/handlers.go
```diff
@@ -76,7 +78,7 @@ func (h handlerGoroutinesV1) SafeLoggable() bool {
 }
 
 func (h handlerGoroutinesV1) Extension() string {
-	return "txt"
+	return "prof"
```
bmoylan: Is this useful somehow on the old format or can we keep it as-is?

jamesross3: oops meant to revert this

jamesross3: fixed

changelog/@unreleased/pr-850.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: feature
+feature:
+  description: |
+    Add DiagnosticTypeGoroutinesV2 debug handler.
+  
+    This diagnostic emits a goroutines pprof, compatible with go tool pprof. It provides additional information which the existing DiagnosticTypeGoroutinesV1 does not, such as pprof.LabelSet attribution of goroutines.
+  links:
+    - https://github.com/palantir/witchcraft-go-server/pull/850
```
nmiyake: nit: add newline at end of file

jamesross3: Thanks - done

bmoylan (approved):

PR #891, opened 2025-01-17
export the DiagnosticHandler type

jamesross3:


witchcraft/wdebug/api.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package wdebug
+
+import (
+	"context"
+	"io"
+)
+
+type DiagnosticType string
```
nmiyake, line 31: Can this be anything, or are there any guidelines/schema/requirements for the format of values of this type?

witchcraft/internal/wdebug/handlers.go
```diff
@@ -26,20 +26,27 @@ import (
 	"github.com/palantir/pkg/metrics"
 	werror "github.com/palantir/witchcraft-go-error"
 	wparams "github.com/palantir/witchcraft-go-params"
+	"github.com/palantir/witchcraft-go-server/v2/witchcraft/wdebug"
 )
 
+// DiagnosticHandler is deprecated in favor of the exported/non-external wdebug.DiagnosticHandler.
```
nmiyake:
```suggestion
// DiagnosticHandler is deprecated in favor of the exported wdebug.DiagnosticHandler.
```

What does "/non-external" mean here? It looks like that interface is exported/external.

witchcraft/wdebug/api.go
```diff
@@ -0,0 +1,31 @@
...
+
+type DiagnosticHandler interface {
+	Type() DiagnosticType
```
nmiyake, line 44: I know this content is copy-pasted/moved from what already existed, but would be nice to have at least some documentation about the interface and behavior for functions where functionality is non-obvious

nmiyake (changes requested):
Minor feedback, but core of change looks good

witchcraft/wdebug/api.go
sfackler: > Type names must be specific, reasonably expected to be unique, and versioned to allow for future major changes of the payload structure. DiagnosticTypes must match the regular expression ([a-z0-9]+\.)+v[0-9]+, i.e. be lower-case, dot-delimited, and end with a version suffix. For example, the threaddump.v1 diagnosticType might indicate a value of ThreadDumpV1 from the DiagnosticLogV1 definition.

From https://github.palantir.build/deployability/sls-spec/blob/develop/docs/debug.md#diagnostic-types-and-values

jamesross3: updated to include a `DiagnosticType.Validate()` method. This method is now called upon wdebug route registration. There is also now a test validating all DiagnosticType constants defined within internal/wdebug

witchcraft/internal/wdebug/handlers.go
jamesross3: thanks for catching, fixed

witchcraft/wdebug/api.go
jamesross3: added documentation

witchcraft/wdebug/api.go
```diff
@@ -0,0 +1,54 @@
...
+func (d DiagnosticType) Validate() error {
+	if !diagnosticTypeRegexp.MatchString(string(d)) {
+		return fmt.Errorf("'%s' does not match expected DiagnosticType format (%s)", d, diagnosticTypeRegexp.String())
+	}
+	return nil
+}
+
+// DiagnosticHandler provides methods for describing the type and nature of a diagnostic payload, as well as writing the payload to a writer.
+type DiagnosticHandler interface {
+	// Type returns the DiagnosticType for
```
nmiyake: Comment is unfinished

nmiyake (changes requested):
Thanks! Feedback on comment and looks like you need to run `./godelw license`

witchcraft/wdebug/api.go
jamesross3: thanks for catching, fixed

nmiyake (approved):

PR #896, opened 2025-01-27
feature: ENDPOINT_FIVE_HUNDREDS health check

bmoylan:
Adds a ENDPOINT_FIVE_HUNDREDS that triggers a WARNING state when one or more wrouter endpoints returns only `500 Internal Server Error` in each 2 minute bucket over a 10 minute period.

### DIFF
--- a/witchcraft/internal/middleware/endpoint_five_hundreds.go
+++ b/witchcraft/internal/middleware/endpoint_five_hundreds.go
@@ -1,0 +1,177 @@
+// Copyright (c) 2025 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middleware
+
+import (
+	"context"
+	"fmt"
+	"net/http"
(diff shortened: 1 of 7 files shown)


witchcraft/internal/middleware/endpoint_five_hundreds.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/palantir/witchcraft-go-health/conjure/witchcraft/api/health"
+	"github.com/palantir/witchcraft-go-server/v2/wrouter"
+)
+
+const (
+	endpointFiveHundredsCheckType              = "ENDPOINT_FIVE_HUNDREDS"
+	endpointFiveHundredsWindowCount            = 5
+	endpointFiveHundredsWindowSize             = 2 * time.Minute
+	endpointFiveHundredsMessage                = "There have been HTTP 500s returned by endpoints in every 2 minutes rolling period in the last 10 minutes. This indicates a non-transient error."
```
nmiyake, line 33:
```suggestion
	endpointFiveHundredsMessage                = "There have been HTTP 500s returned by endpoints in every 2 minute rolling period in the last 10 minutes. This indicates a non-transient error."
```

nmiyake: I think this is fine since there's not a way to do this without making the message a non-constant, but will note that the message content here is semantically tied to the value of the `endpointFiveHundredsWindowSize` constant but this isn't enforced (if that constant is changed, message will no longer be accurate)

nmiyake (changes requested):
Minor feedback on message, but otherwise looks good!

witchcraft/internal/middleware/endpoint_five_hundreds.go
nmiyake: Clarified with Brad that this message matches the Java implementation exactly, so keeping as-is

nmiyake (approved):

PR #911, opened 2025-02-21
Add counter metric tracking main server connection state changes.

jamesross3:
## Before this PR
WGS provided no telemetry around server connections.

## After this PR
WGS provides connection state change telemetry. This enables developers to observe:
- total number of open connections the server has (`count{state="new"} - count{state="closed"}`)
- number of active versus idle connections

==COMMIT_MSG==
Add counter metric tracking main server connection state changes.
==COMMIT_MSG==

## Possible downsides?
This is an additional metric, so this technically costs money for any infrastructure that collects metrics for WGS applications.

### DIFF
--- a/changelog/@unreleased/pr-911.v2.yml
+++ b/changelog/@unreleased/pr-911.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    Add counter metric tracking main server connection state changes.
+  links:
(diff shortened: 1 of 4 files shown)


jamesross3:
Re: question about `state` tag cardinality - the set of possible values are documented [here in the golang http package](https://pkg.go.dev/net/http#ConnState). The possible values are:
- new
- active
- idle
- hijacked
- closed

bmoylan (approved):

PR #928, opened 2025-03-27
improvement: Avoid computing unused metric values

bmoylan:
Depends on https://github.com/palantir/pkg/pull/374

We are seeing a lot of memory usage wasted calculating metric cardinality and values that exist in the blacklists. By checking keys before evaluating values, we can reduce this extra memory pressure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/928)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/palantir/go-encrypted-config-value v1.39.0
 	github.com/palantir/go-metrics v1.1.1
 	github.com/palantir/pkg/httpserver v1.1.0
-	github.com/palantir/pkg/metrics v1.7.0
+	github.com/palantir/pkg/metrics v1.7.1-0.20250327175756-b74771023a41
```
bmoylan: DRAFT until there's a release with https://github.com/palantir/pkg/pull/374

witchcraft/server_metrics.go
```diff
@@ -141,14 +141,31 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
 	}, nil
 }
 
-func (s *Server) filterMetricValues(metricID string, metricVal metrics.MetricVal) map[string]interface{} {
+func (s *Server) collectMetricValues(metricName string, metricVal metrics.MetricVal) map[string]interface{} {
+	valuesToUse := make(map[string]interface{})
```
nmiyake: nit, but for new code, slight preference for `any` over `interface{}` (but don't feel strongly about this, especially given that `interface{}` is still used in other existing code in this file)

```suggestion
func (s *Server) collectMetricValues(metricName string, metricVal metrics.MetricVal) map[string]any {
	valuesToUse := make(map[string]any)
```

witchcraft/server_metrics.go
```diff
@@ -141,14 +141,31 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
...
+	s.visitMetricValues(metricName, metricVal, func(key string) {
+		valuesToUse[key] = metricVal.Value(key)
+	})
+	return valuesToUse
+}
+
+// visitMetricValues calls the provided visit function for each key in the provided metric value,
+// If the metric name or key is blacklisted, the visit function is not called.
```
nmiyake: Documentation could be a bit clearer

```suggestion
// visitMetricValues calls the provided visit function for each key in the provided metric value.
// If the metric with the provided metricID is in the metrics blacklist, the visit function is not called for any keys.
// The visit function is not called for keys that are in the metric type value blacklist for the type.
```

nmiyake (changes requested):
Minor feedback on documentation

nmiyake (approved):

PR #929, opened 2025-03-27
improvement: Mark uptime, cardinality, and go runtime metrics immediately before logging metric.1 lines

bmoylan:
Requires https://github.com/palantir/pkg/pull/373

`server.uptime`, `server.metrics_cardinality`, and `go.runtime.*` are all incremented on their own schedules. This means the values are outdated when logged, and updates that rewritten before any logging occurs is wasteful.

This PR updates `initServerMetrics` to include these update in the same schedule that we log the metric.1 lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/witchcraft-go-server/929)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/palantir/go-encrypted-config-value v1.39.0
 	github.com/palantir/go-metrics v1.1.1
 	github.com/palantir/pkg/httpserver v1.1.0
-	github.com/palantir/pkg/metrics v1.7.0
+	github.com/palantir/pkg/metrics v1.7.1-0.20250327175645-72c2fb0f8d96
```
bmoylan: DRAFT until there's a release with https://github.com/palantir/pkg/pull/373

witchcraft/server_metrics.go
```diff
@@ -130,14 +127,33 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
 		localMetricLogger.Metric(metricID, metricType, metric1log.Values(valuesToUse), metricTagsParam)
 	}
 
+	emitAllFn := func() {
+		// emit all metrics a final time on termination
```
nmiyake: I don't think this comment makes sense -- doesn't `emitAllFn` get called both periodically while alive *and* on termination?

witchcraft/server_metrics.go
```diff
@@ -130,14 +127,33 @@ func (s *Server) initMetrics(ctx context.Context, installCfg config.Install) (rR
...
+		if !s.disableGoRuntimeMetrics {
+			metrics.CaptureRuntimeMemStatsOnce(metricsRegistry)
```
nmiyake: I'm confused about this -- the still existing logic at line 82 calls `metrics.CaptureRuntimeMemStatsWithContext`, which starts a routine that captures the runtime stats at a given frequency.

This `emitAllFn` is also called in a goroutine at line 147, which is also called periodically. Was the intention to remove the call at line 82?

nmiyake (changes requested):

witchcraft/server_metrics.go
bmoylan: Yes, apologies, I incorrectly copied this from the function literal below. Removed

witchcraft/server_metrics.go
bmoylan: Yes -- this now initializes using the CaptureRuntimeMemStatsFunc updated in https://github.com/palantir/pkg/pull/373 and the function is (only) called below in emitAllFn.

nmiyake (dismissed):

nmiyake (approved):

PR #942, opened 2025-04-30
Add support for audit.v3 logs

nmiyake:
## After this PR

==COMMIT_MSG==
Adds support for `audit.3` logs.

Specifically, ensures that an `audit3.Logger` is set on all server contexts and that there is a middleware that properly populates required fields. Also defines an `AuditConfig` struct and adds it as an optional field to the base `Runtime` configuration.
==COMMIT_MSG==


nmiyake:
Requires https://github.com/palantir/witchcraft-go-logging/pull/400

config/runtime.go
```diff
@@ -26,10 +26,25 @@ import (
...
 	HealthChecks      HealthChecksConfig        `yaml:"health-checks,omitempty"`
+	AuditConfig       *AuditConfig              `yaml:"audit,omitempty"`
 	LoggerConfig      *LoggerConfig             `yaml:"logging,omitempty"`
 	ServiceDiscovery  httpclient.ServicesConfig `yaml:"service-discovery,omitempty"`
 }
 
+type AuditConfig struct {
+	Deployment string `yaml:"deployment,omitempty"`
+
+	Product string `yaml:"product-name,omitempty"`
```
bmoylan, line 37: Is there a reason to not use the existing `ProductName` and `ProductVersion` from install config?

config/runtime.go
```diff
@@ -26,10 +26,25 @@ import (
 type Runtime struct {
 	DiagnosticsConfig DiagnosticsConfig         `yaml:"diagnostics,omitempty"`
 	HealthChecks      HealthChecksConfig        `yaml:"health-checks,omitempty"`
+	AuditConfig       *AuditConfig              `yaml:"audit,omitempty"`
```
bmoylan, line 29: Do we expect any of this to actually be updated at runtime? Would it be simpler to put in install config?

nmiyake: Similar to the above, this is based on the Java implementation. For Java products, the recommended way to adopt `audit.3` is to add the following block to the runtime configuration:

```
    audit:
      deployment: "{{environment.deployment}}"
      product-name: "{{conf.install.product-name}}"
      product-version: "{{conf.install.product-version}}"
      service: "{{service_name}}"
      stack: "{{stack_name}}"
      environment: "{{environment.name}}"
```

I'm not sure if a subset of these can be updated at runtime or if there's another reason for the block being there -- @IlanaRadinsky do you have any further context for this?

config/runtime.go
nmiyake: Good question -- so this config mirrors `AuditConfiguration.java` on the Java side. @IlanaRadinsky do you have context for why this is defined separately? Is it because in Java it's hard to read off the "general" one?

witchcraft/loggers.go
```diff
@@ -135,6 +157,56 @@ func (s *Server) initWrappedLoggers(useConsoleLog bool, productName, productVers
...
+
+	audit3Logger = audit3log.WithParams(audit3Logger,
+		audit3log.Deployment(deployment),
+		audit3log.Host(host),
+		audit3log.Product(product),
+		audit3log.ProductVersion(productVersion),
+		audit3log.Stack(stack),
+		audit3log.Service(service),
+		audit3log.Environment(environment),
+		audit3log.ProducerType(audit3log.AuditProducerServer),
```
bmoylan, line 203: Could this context parameter going to flow down to audit.3 logs emitted as part of client requests?

nmiyake: @IlanaRadinsky I believe you said this value should always be "Server", correct?

IlanaRadinsky: Yes, because these are server-generated audit logs. In foundry-audit this value is always Server as well.

config/runtime.go
IlanaRadinsky: This predates me on audit so I don't have context - I'll look into it!

config/runtime.go
IlanaRadinsky: I don't think there's a good reason for this - it should work to just use the values from the install config.

nmiyake:
@bmoylan @IlanaRadinsky ready for review -- per the Slack conversation, I believe the plan is to move forward with keeping the Audit Configuration the same as Java for compatibility/consistency.

IlanaRadinsky (commented):
LGTM!

bmoylan (approved):

config/runtime.go
bmoylan: Discussed on internal thread, and concluded that it's not really expected to change but we'll leave this as-is for consistency with existing implementations on other toolchains

PR #1014, opened 2025-10-07
improvement: Allow disabling HTTP/2 support

okushchenko:
This is required when witchcraft-go-server is used behind a proxy that attempts to do HTTP/1 to HTTP/2 upgrade based on ALPN info returned in the TLS handshake response. The default for Go is to always return info that a correctly configured proxy will use to upgrade the connection to HTTP/2. This change allows disabling that default in cases when HTTP/2 upgrade breaks functionality, e.g. with WebSocket.

See https://pkg.go.dev/net/http

> Starting with Go 1.6, the http package has transparent support for the HTTP/2 protocol when using HTTPS. Programs that must disable HTTP/2 can do so by setting [Transport.TLSNextProto] (for clients) or [Server.TLSNextProto] (for servers) to a non-nil, empty map.

### DIFF
--- a/config/install.go
+++ b/config/install.go
@@ -39,4 +39,5 @@ type Server struct {
 	ClientCAFiles  []string `yaml:"client-ca-files,omitempty"`
 	CertFile       string   `yaml:"cert-file,omitempty"`
 	KeyFile        string   `yaml:"key-file,omitempty"`
+	DisableHTTP2   bool     `yaml:"disable-http2,omitempty"`
 }
--- a/config/install.go
+++ b/config/install.go
@@ -39,5 +39,4 @@ type Server struct {
 	ClientCAFiles  []string `yaml:"client-ca-files,omitempty"`
 	CertFile       string   `yaml:"cert-file,omitempty"`
 	KeyFile        string   `yaml:"key-file,omitempty"`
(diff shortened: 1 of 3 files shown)


nmiyake (commented):
Thanks -- this functional change looks good.

What's the decision criteria for services to actually set this in configuration?

okushchenko:
The one that applies directly to my use-case is if the service is mounting WebSocket endpoint handlers on the witchcraft router - then HTTP/2 must be disabled for them to work (WebSockets are HTTP/1 only).

nmiyake (changes requested):
Right -- to further clarify, do you expect/require this to be true for every instance of your service, or will it be per deployment?

If it's the former (and you expect that most legitimate uses of this option to be the former), then I think it would make more sense to have this as a `func (s *Server) WithDisableHTTP2() *Server {` builder function that is then called and set at a product/service level in code.

Having it be part of configuration (as it is in this PR) probably only makes sense if we expect this value to be configured differently on a per-environment/deployment basis.

(Marking review as "Request changes" until we get this thread resolved)

okushchenko:
@nmiyake the suggestion to move this setting to an option makes sense. I went ahead and made these changes. Please, take another look.

nmiyake (approved):
Great, this looks much better than adding to config -- thanks!

Issue #1026, opened 2025-11-04
Witchcraft-go-server V3


tabboud:
This issue tracks the full set of changes that will be made for the new V3 release of `witchcraft-go-server`:

- [x] Remove deprecated `rest` package
- [x] Migrate to a generic server object over the config types
	- [x] Migrate to `refreshable/v2` and remove deprecated refreshables
	- [x] Update conjure-go-runtime to v3
- [x] Use inclusive language (`blocklist`, `denylist`, etc) for public server options
~- [ ] Remove `WithStrictUnmarshalConfig` in favor of a new config validation approach~
- [x] Update module to v3
- [x] Add backport instructions to the README

PR #1030, opened 2025-11-06
Break: Support generic Server[I, R] over configuration types

tabboud:
## Before this PR
WGS heavily relied on `interface{}` types and reflection to enforce config type constraints since this was written prior to generic type parameters being introduced to Go. Conjure-go-runtime and the refreshables module both support the use of generics in their respective APIs which allows for WGS to parameterize the `*Server` object over the config types using strongly typed APIs throughout.

## After this PR

==COMMIT_MSG==
Support generic `witchcraft.Server[I, R]` over configuration types

witchcraft-go-server now supports generic type parameters over the configuration types, removing the need to type assert within the server `InitFunc`. The `WithRuntimeConfigType` and `WithInstallConfigType` options have been removed since the `witchcraft.Server` object is now strongly typed. Custom configuration types must now implement the `BaseInstallConfig` and `BaseRuntimeConfig` constraint interfaces and provide the types at server construction time. If the configuration types embed the base witchcraft types, then the constraint interfaces will be satisfied.

Generated refreshables have been removed in favor of using the refreshable/v2 module that supports type parameters. Similarly, conjure-go-runtime has been updated to the new v3 module which now uses the generic refreshable/v2 module.
==COMMIT_MSG==

Part of #1026
This was heavily pulled from @bmoylan (#1005) but also does the full conversion to CGR v3 and refreshable v2


go.mod
```diff
@@ -6,13 +6,13 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/nmiyake/pkg/dirs v1.1.0
-	github.com/palantir/conjure-go-runtime/v2 v2.99.0
+	github.com/palantir/conjure-go-runtime/v3 v3.0.0-20251105030408-a8855f8fbbf7
```
tabboud: This will be updated to the proper release before merging. I intentionally did not cut the v3.0.0 release of CGR to ensure everything looks good ETE.

README.md
```diff
@@ -716,51 +692,18 @@ not relevant to the function, there is no need to subscribe to it. The following
 the `runtimeConfig` refreshable: 
 
 ```go
-myNumRefreshable := runtimeConfig.Map(func(in interface{}) interface{} {
-    return in.(AppRuntimeConfig).MyNum
+myNumRefreshable, _ := refreshable.Map(runtimeConfig, func(in AppRuntimeConfig) int {
```
nmiyake: I think this documentation should mention the second return value (and/or there should be a comment on this example code to explain why it is being dropped/ignored)

witchcraft/ratelimit/middleware.go
```diff
@@ -49,7 +48,7 @@ var MatchMutating MatchFunc = func(req *http.Request, vals wrouter.RequestVals)
 // TODO: We should set the Retry-After header based on how many requests we're rejecting.
 //
 //	Maybe enqueue requests in a channel for a few seconds in case other requests return quickly?
-func NewInFlightRequestLimitMiddleware(limit refreshable.Int, matches MatchFunc, healthcheck reporter.HealthComponent) wrouter.RouteHandlerMiddleware {
+func NewInFlightRequestLimitMiddleware(limit func() int, matches MatchFunc, healthcheck reporter.HealthComponent) wrouter.RouteHandlerMiddleware {
```
nmiyake: Why is this `func() int` instead of being a refreshable?

witchcraft/witchcraft.go
```diff
@@ -105,7 +103,7 @@ type Server struct {
 	// the server. The ctx provided to the function is valid for the lifetime of the server. If nil, uses a function
 	// that returns a default file-based Refreshable that reads the file at "var/conf/runtime.yml". The value of the
 	// Refreshable is "[]byte", where the byte slice is the contents of the runtime configuration file.
-	runtimeConfigProvider func(ctx context.Context) (refreshable.Refreshable, error)
+	runtimeConfigProvider func(ctx context.Context) refreshable.Validated[[]byte]
```
nmiyake, line 111: Update documentation to include information about validation

witchcraft/witchcraft.go
```diff
@@ -251,78 +241,57 @@ type Server struct {
...
-	// "WithInstallConfigType" function (the default is config.Install).
-	InstallConfig interface{}
+	// InstallConfig is the install configuration.
+	InstallConfig I
 
-	// RuntimeConfig is a refreshable that contains the initial runtime configuration. The type returned by the
-	// refreshable is determined by the struct provided to the "WithRuntimeConfigType" function (the default is
-	// config.Runtime).
-	RuntimeConfig refreshable.Refreshable
+	// RuntimeConfig is a refreshable that contains the initial runtime configuration.
```
nmiyake: Why the "initial" qualifier here?

nmiyake (commented):
Thanks! Looks good overall, but some questions

witchcraft/witchcraft.go
tabboud: This was the original godoc - I just removed the portion about the `WithRuntimeConfigType`. Agree with the sentiment though since it's not just the "initial" value but could also change over time. Will remove "initial".

README.md
tabboud: Good idea. I added in the `unsubscribe` portion and also a new section below which uses `refreshable.View`.

witchcraft/ratelimit/middleware.go
tabboud: Good catch. Updated to `refreshable.Refreshable[int]`

nmiyake (dismissed):
Thanks, looks good

tabboud:
@nmiyake Mind giving this another +1? I updated the CGR version to the newly released [v3.0.0](https://github.com/palantir/conjure-go-runtime/releases/tag/v3.0.0)

nmiyake (approved):

PR #1031, opened 2025-11-06
Break: Remove deprecated /debug/pprof endpoints

tabboud:
## Before this PR
`/debug/pprof` endpoints are registered on the management server to allow for debugging and telemetry collection, however these have been superseded by the `/debug/diagnostic/{diagnosticType}` endpoints. 

## After this PR

==COMMIT_MSG==
Remove deprecated `/debug/pprof` endpoints

The `/debug/diagnostic/{diagnosticType}` endpoints can be used to retrieve most of the same information that was previously retrieved by the `/debug/pprof/*` endpoints
==COMMIT_MSG==

Part of #1026

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,16 +73,6 @@ specifies the encoding and format of the response. The following types are curre
 * `metric.names.v1`: Records all metric names and tag sets in the process's metric registry.
 * `os.system.clock.v1`: Plaintext string representing the current time as measured by the process in the RFC 3339 Nano format.
 
-#### \[Deprecated] Pprof routes
-The following routes are registered on the management server (if enabled) to aid in debugging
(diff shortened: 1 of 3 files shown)


nmiyake (approved):
Thanks -- approving, but double-checking that you have a general sense for clients that do actually call this endpoints and their workflows, an are confident that new servers that don't respond for these endpoints will not crash any critical workflows?

I think it is fair to do this as part of the v3, but we should definitely highlight it in release notes and perhaps have a walkthrough / check list for migrating services for them to do a request log analysis to ensure that these endpoints aren't being called (or that it is okay to break the callers) when doing the migration.

tabboud:
It turns out there are users that rely on these endpoints to assist in debugging high-side, so I think it's best if we leave these endpoints registered for now.

PR #1050, opened 2025-12-18
improvement: Add in a TaskManager and JobManager for starting application background jobs

k-simons:
- Add in a TaskManager and JobManager for starting application background jobs
- TaskManager will be used for additional tasks, but for the moment just add the ability to run jobs
- Add an option to configure the JobRunner health check, if not supplied a default value is used


witchcraft/job_manager.go
```diff
@@ -0,0 +1,67 @@
...
+// JobManager manages the lifecycle of background jobs within a witchcraft server.
+type JobManager interface {
+	// AddJobs registers and starts the provided jobs. Each job runs in its own goroutine
+	// and executes periodically based on its configured interval. Jobs that have
+	// WithStartImmediately set will run once immediately upon being added.
+	//
+	// The first call to AddJobs with at least one job will register a health check source
+	// with the server. Subsequent calls will not register additional health sources.
+	// Job execution results (success or error) are reported to this health check.
+	// If no job is every registered, the health check will never be added
```
tabboud:
```suggestion
	// If no job is ever registered, the health check will never be added
```

witchcraft/job_manager.go
```diff
@@ -0,0 +1,67 @@
...
+	registerHealth              func(healthSource healthstatus.HealthCheckSource)
+}
+
+func NewJobManager(
+	keyedErrorHealthCheckSource window.KeyedErrorHealthCheckSource,
+	registerHealth func(healthSource healthstatus.HealthCheckSource)) JobManager {
+	return &defaultJobManager{
+		jobRunner:                   jobs.NewDefaultJobRunner(keyedErrorHealthCheckSource),
+		keyedErrorHealthCheckSource: keyedErrorHealthCheckSource,
+		registerHealth:              registerHealth,
```
tabboud, line 55: nit: Let's guard against a nil `registerHealth` func, so that we can avoid a panic

witchcraft/job_manager.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (d *defaultJobManager) AddJobs(ctx context.Context, job ...jobs.Job) {
```
tabboud: nit: Let's make jobs plural -> `jobs ...jobs.Job`. I know it stutters with the pkg name but it reads a bit nicer below

witchcraft/job_manager_test.go
```diff
@@ -0,0 +1,87 @@
...
+	)
+	registerHealth := func(healthSource healthstatus.HealthCheckSource) {}
+	jobManager := NewJobManager(healthCheckSource, registerHealth)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	var jobRan atomic.Bool
+	job := jobs.NewDefaultJob("test-job", function.NewRunnableFromFunc(func(ctx context.Context) error {
+		jobRan.Store(true)
+		return nil
+	}), jobs.WithStartImmediately(true))
```
tabboud, line 82: We should remove the bool from this option since you have to opt-in to it anyways.
```go
jobs.WithStartImmediately()
```

witchcraft/job_manager.go
```diff
@@ -0,0 +1,67 @@
...
+	// and executes periodically based on its configured interval. Jobs that have
+	// WithStartImmediately set will run once immediately upon being added.
+	//
+	// The first call to AddJobs with at least one job will register a health check source
+	// with the server. Subsequent calls will not register additional health sources.
+	// Job execution results (success or error) are reported to this health check.
+	// If no job is every registered, the health check will never be added
+	//
+	// The provided context is passed to each job and can be used for cancellation.
+	// When the context is cancelled, all jobs will stop after their current execution completes.
```
tabboud: So If I wanted to stop/cancel a subset of jobs, is it expected that I would configure a different context with cancelation versus the other jobs that should keep running? That appears to be the only way to stop/cancel any jobs.

witchcraft/job_manager_test.go
k-simons: At some point Nick and I disucssed that and the downside is that you can't ever set it back(not that you would want to), so we kept it as an arg

witchcraft/job_manager.go
k-simons: I don't follow? Why would that be different than any other arg that is passed in? Like same with keyedErrorHealthCheckSource?

witchcraft/job_manager.go
k-simons: Yep, updated the comment

tabboud (approved):

PR #1069, opened 2026-02-02
Add install/runtime config validation

tabboud:
## Before this PR
There was no built-in way to run validation for install/runtime config types

## After this PR

==COMMIT_MSG==
Add optional config validation support for install and runtime configs

Validation can be enabled via `WithInstallConfigValidation` and `WithRuntimeConfigValidation`. When enabled, if the config type implements the `config.Validator` interface, the `Validate` method is called after unmarshaling. Runtime config validation supports configurable failure behaviors including logging, metrics, health check, startup failure, and reload rejection. A `StrictUnmarshaling` option is also available for both install/runtime config to fail on unknown YAML fields at deserialization time.
==COMMIT_MSG==

### DIFF
--- a/config/validation.go
+++ b/config/validation.go
@@ -1,0 +1,81 @@
+// Copyright (c) 2026 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


nmiyake (approved):
Thanks -- the implementation here looks good to me.

PR #1088, opened 2026-02-26
Feature: Add runtime config option to control audit.2 log emission

tabboud:
## Before this PR
No ability to dynamically control whether audit.2 logs are emitted or not.

## After this PR

==COMMIT_MSG==
Add runtime config option to control audit.2 log emission

`Runtime.AuditConfig.ProduceAuditV2Logs` controls whether audit.2 logs are emitted. When false, all audit.2 output is suppressed, including entries dual-logged from the audit.3 logger. Defaults to enabled when unset.
==COMMIT_MSG==

### DIFF
--- a/config/runtime.go
+++ b/config/runtime.go
@@ -49,10 +49,10 @@ type AuditConfig struct {
 
 	Environment string `yaml:"environment,omitempty"`
 
-	// ProduceAuditV2Logs controls whether audit.2 logs are emitted.
+	// ProduceAudit2Logs controls whether audit.2 logs are emitted.
 	// If false, all audit.2 log output is suppressed, including dual-logging from audit.3.
 	// If not set or set to true, audit.2 logs are emitted.
-	ProduceAuditV2Logs *bool `yaml:"produce-audit2-logs,omitempty"`
+	ProduceAudit2Logs *bool `yaml:"produce-audit2-logs,omitempty"`
 }
 
 // BaseRuntimeConfig implements the BaseRuntimeConfig interface.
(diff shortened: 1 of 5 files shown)


config/runtime.go
```diff
@@ -48,6 +48,11 @@ type AuditConfig struct {
 	Service string `yaml:"service,omitempty"`
 
 	Environment string `yaml:"environment,omitempty"`
+
+	// ProduceAuditV2Logs controls whether audit.2 logs are emitted.
+	// If false, all audit.2 log output is suppressed, including dual-logging from audit.3.
+	// If not set or set to true, audit.2 logs are emitted.
+	ProduceAuditV2Logs *bool `yaml:"produce-audit2-logs,omitempty"`
```
nmiyake: Given the contract outlined here, semantically, would it make more sense for this to be `DisableAuditV2LogOutput`?

In that case, if the value is present and set to `true`, then it clearly means that `audit.2` output is disabled/will not occur, and if the value is not present or false, that in itself doesn't provide any guarantees/implications about what would happen (`audit.2` logs may or may not be produced/recorded based on other configuration, but there are no guarantees).

nmiyake (changes requested):
Suggestion to invert config/setting for clarity and usability

config/runtime.go
tabboud: Discussed offline - While inverting the boolean makes things more clear we are going to stick with the existing implementation to match what exists in other libraries.

nmiyake (approved):

PR #1090, opened 2026-03-02
Improvement: Deprecate wrouter and wresource in favor of palantir/witchcraft-go-router

tabboud:
## Before this PR
Both `wrouter` and `wresource` have been pulled out into https://github.com/palantir/witchcraft-go-router

## After this PR

==COMMIT_MSG==
Deprecate `wrouter` and `wresource` 

`wrouter` and `wresource` pkgs provided by `github.com/palantir/witchcraft-go-router` should be used instead.
==COMMIT_MSG==


witchcraft/wresource/resource.go
```diff
@@ -12,104 +12,24 @@
...
-
-func (r *resourceImpl) Patch(endpointName, path string, handler http.Handler, params ...wrouter.RouteParam) error {
-	return r.Register(endpointName, http.MethodPatch, path, handler, params...)
-}
-
-func (r *resourceImpl) Delete(endpointName, path string, handler http.Handler, params ...wrouter.RouteParam) error {
-	return r.Register(endpointName, http.MethodDelete, path, handler, params...)
-}
+// Deprecated: Use github.com/palantir/witchcraft-go-router/wresource.New instead.
+var New = routerwresource.New
```
nmiyake: Does this work as expected? Would have expected needing to do:

```suggestion
func New(resourceName string, router wrouter.Router) Resource {
	return routerwresource.New(resourceName, router)
}
```

The basic case of "calling" the function may still compile, but I think this also makes it such that `wresource.New` is an exported variable that could be reassigned as well (and also may change the type from being a strict function to a variable that stores a function).

After reading over the article that introduced aliases again (https://go.dev/talks/2016/refactor.article#TOC_4.2.), I think it is the case that the recommended way to migrate functions is to wrap them.

The original proposal for aliases and gradual refactors (https://github.com/golang/proposal/blob/master/design/16339-alias-decls.md, https://github.com/golang/go/issues/18130) had a proposal for syntax for creating function aliases, but they ended up concluding that the existing method of wrapping was sufficient, and restricted the alias syntax only for types.

wrouter/wgorillamux/routerimpl.go
```diff
@@ -12,97 +12,27 @@
...
-
-type router mux.Router
+// Deprecated: Use github.com/palantir/witchcraft-go-router/wrouter/wgorillamux.Param instead.
+type Param = routerwgorillamux.Param
 
-func (r *router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	(*mux.Router)(r).ServeHTTP(w, req)
-}
+// Deprecated: Use github.com/palantir/witchcraft-go-router/wrouter/wgorillamux.New instead.
+var New = routerwgorillamux.New
```
nmiyake: Similar to comment above, I think it may be better to keep functions and return new type/call to equivalent rather than just reassigning them.

wrouter/whttprouter/routerimpl.go
```diff
@@ -12,100 +12,27 @@
...
-
-func HandleOPTIONS(handle bool) Param {
-	return paramFunc(func(r *httprouter.Router) {
-		r.HandleOPTIONS = handle
-	})
-}
-
-type router httprouter.Router
+// Deprecated: Use github.com/palantir/witchcraft-go-router/wrouter/whttprouter.Param instead.
+type Param = routerwhttprouter.Param
```
nmiyake, line 24: Same as above for functions

wrouter/paramperms.go
```diff
@@ -12,129 +12,18 @@
...
-			continue
-		}
-		if currPerms.Forbidden(paramName) {
-			return true
-		}
-	}
-	return false
-}
+// Deprecated: Use github.com/palantir/witchcraft-go-router/wrouter.NewCombinedParamPerms instead.
+var NewCombinedParamPerms = routerwrouter.NewCombinedParamPerms
```
nmiyake: Rewrap functions

nmiyake (changes requested):
Thanks -- I think the type alias makes sense, but as mentioned in the comment, for functions, I think the original function declarations should be maintained but just call through to the new type implementation for maximal compatibility (fairly mechanical, but also something that a tool like Claude could likely do pretty easily if helpful)

witchcraft/wresource/resource.go
tabboud: Interesting... I was following the same approach that was taken for Refreshables
https://github.com/palantir/witchcraft-go-server/blob/668037cae5d89c5ab9b14856abd9698a8c6f04d3/witchcraft/refreshable/deprecated_types.go#L31

Agree with your synthesis on this changing to an exported var containing a func pointer. Wrapping them seems like the right approach. Thanks for the links.

nmiyake (approved):

PR #1093, opened 2026-03-06
improvement: Bump to C-G-R 3.4.0 and refreshable/v2 v2.7.1

k-simons:
-  Bump to C-G-R 3.4.0 and refreshable/v2 v2.7.1
- Fix compile time breaks to the validated library in tests and pass through ctx


witchcraft/witchcraft.go
```diff
@@ -771,10 +771,13 @@ func (s *Server[I, R]) Start() (rErr error) {
 	ctx = s.withLoggers(ctx)
 
 	// load runtime configuration
-	refreshableRuntimeCfg, configHealthCheckSources, err := s.initRuntimeConfig(ctx)
+	refreshableRuntimeCfgValidated, configHealthCheckSources, err := s.initRuntimeConfig(ctx)
```
bmoylan, line 774: Are we going to do anything with the errors from the Validated?

bmoylan (approved):

witchcraft/witchcraft.go
k-simons: I didn't change that logic; but yes at the moment it is in `initRuntimeConfig` we do:

```
if _, err := runtimeConfigProvider.Validation(); err != nil {
		return nil, nil, err
	}
```

